PR #1, opened 2018-11-12
Lift and shift.

samiwel:
This PR takes existing author, author-api, publisher and graphql schema projects and brings all of the source files under one roof.


samiwel:
[Switch to monorepo](https://trello.com/c/r0tAdpyn/638-switch-to-monorepo)

PR #23, opened 2018-11-19
Dont strip html from introductions when piping

jonnyshaw89:
Publisher was stripping some of the HTML out of the description when published.

This PR limits that change to just inserting the piped values.

### How to review 
Add a formatted section introduction and see that the formatting and piping still work

### DIFF
--- a/eq-publisher/src/eq_schema/Block.js
+++ b/eq-publisher/src/eq_schema/Block.js
@@ -1,7 +1,10 @@
 const Question = require("./Question");
 const RoutingRule = require("./RoutingRule");
 const RoutingDestination = require("./RoutingDestination");
-const { getInnerHTMLWithPiping } = require("../utils/HTMLUtils");
+const {
+  getInnerHTMLWithPiping,
+  unescapePiping
+} = require("../utils/HTMLUtils");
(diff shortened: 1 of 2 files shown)


eq-publisher/src/eq_schema/Block.js
```diff
@@ -19,7 +19,7 @@ const getLastPage = flow(
 const processPipedText = ctx =>
   flow(
     convertPipes(ctx),
-    getInnerHTMLWithPiping
+    unescapePiping
```
tgandrews: I guess we should add some tests to cover this.

I think we want to get the inner html for title. So is that affected with this change?

eq-publisher/src/eq_schema/Block.test.js
```diff
@@ -108,7 +116,23 @@ describe("Block", () => {
...
+      const introBlock = Block.buildIntroBlock(
+        introduction,
+        0,
+        createContext()
+      );
+
+      expect(introBlock.title).toEqual("{{ answers['answer123'] }}");
+    });
+
+    it("should handle piped values in title while stripping html", () => {
```
tgandrews: Test says `title` but expectation is on `description`

PR #24, opened 2018-11-19
Pull latest front-end changes from upstream.

samiwel:
This PR pulls the latest front-end author changes into the monorepo.

### How to review 
All recent changes should be included with this PR.


samiwel:
AFAIK the vulnerability being reported by Snyk doesn't have a fix yet. For now we should accept this as a risk and look to resolve once a fix becomes available.

tgandrews:
@samiwel we can ignore this in snyk

tgandrews:
I have ignored the issue in snyk for 90 days.

tgandrews:
@samiwel I guess root `.travis.yml` should also be updated to 10?

eq-author/.travis.yml
```diff
@@ -8,7 +8,7 @@ services:
 language: node_js
 
 node_js:
-  - "8"
```
tgandrews: I guess this file should be removed and the changed pushed up to the root.

samiwel: Yeah this will be the same with each of the subdirectories. I think removing them all together in a separate PR makes sense.

samiwel: Another thing that occurred to me this morning is that we'll have to make changes so that the schema file is imported from the local filesystem as opposed to being an npm dependency. This is because the concourse pipeline has changed to deploy on a new tag in github. So we want to avoid the npm publishing process going forward as this will create a tag and deploy changes to pre-prod.

PR #26, opened 2018-11-20
Upgrade dependencies to fix issues building on node 10.

samiwel:
Publisher did not seem to build correctly on a fresh install with node 10.
The issue was due to missing pre-compiled binaries for the fsevents package.
Performing a `yarn upgrade` fixed the dependencies in the `yarn.lock` file.

### How to review 
- Delete node_modules.
- Ensure you're using Node 10
- Run yarn install and ensure publisher is built correctly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ services:
 language: node_js
 
 node_js:
-  - "8"
+  - "10"
 
 cache:
   yarn: true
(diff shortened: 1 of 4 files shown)


samiwel:
No code changes here. I think the `codecov` check can be ignored. Or at least we can look into the why the coverage metric goes down when there are no changes separately.

PR #30, opened 2018-11-20
Fixed confirmation title bug

SamGodwin2:
Small bug where the id of a field didn't match the field name of the graphql query i.e rather than sending title:"foo" to the mutation it was instead sending confirmation-title:"foo" meaning that title was null. 

This is a quick fix for this issue but was wondering if instead we needed to me a bit more though and add a required proptype to the RTE of `graphqlFieldName` this might mean we could use less generic ids as well as I think there are a few places where we have different RTE's with an id of "title".

### How to review 
Not sure how to test for this as the cypress test will all pass due to just asserting that the stuff held in the inputs state is correct and they dont make sure that stuff has been stored in the db. Any ideas?

### DIFF
--- a/eq-author/src/components/QuestionConfirmationRoute/Editor.js
+++ b/eq-author/src/components/QuestionConfirmationRoute/Editor.js
@@ -48,6 +48,7 @@ export class UnwrappedEditor extends React.Component {
       <Wrapper>
         <RichTextEditor
           id="confirmation-title"
+          name="title"
           label="Confirmation question"
           value={title}
           onUpdate={this.handleRichTextUpdate}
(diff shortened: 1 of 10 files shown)


eq-author/src/components/QuestionConfirmationRoute/Editor.js
```diff
@@ -48,6 +48,7 @@ export class UnwrappedEditor extends React.Component {
       <Wrapper>
         <RichTextEditor
           id="confirmation-title"
```
tgandrews: You can remove this id as its not used anymore.

SamGodwin2: `id`'s a required prop in the RTE, so I'll have a look at if that still needs to be the case in the refactor I'll do in the next pr.

SamGodwin2:
Yeah thats the best sort of thing I could come up with, issue is we should really do it everywhere we use the RTE, might cause the tests to slow down even more? Maybe worth just waiting for now and try get some true e2e tests done soonish?

tgandrews:
We have some starting e2e but I guess it needs to be something we keep adding to

PR #31, opened 2018-11-20
Yarn build outside of container

jonnyshaw89:
Just giving it a go as currently `yarn install` is being run twice

### How to review 
Don't yet


tgandrews:
I guess we'd have to mirror our build env in concourse pipeline as well.

You could just try copying the node_modules or the npm cache from travis.

jonnyshaw89:
> I guess we'd have to mirror our build env in concourse pipeline as well.

PR #37, opened 2018-11-21
View and print a copy of your response

jonnyshaw89:
Adds the ability to view your answers after submission to all questionnaires

### How to review 
Launch a questionnaire and see if you have a link to view your answers after you submit


samiwel:
Interesting that the build didn't fail though...

tgandrews:
@samiwel the build didn't fail because the e2e was never run. Also, the e2e references the published version of publisher not the one in the monorepo. This needs looking at.

PR #38, opened 2018-11-21
Remove npm publishing 2a

tgandrews:
This PR is to enable to people make changes to all projects in monorepo and prove it still works. So, the monorepo should never be red or dependendent on other PRs. This means schema changes are used in the API and UI.

Additionally, there seemed to be instances where api was ready before postgres so this now ensures we are waiting postgres to be ready before starting the app.

### How to review 
1. Ensure you can run and make changes to schema locally which are reflected in UI and API.
1. CI should pass ~~(apart from API which will fail with breaking change)~~

### Dependencies
- [x] - #47 -  Adds the schema to the new location so this can be shown to run integration tests.


tgandrews:
[Remove NPM publish](https://trello.com/c/0WCzYz9z/750-remove-npm-publish)

eq-author-api/scripts/findBreakingChanges.js
```diff
@@ -37,7 +36,7 @@ const findDeprecatedDirective = directives => {
     const deprecatedDirectives = directives.filter(
       directive => directive.name.value === "deprecated"
     );
-    return deprecatedDirectives.length > 0 ? true : false;
+    return deprecatedDirectives.length > 0;
```
jonnyshaw89: 👍

tgandrews:
Ignore checks:
- Codecov - The JS changes are related to scripts and we are not testing scripts. We should probably add it as an explicit ignore.
- Snyk - Its complaining it cannot find the schema `package.json` which was removed.

PR #41, opened 2018-11-22
Fixed small bug in metadata piping

SamGodwin2:
There was a small bug in the piping of metadata due to the query requesting too many fields meaning that the get name function got confused and would end up returning a pointer to the current field in the content picker rather than the individual piece of metadata. 

### How to review 
Change makes sense, new assertions pass and the piping of metadata now works as expected.


samiwel:
[Fix display issue when piping metadata.](https://trello.com/c/cXovekKQ/748-fix-display-issue-when-piping-metadata)

PR #42, opened 2018-11-22
741 spike investigate removing empty html tags

rmccar:
This PR is to stop the creation of empty tags in the database and also adding a migration to remove any empty tags already in the database

### How to review 
- Make sure tests pass
- Check out and test by spinning up and tabbing through the fields not entering anything in, then check the database to make sure just empty fields are created


rmccar:
[Spike: Investigate removing empty HTML tags](https://trello.com/c/yvsCHwGZ/741-spike-investigate-removing-empty-html-tags)

eq-author/src/components/RichTextEditor/index.js
```diff
@@ -113,6 +115,18 @@ const getContentsOfPipingType = type => contents =>
 const getAnswerPipes = getContentsOfPipingType("answers");
 const getMetadataPipes = getContentsOfPipingType("metadata");
 
+const filterEmptyTags = value => {
+  if (!isNull(cheerio(trim(value)).html())) {
```
tgandrews: What is this first check doing?

eq-author/src/components/RichTextEditor/index.js
```diff
@@ -339,8 +353,8 @@ class RichTextEditor extends React.Component {
 
   handleBlur = () => {
     this.props.onUpdate({
-      name: this.props.name,
-      value: this.getHTML()
+      name: this.props.name ? this.props.name : this.props.id,
```
tgandrews: This looks like an accidental revert of @SamGodwin2's change.

eq-author/src/components/RichTextEditor/RichTextEditor.test.js
```diff
@@ -113,11 +113,20 @@ describe("components/RichTextEditor", function() {
...
+  it("empty RTE should call onUpdate with empty value onBlur", () => {
     wrapper.find("[data-test='rte-field']").simulate("blur");
     expect(props.onUpdate).toHaveBeenCalledWith({
       name: "test-name",
-      value: "<p></p>"
+      value: ""
+    });
+  });
+
+  it("shouldnt strip tags with a value", () => {
```
tgandrews:
```suggestion
  it("should not strip tags with a value", () => {
```

eq-author/src/components/PreviewSectionRoute/__snapshots__/SectionIntroPreview.test.js.snap
```diff
@@ -11,25 +11,26 @@ exports[`SectionIntroPreview should render an error when the content is missing
...
-    data-test="no-content"
-    large={true}
-    margin={false}
-  >
-    Missing Introduction Content
-  </Error>
+  <SectionIntroPreview__ContentBlock
+    dangerouslySetInnerHTML={
+      Object {
+        "__html": "<h2></h2>",
```
tgandrews: That's an empty tag so I guess the test needs to be updated as it looks like it was checking for an empty intro and `<h2></h2>` was empty before.

eq-author-api/utils/html.js
```diff
@@ -1,7 +1,7 @@
 const cheerio = require("cheerio");
-const { isNull, toString } = require("lodash");
+const { isNull, toString, trim } = require("lodash");
 
-const isHtml = value => !isNull(cheerio(value).html());
+const isHtml = value => !isNull(cheerio(trim(value)).html());
```
tgandrews: Why do we need to trim?

rmccar: This is to make sure white space is accounted for, incase there is white space either side of the tags for some reason

tgandrews: Did we find that anywhere?

eq-author/src/components/RichTextEditor/index.js
rmccar: It makes sure there is html in the value otherwise the cheerio.text on the next line wont work

tgandrews: Can we give it a name then like - `isHtml`?

eq-author/src/components/PreviewSectionRoute/__snapshots__/SectionIntroPreview.test.js.snap
rmccar: I have updated the snapshots so they reflect the new empty state

PR #44, opened 2018-11-23
Tidy up Author vars

jonnyshaw89:
Remove unused Vars and re-organise them into the "production by default" best practice.
This allows us to have the minimum amount of "switches" in production, which should reduce risk.
Vars should then be overridden in dev/test

Auth is now set to always on. This will be required going forward as the API will require an auth token.

### How to review 
Does is still work as expected?


eq-author/src/index.js
```diff
@@ -52,7 +50,7 @@ const client = createApolloClient(link, cache);
 
 store = configureStore(history, client);
 
-if (window.Cypress && config.REACT_APP_FUNCTIONAL_TEST === "true") {
+if (window.Cypress && process.env.REACT_APP_FUNCTIONAL_TEST === "true") {
```
tgandrews: We need this to be true for cypress tests. And we only build the UI once so it needs to be configurable from the docker run args.

jonnyshaw89:
@tgandrews fancy having another look?

eq-author/.env.development
```diff
@@ -1,10 +1,2 @@
-REACT_APP_BASE_NAME=""
 REACT_APP_API_URL="http://localhost:4000/graphql"
-REACT_APP_LAUNCH_URL="http://localhost:4000/launch"
-REACT_APP_USE_FULLSTORY="false"
-REACT_APP_USE_SENTRY="false"
-
-REACT_APP_ENABLE_AUTH="true"
```
rmccar: Just wondering why you removed this variable? I normally override it as false to allow me to get past the sign in process locally

jonnyshaw89: When the API required auth, we wont be able to use "Guest" anyway so we should all start setting the app up so it works the same for dev as it does when deployed

rmccar: Ok

PR #45, opened 2018-11-23
Fix slow connection removing section title

rmccar:
When using Author on a slower connection there is a race condition with the section introduction where adding section title and then adding a section introduction would remove the section title. This is to remove the second entity editor on the page so that the section introduction is controlled by the section one

### How to review 
- Run tests
- Throttle connection to slow 3g
- Create a new section
- Add a section title
- Add a section introduction
- Make sure section title doesn't disappear
- Make sure all other parts of the section introduction still work (delete, undo delete, etc)


rmccar:
[Fix for section title disappearing](https://trello.com/c/NyCnEVXF/749-fix-for-section-title-disappearing)

PR #49, opened 2018-11-26
Implemented automatic piping of checkbox answers

SamGodwin2:
One of the requirements for confirmation questions was that all checkbox answers would be automatically piped into the confirmation description, however this was missed off the original PR. this PR rectifys that and causes all checkbox answers to be automatically piped.

### How to review 
Tests pass and Publisher output is now correct.

### DIFF
--- a/eq-publisher/src/eq_schema/Group.test.js
+++ b/eq-publisher/src/eq_schema/Group.test.js
@@ -511,5 +511,48 @@ describe("Group", () => {
         expectedRunnerRouting
       );
     });
+
+    it("pipes in checkbox values from the previous questions", () => {
+      const ctx = ctxGenerator(null);
+
+      ctx.questionnaireJson.sections[0].pages[0].answers[0] = {
+        id: "6",
+        type: "Checkbox",
+        label: "Test",
+        description: "",
+        guidance: "",
+        options: [
+          {
+            id: "Foo",
+            label: "Foo"
+          },
+          {
+            id: "Bar",
+            label: "Bar"
+          },
+          {
+            id: "Baz",
+            label: "Baz"
+          }
+        ],
+        properties: {
+          required: false
+        }
+      };
+
+      const resultantJson = new Group(
+        ctx.questionnaireJson.sections[0].id,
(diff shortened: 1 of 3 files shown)


eq-publisher/src/eq_schema/Group.test.js
```diff
@@ -511,5 +511,48 @@ describe("Group", () => {
         expectedRunnerRouting
       );
     });
+
+    it("pipes in checkbox values from the pervious questions", () => {
```
tgandrews:
```suggestion
    it("pipes in checkbox values from the previous questions", () => {
```

eq-publisher/src/eq_schema/builders/confirmationPage/ConfirmationPage.js
```diff
@@ -66,6 +66,10 @@ const buildAuthorConfirmationQuestion = (
   const confirmationQuestionObject = {
     id: `confirmation-page-for-${page.id}`,
     title: page.confirmation.title,
+    description:
+      page.answers[0].type === "Checkbox"
```
tgandrews: Do we have a constant for this. Or not in `publisher`?

eq-publisher/src/eq_schema/Group.test.js
```diff
@@ -511,5 +511,48 @@ describe("Group", () => {
...
+      const ctx = ctxGenerator(null);
+
+      Object.assign(ctx.questionnaireJson.sections[0].pages[0].answers[0], {
```
tgandrews: It looks like you are defining everything on the answer. If you are not taking some its previous default values I'd prefer `ctx.questionnaireJson.sections[0].pages[0].answers[0] = ...` just so its obvious what is going on

eq-publisher/src/eq_schema/builders/confirmationPage/ConfirmationPage.js
SamGodwin2: Not in Publisher, probably worth adding them though.

PR #50, opened 2018-11-26
Fix for Codecov

rmccar:
Fix for Codecov to stop it erroring when we push to git

### How to review 
- Push something to git that will trigger Codecov
- Make sure it passes

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,13 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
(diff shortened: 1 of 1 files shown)


tgandrews:
Looks like we used to have one of these in author.
https://github.com/ONSdigital/eq-author/blob/master/.codecov.yml

rmccar:
Looks like codecov has started working properly again now because of this ill close this for now

PR #53, opened 2018-11-27
Add content picker to routing

adammahmood:
### What is the context of this PR?
- Implements content pickers for routing conditions and routing destinations
- Refactors some of the routing logic around filtering valid routing conditions and destinations by pushing it to the API
- Refactors some of the content picker to be able to handle new content picker types
- Tidy's up some small UI / UX aspects relating to the content picker as per discussions with @hamishtaplin 

### How to review 
- Run tests
- Attempt to create a complex routing scenario and ensure you are able to select expected content  for routing conditions and routing destinations

PR #55, opened 2018-11-28
732 -Question confirmation piping

tgandrews:
Initial question confirmation piping. This is reliant on #53 

### How to review 
1. Ensure you can view a survey with piped data in the question confirmation
1. The correct answers are pickable (this is not true as it does not allow those on from the current page)

### Dependencies
- [x] - #53 - Need the method `getPreviousAnswersForPage`

### DIFF
--- a/eq-author-api/repositories/QuestionConfirmationRepository.js
+++ b/eq-author-api/repositories/QuestionConfirmationRepository.js
@@ -1,6 +1,10 @@
 const { head, pick } = require("lodash/fp");
 
 const QuestionConfirmation = require("../db/QuestionConfirmation");
+const { PIPING_ANSWER_TYPES } = require("../constants/pipingAnswerTypes");
+
+const QuestionPageRepostory = require("./QuestionPageRepository");
+const PreviousAnswerStrategy = require("./strategies/previousAnswersStrategy");
 
 const USER_MODIFIABLE_FIELDS = [
   "title",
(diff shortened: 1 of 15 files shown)


tgandrews:
[Piping into playback confirmation title](https://trello.com/c/NZHx1ULL/732-piping-into-playback-confirmation-title)

eq-author-api/repositories/QuestionConfirmationRepository.js
```diff
@@ -45,12 +49,28 @@ module.exports = knex => {
       .restore(id)
       .then(head);
 
+  const getPipingAnswers = async id => {
+    const { pageId } = await findById(id);
+    return PreviousAnswerStrategy(knex).getPreviousAnswersForPage({
+      id: pageId,
+      includeSelf: true,
```
samiwel: Perhaps I misunderstood how this flag works, but when I tested e2e i wasn't able to select answers on the page that the confirmation page hangs off. I set something up like this (note I created Q2 after Q1):

- S1
  - Q2
    - Answer 2
  - Q1
    - Answer 1
    - Confirmation page

On the confirmation page, I was able to pipe answer 1 but not answer 2. Is that expected?

tgandrews: No thats not intentional. The include self should include it in the list of pages. I'll look into it.

tgandrews: Is your branch up to date this was resolved and I cannot recreate it locally. I also have a cypress test that covers it.

samiwel: Apologies, looks like my branch didn't have the latest changes. After a pull this worked as expected for me. 👍

PR #56, opened 2018-11-28
Bug - Section intro sidenav highlight

tgandrews:
Fix issue where highlight lost in sidenav on section into preview

### How to review 
1. In staging, see that the side nav section highlight is lost when navigating to a section preview (add a section introduction). 
1. See that with this branch the highlight is maintained.

### DIFF
--- a/eq-author/src/components/NavigationSidebar/SectionNavItem.js
+++ b/eq-author/src/components/NavigationSidebar/SectionNavItem.js
@@ -1,28 +1,32 @@
+import gql from "graphql-tag";
 import React from "react";
 import styled from "styled-components";
+import { withRouter } from "react-router-dom";
 
+import CustomPropTypes from "custom-prop-types";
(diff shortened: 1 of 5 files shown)


eq-author/src/components/NavigationSidebar/SectionNavItem.js
```diff
@@ -1,28 +1,36 @@
...
+    match: CustomPropTypes.match
   };
 
   render() {
-    const { questionnaire, section, ...otherProps } = this.props;
-    const url = buildSectionPath({
+    const { questionnaire, section, match, ...otherProps } = this.props;
+    let tab = match.params.tab;
+    if (tab === "routing") {
+      tab = "design";
```
hamishtaplin: This feels weird. I trust there's no better way?

tgandrews: I pushed this into the Routing page route itself so it will ensure that it only gets pages.

PR #57, opened 2018-11-28
Changes to support multiple additional detail write in fields

SamGodwin2:
Changed the implementation of "Other" answers in our app to allow for multiple "other" answers in a single question. Most instances of the use of "other" have been changed to either "additionalAnswer". Answers are now a optional field on options. 

### How to review 
Tests pass, and the changes allow for multiple "other"/"additionalAnswers" per multiple choice answer. 

#### Dependant on work currently in flight by Runner, this PR cannot be implemented until after Runners implementation of "detail_answer"

https://github.com/ONSdigital/eq-survey-runner/pull/1920


eq-publisher/src/eq_schema/RoutingRule.test.js
```diff
@@ -200,9 +202,9 @@ describe("Rule", () => {
     });
   });
 
-  it("should build valid runner routing with 'other' answers", () => {
+  it("should build valid runner routing with 'additionAnswer' answers", () => {
```
tgandrews: I presume this should be `additionalAnswer`

eq-author/cypress/integration/authenticated/builder_spec.js
```diff
@@ -640,13 +640,13 @@ describe("builder", () => {
       }
     };
 
-    it('should add an "other" answer', () => {
+    it("should add an additionalValue option answer", () => {
       addMultipleChoiceAnswer();
       cy.get(testId("option-label")).should("have.length", 2);
       cy.get(testId("other-answer")).should("have.length", 1);
```
tgandrews: Can we rename these from `other-answer` to `additional-answer`

eq-author-api/tests/utils/graphql.js
```diff
@@ -136,25 +131,8 @@ const createAnswerMutation = `
 const createOtherMutation = `
   mutation CreateOther($input: CreateOtherInput!) {
```
tgandrews: I guess we should rename this as well. `CreateAdditional`?

eq-author-api/tests/utils/buildTestQuestionnaire.js
```diff
@@ -103,6 +107,19 @@ const createOrUpdateOption = async (
...
+    if (optionDetails.additionalAnswer) {
+      const additionalAnswer = await AnswerRepository.insert({
+        ...optionDetails.additionalAnswer
+      });
+      await AnswerRepository.update({
+        id: additionalAnswer.id,
+        parentAnswerId: answer.id
+      });
+      await createOption({
+        otherAnswerId: additionalAnswer.id,
```
tgandrews: More places to rename

eq-author-api/schema/typeDefs.js
```diff
@@ -421,8 +416,8 @@ type Mutation {
   updateOption(input: UpdateOptionInput!): Option
   deleteOption(input: DeleteOptionInput!): Option
   undeleteOption(input: UndeleteOptionInput!): Option
-  createOther(input: CreateOtherInput!): OptionWithAnswer
-  deleteOther(input: DeleteOtherInput!): OptionWithAnswer
+  createOther(input: CreateOtherInput!): Option
```
tgandrews: This should be removed as per the delete and merged with `createOption`

eq-author/cypress/integration/authenticated/builder_spec.js
SamGodwin2: Yeah, wasn't really sure where to draw the line here. From a UI perspective  it will still be an "other" answer, least until/if we get a new design so was a bit unsure on whether the labels and the other attribs should be changed or not?

eq-author-api/tests/utils/graphql.js
SamGodwin2: Should be able to remove it altogether and fold it into create option hopefully.

SamGodwin2:
Yeah the doc changes are due to prittier auto-formatting the md when I changed the import schema query, I thought made sense to keep them, unless they conflict with anyone else's prittier config?

eq-author/cypress/integration/authenticated/builder_spec.js
tgandrews: I worry about new people not knowing the history and getting confused by this inconsistency.

tgandrews: As discussed there is a redesign of the UI to support different answer types for this as well so we look at fully removing "Other" later.

eq-publisher/src/eq_schema/Answer.js
```diff
@@ -155,14 +153,16 @@ class Answer {
     if (description) {
       option.description = description;
     }
+    if (additionalAnswer) {
+      option.detail_answer = {
```
hamishtaplin: is there a reason for using snake case?

SamGodwin2: Yeah Runner expects snake case instead of camel unfortunately.

eq-author-api/schema/typeDefs.js
```diff
@@ -421,8 +416,8 @@ type Mutation {
   updateOption(input: UpdateOptionInput!): Option
   deleteOption(input: DeleteOptionInput!): Option
   undeleteOption(input: UndeleteOptionInput!): Option
-  createOther(input: CreateOtherInput!): OptionWithAnswer
-  deleteOther(input: DeleteOtherInput!): OptionWithAnswer
+  createOther(input: CreateOtherInput!): Option @deprecated
+  deleteOther(input: DeleteOtherInput!): Option @deprecated
```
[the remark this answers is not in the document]
SamGodwin2: The deprecation was just to make it pass the is backwards compatible check on the schema, then make a new PR and delete them altogether. But I could just delete them here, or just add a resolvers that return nothing just so they're valid? Seems pointless to add the old functionality back in just delete in in the next PR?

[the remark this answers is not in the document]
samiwel: I'd say just delete them. Even deprecated query fields can be queried and if someone were to do that, then there would be no resolvers there to serve the request. This is a breaking change, we're not preserving backwards compatibility in the API, AFAIK but data is preserved? So any existing additional details fields I expect would still be there following the switch to the new API?

eq-author-api/repositories/OptionRepository.js
```diff
@@ -22,27 +24,48 @@ const checkForExistingExclusive = async answerId => {
...
   trx = db
 ) => {
+  let additionalAnswer;
   if (mutuallyExclusive) {
     await checkForExistingExclusive(answerId);
   }
+  if (hasAdditionalAnswer) {
+    additionalAnswer = await Answer.insert(
+      {
+        properties: getDefaultAnswerProperties("TextField"),
```
samiwel: If you use `Answer.creaateAnswer` instead of `insert` this should be taken care of for you.

jonnyshaw89:
Minor comment that we should improve on is that the commit title doesn't make sense on its own "Removed functionality preveventing people from"
The commit titles, like the PR titles, should be what the change is, not what was done or why.
The description should also be along the same context as to what this changes adds

SamGodwin2:
[Multiple "other" write in fields (12)](https://trello.com/c/vvgNO8ED/731-multiple-other-write-in-fields-12)

PR #58, opened 2018-11-28
Tech Debt - Reorganise ui

tgandrews:
Starting point for a discussion about reorganising the UI folder structure

I have tried to move things under "features" where the features we have are dashboard (the home page) and design, which is pretty much everything else. The `components` folder would become just shared components.

I've tried to push the controlling of the routes down to the entities and break it down by the tabs so its obvious where each of the things go.

You probably need to pull it to have a look at the folder structure.

### To do
1. Move more components into the relevant folders including enhancers.

### How to review 
1. Look at the folder structure.


adammahmood:
Although I'm not overly opinionated on the folder structure, it would be nice to have as many components/enhancers/containers co-located and moving to a feature based structure would most likely be beneficial.

It would be useful to have enhancers located nearer to the component it's being used in (often only one component) and would probably make it easier for new developers to find features/route components.

tgandrews:
@adammahmood I have pushed some enhancers down now into at least the page design page.

samiwel:
Appreciate you may not have got far enough into the refactoring, but I'm sure we have more features than design and dashboard. Was the plan to maybe have `validation`, `routing`, `piping` etc live under `features` as well?

tgandrews:
I pushed routing under question and the plan was to do the same for
validation as well as they are specific to those.

samiwel:
Happy to hear other people's thoughts, but I wonder if a slightly flatter structure (i.e. one level of subfolders within features) would be better. It's not immediately obvious that routing and validation would live under question route IMO (especially to someone new to the team). I think it does make sense to have top level routes but I think everything below these routes are either layout or feature specific components.

Other react frameworks that i've seen follow this kind of approach from what I can tell.

Would something like this make sense?

`components` -> common components that are shared across the app
`layout` -> layout specific components e.g. dashboard, editor layout
`pages` -> route specific pages e.g. questionnaire list, design, routing, preview
`features` -> feature specific components e.g. validation, routing, properties panel, settings modal, navigation, sign-in etc.

hamishtaplin:
I still `components` should be the root folder for all React code, with any subfolders going there instead of into top-level directories.

Also need to have a plan for the "containers" as they are used inconsistently and pretty much just regular components in some cases. 'enhancers' can be moved out of this directory and into the top-level, imo.

Agree with @samiwel here too, I'd prefer a flatter structure with some broader strokes of organisation.

hamishtaplin:
Tom and I talked this through and came up with a slightly flatter tree illustrated here:

https://docs.google.com/spreadsheets/d/1ScRRveCnb4BglQMktQ9-NL0m7E5c9IZvnA7Gxtbkeyo/edit?usp=sharing

adammahmood:
The structure illustrated in the spreadsheet looks good. 👍 I think the `App/` folder should be slightly flatter and we should flatten this layer as `Dashboard` doesn't really have much in it and most of it's within `Design` anyways: 

```
Dashboard
Design
SignIn
NotFound
Admin
```

samiwel:
👍 The structure proposed in the spreadsheet works for me. Like I said, I rarely look at the folder structure anyway tbh, I just Cmd+Click in the IDE! That said I think the proposed structure is an improvement over what we currently have.

We should get rid of the top level tests directory too and move the helpers.js into the src/tests/utils instead if possible - that always catches me out!

Whatever structure we settle on, we should aim for it being intuitive for any new starters, and as long as we have clear guidance (perhaps documented in the README) about what should live where then I don't have strong opinions either way.

tgandrews:
I think @adammahmood's comment makes sense as @hamishtaplin made the same comment as well. Until we a large feature then that design folder was basically 90% of the app.

@samiwel The docs comment is a good one as well and would probably help clarify some of the thinking behind it so when a new change comes along we can make appropriate restructure for it easily.

adammahmood:
Closing this for the moment to clean up our PR's.  We'll re-open when we decide to tacke this tech debt.  

https://trello.com/c/otdt0EY8/315-ui-structure

PR #60, opened 2018-11-29
Load DB credentials from AWS SecretsManager

jonnyshaw89:
This PR removes the need to pass database credentials as environment variables
This PR also fixes an issue where each Repository creates it own connection and pool of database connections.

### How to review 
Does the app work in the same way for Dev?


jonnyshaw89:
@samiwel yeh not sure how that happened. i don't think the `.gitignore` is working properly

eq-author-api/db/index.js
```diff
@@ -1,4 +1,45 @@
...
+const config = require("../config/knexfile.js");
 
-module.exports = require("knex")(config);
\ No newline at end of file
+const getConfig = async function(secretId) {
+  if (secretId && secretId !== "") {
+    const AWS = require("aws-sdk");
+    const client = new AWS.SecretsManager({ region: "eu-west-1" });
+
+    let secretPromise = client.getSecretValue({ SecretId: secretId }).promise();
```
samiwel: This can be `const`

eq-author-api/db/index.js
```diff
@@ -1,4 +1,45 @@
...
+
+    try {
+      let secret = await secretPromise;
+      let dbCredentials = JSON.parse(secret.SecretString);
```
samiwel: Both of these `lets` could be `const`. They do not get re-assigned.

eq-author-api/db/index.js
```diff
@@ -1,4 +1,45 @@
...
+      config.connection = {
+        host: dbCredentials.host,
+        port: dbCredentials.port,
+        user: dbCredentials.username,
+        password: dbCredentials.password,
+        database: dbCredentials.dbname
+      };
+      return config;
```
samiwel: Instead of mutating the `config` object (which is imported). I think it would be cleaner to define a locally scoped variable to hold the configuration and have a single return at the end of the function.

e.g. 

```js
const getConfig = () => {
 const result = {
   ...config
 }

 if (hasAWSCreds) {
   result.connection = ...
 }


 return result;
}

```

.gitignore
```diff
@@ -36,8 +36,8 @@ graphql.config.json
 public/status.json
 .env
 .env.*local
-cypress/videos/
-cypress/screenshots/
+eq-author/cypress/videos/
```
tgandrews:
```suggestion
**/cypress/videos/
```

.gitignore
```diff
@@ -36,8 +36,8 @@ graphql.config.json
...
+eq-author/cypress/screenshots/
```
tgandrews:
```suggestion
**/cypress/screenshots/
```

eq-author-api/config/knexfile.js
```diff
@@ -1,35 +1,7 @@
...
-    connection: process.env.DB_CONNECTION_URI,
-    pool: {
-      min: 2,
-      max: 10
-    },
-    migrations: {
-      tableName: "knex_migrations"
-    }
-  }
+  acquireConnectionTimeout: 5000
```
tgandrews: Why do we need this?

eq-author-api/db/index.js
```diff
@@ -1,4 +1,48 @@
...
-var config = require("../config/knexfile.js")[environment];
+const config = require("../config/knexfile.js");
+const createLogger = require("../utils/createLogger");
+const pinoMiddleware = require("express-pino-logger");
+const pino = pinoMiddleware();
+const logger = createLogger(pino.logger);
 
-module.exports = require("knex")(config);
\ No newline at end of file
+const getConfig = async function(secretId) {
```
tgandrews: This call is made a lot now and is now making a network call every time.

Could we do this once at startup?

eq-author-api/db/index.js
tgandrews: This could also be pushed down to the secret definition. So that becomes `const secret = client.getSecretValue...`

eq-author-api/db/index.js
```diff
@@ -1,4 +1,48 @@
...
+      connection = require("knex")(conf);
+    });
+  }
+  return connection;
+}
+
+getConnection();
+
+module.exports = {
+  getConnection
```
tgandrews: Could we export the connection directly? And then no one needs to make a call it just happens.

We could export getConnection and have that run as part of the initial start up? So that the connection is always available.

eq-author-api/db/index.js
jonnyshaw89: 👍

eq-author-api/config/knexfile.js
jonnyshaw89: Default connection timeout is 60000 which I thought was quite excessive. if we haven't go a connection in 5 seconds we arn't getting one

jonnyshaw89:
This is trying to do too much in 1 PR

PR #63, opened 2018-11-30
Pass the knex connection to the repositories

jonnyshaw89:
This is the correct way to do this rather then create a new connection pool per repository

### How to review 
Have a look at the implementation, Ive only made this change for 1 Repository so far. But open to feedback

### DIFF
--- a/eq-author-api/config/knexfile.js
+++ b/eq-author-api/config/knexfile.js
@@ -1,35 +1,11 @@
 module.exports = {
-  development: {
-    client: "postgresql",
-    connection: process.env.DB_CONNECTION_URI
+  client: "postgresql",
+  connection: process.env.DB_CONNECTION_URI,
+  pool: {
+    min: 10,
+    max: 50
   },
-
-  test: {
-    client: "postgresql",
-    connection: process.env.DB_CONNECTION_URI
-  },
-
-  staging: {
-    client: "postgresql",
-    connection: process.env.DB_CONNECTION_URI,
-    pool: {
-      min: 2,
-      max: 10
-    },
-    migrations: {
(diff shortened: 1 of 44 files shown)


tgandrews:
Doesn't look too bad but this is going to have a big impact on existing PRs so we need to manage how we do it.

tgandrews:
@jonnyshaw89 do you have prettier installed? There are a lot of formatting changes in this.

eq-author-api/config/knexfile.js
```diff
@@ -1,12 +1,20 @@
...
   },
 
   test: {
     client: "postgresql",
-    connection: process.env.DB_CONNECTION_URI
+    connection: process.env.DB_CONNECTION_URI,
+    pool: {
+      min: 2,
+      max: 10
+    }
```
samiwel: Can we do away with separate configurations or is one specifically required for test?
If they're the same maybe they can be combined into one config:

```js
const baseDbConfig = {
	client: "postgresql",
    connection: process.env.DB_CONNECTION_URI,
    pool: {
      min: 2,
      max: 10
    }
}

module.exports = {
	development: baseDbConfig,
    test: baseDbConfig,
    production: {
	  ...baseDbConfig,
      pool: {
        min: 2,
        max: 20
      }
   }
}

```

jonnyshaw89: 👍 already done

jonnyshaw89:
@samiwel @tgandrews Can you guys have another look over this one please?

adammahmood:
Is this really necessary?

https://stackoverflow.com/questions/49588672/knex-js-sql-knex-sql-connection-pools

PR #67, opened 2018-12-03
754 full end to end test expanded to include routing on radio and number answers 4

SamGodwin2:
This Pr is part of the setup work for the routing refactor that will be taking place this sprint. It add a end-2-end test that will test the ui, publisher and the runner implementation of routing and help us to ensure that all functionality is preserved after this routing refactor.

### How to review 
Test looks comprehensive and all passes.

### Test is currently broken due to a bug that is being fixed in: https://github.com/ONSdigital/eq-author-app/pull/53

### DIFF
--- a/eq-author/cypress/builders/answer.js
+++ b/eq-author/cypress/builders/answer.js
@@ -1,23 +1,50 @@
 import { testId } from "../utils";
 
-const configureCheckbox = ({ options }) => {
+const configureMultipleChoice = ({ initialCount }, { options }) => {
   options.forEach(({ label }, index) => {
-    cy.get(testId("option-label")).should("have.length", index + 1);
+    cy.get(testId("option-label")).should(
+      "have.length",
+      Math.max(index + 1, initialCount)
+    );
     cy.get(testId("option-label"))
-      .last()
+      .eq(index)
       .type(label)
       .blur();
-    const isLast = index + 1 === options.length;
-    if (!isLast) {
+    const nextIndex = index + 1;
+    const shouldAddOne =
+      initialCount <= nextIndex && options.length > nextIndex; // index + 1 === options.length;
+    if (shouldAddOne) {
       cy.get(testId("btn-add-option")).click();
     }
   });
 };
 
+const configueBasicAnswer = ({ label, description }) => {
+  if (label) {
+    cy.get(testId("txt-answer-label")).type(label);
+  }
+  if (description) {
+    cy.get(testId("txt-answer-description")).then(
+      $el =>
+        $el
+          ? cy.wrap($el).type(description)
+          : new Error(`Answer type doesn't support description`)
+    );
+  }
+};
+
 const configure = config => {
   switch (config.type.toLowerCase()) {
     case "checkbox":
-      configureCheckbox(config);
+      configureMultipleChoice({ initialCount: 1 }, config);
+      break;
+    case "radio":
+      configureMultipleChoice({ initialCount: 2 }, config);
+      break;
+    case "number":
+    case "currency":
+    case "textfield":
+      configueBasicAnswer(config);
       break;
     default:
       throw new Error(`Answer type not supported: ${config.type}`);
--- a/eq-author/cypress/routingTests/routing_e2e_readme.md
+++ b/eq-author/cypress/routingTests/routing_e2e_readme.md
@@ -1,0 +1,43 @@
+# Routing E2E Test
+
+This test spec file was created as part of the setup work for the routing refactor. It is designed to create a simple survey that utilizes all
+features offered by the current routing implementation. This test runs across the entire EQ structure and asserts that the route is correct
+by following it though on EQ-Survey-Runner.
+
+The layout of the survey is as follows:
+
+```
+                                   +-------------------+
+                                   | Section 1. Page 1 |
+                                   +----------+--------+
+                                   |          |        |
+           Chose A                 |          |        |                         Chose B
+           +-----------------------+          |        +--------------------------------+
+           |                           Not Answered                                     |
+           |                                  |                                         |
+           v                                  v                                         v
+
++-----------------+                 +-----------------+                       +--------------------+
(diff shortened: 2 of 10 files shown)


eq-author/cypress/routingTests/routing_e2e_readme.md
```diff
@@ -0,0 +1,43 @@
+# Routing E2E Test
+
+This test spec file was created as part of the setup work for the routing refactor. It is designed to create a simple survey that utilizes all
```
adammahmood: Do we plan on keeping these tests after the refactor?

eq-author/cypress/builders/answer.js
```diff
@@ -1,23 +1,50 @@
...
   });
 };
 
+const configueBasicAnswer = ({ label, description }) => {
+  if (label) {
+    cy.get(testId("txt-answer-label")).type(label);
+  }
+  if (description) {
+    cy.get(testId("txt-answer-description")).then(
+      ele =>
```
adammahmood: Why `ele`?

eq-author/cypress/builders/answer.js
```diff
@@ -1,23 +1,50 @@
 import { testId } from "../utils";
 
-const configureCheckbox = ({ options }) => {
+const configureMultipleChoice = initialCount => ({ options }) => {
```
adammahmood: Why not pass this as an object shaped as `{initialCount, options}` ?

eq-author/cypress/routingTests/routing_e2e_spec.js
```diff
@@ -0,0 +1,305 @@
...
+      }
+    });
+  });
+
+  it("adds radio routing", () => {
+    navigateToPage("Section 1, Page 1");
+    navigateToRoutingTab();
+    cy.get(testId("btn-add-rule")).click();
+
+    cy.get(testId("routing-editor")).within(() => {
```
adammahmood: Some of these tests are shared with `routing_spec.js`.  Can we share them?

eq-author/cypress/routingTests/routing_e2e_spec.js
```diff
@@ -0,0 +1,305 @@
...
+          cy.get(testId("comparator-selector")).select("GreaterThan");
+          cy.get(testId("number-input")).type("5");
+          findByLabel("THEN")
+            .last()
+            .select("End of Questionnaire");
+        });
+    });
+  });
+
+  describe("Runner Side", () => {
```
adammahmood: Can we rename this to something more descriptive? Maybe just `Runner`?

eq-author/cypress/routingTests/routing_e2e_readme.md
SamGodwin2: I wasn't planning on it as they're unlikely to be very reliable as they depend on Runner which could change without our knowledge?

eq-author/package.json
```diff
@@ -23,7 +24,8 @@
     "precommit": "lint-staged",
     "coverage": "yarn test -- --coverage",
     "cypress:open": "CYPRESS_baseUrl=${CYPRESS_baseUrl:=http://localhost:3000/} cypress open",
-    "cypress:open:e2e": "CYPRESS_baseUrl=${CYPRESS_baseUrl:=http://localhost:13000/} cypress open --config integrationFolder=cypress/e2e"
+    "cypress:open:e2e": "CYPRESS_baseUrl=${CYPRESS_baseUrl:=http://localhost:13000/} cypress open --config integrationFolder=cypress/e2e",
+    "cypress:open:routing": "CYPRESS_baseUrl=${CYPRESS_baseUrl:=http://localhost:3000/} CYPRESS_chromeWebSecurity='false' cypress open --config integrationFolder=cypress/routingTests"
```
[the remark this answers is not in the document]
SamGodwin2: It also relies on having a local version of the app/api/publisher running as well, and we don't spin them up in the other cypress tests so as long as it's detailed in the readme what you need running it seems okay? I'll double check the readme and make sure that's clear though.

eq-author/cypress/routingTests/routing_e2e_spec.js
SamGodwin2: I didn't really want to share anything here except for the utils as we wanted these tests to be very specific and isolated from the other routing implementation.

SamGodwin2:
@adammahmood think the issue with the runner side is due to the api not being spun up using `RUNNER_SESSION_URL="http://localhost:5000/session?token="` as for some reason cypress doesn't play nice with `docker.for.mac`?

eq-author/cypress/builders/answer.js
SamGodwin2: Just the name I gave to the Jquery element the `cy.get` returns, got a clearer name?

eq-author/cypress/routingTests/routing_e2e_spec.js
tgandrews: We could move the actions to builders? So we can reuse the same functionality in both places

eq-author/cypress/routingTests/routing_e2e_readme.md
tgandrews: Although the team restructure may be stopping that as runner is code frozen?

eq-author/cypress/builders/answer.js
tgandrews: `$el` was the jQuery convention IIRC

eq-author/cypress/routingTests/routing_e2e_readme.md
tgandrews: The publisher output sounds like a good idea once we have proven the refactor works.

eq-author/package.json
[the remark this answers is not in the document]
tgandrews: The cypress dockerfile includes publisher and validator which is probably fine as thats the limit or our remit. But including runner would seem like too much.

PR #72, opened 2018-12-05
Update webpack version range

adammahmood:
Test greenkeeper will update lock files for in-range packages

### How to review 
- Tests pass
- See if https://github.com/ONSdigital/eq-author-app/pull/68 has its lock file updated after this branch is merged


jonnyshaw89:
@adammahmood Didn't update the lock file did you

PR #73, opened 2018-12-05
Upgrade package ranges to include latest minor release within current…

adammahmood:
Upgrade package ranges to include latest minor release within current major versions so greenkeeper will automatically update yarn.lock files for these package upgrades.

### How to review 
- Run tests


adammahmood:
There seems to be an issue when deleting entities such as answers and pages.  We will need to investigate further and do a bit of refactoring before merging this. Some of the other greenkeeper PR's are failing for the same reason.  There's some sort of breaking change that's been introduced in a non-major release in one of our dependencies (maybe Apollo related) 😠.

eq-author/package.json
```diff
@@ -66,26 +66,26 @@
...
+    "eslint-loader": "^2.1.1",
     "eslint-plugin-graphql": "^1.5.0",
-    "file-loader": "2.0.0",
-    "fs-extra": "7.0.0",
+    "file-loader": "^2.0.0",
+    "fs-extra": "^7.0.0",
     "gh-pages": "^1.0.0",
     "gzip-size": "^4.0.0",
-    "html-webpack-plugin": "4.0.0-alpha.2",
+    "html-webpack-plugin": "^4.0.0-alpha.2",
```
samiwel: Should we be depending on an alpha release?

tgandrews: It is required for webpack 4. Bit rubbish, but there is now a beta which this resolves to in the `yarn.lock` file. Also I have a greater tolerance for alpha/beta build tools as our tests run on the output.

PR #74, opened 2018-12-06
Ability to retrieve DB credentials from AWS SecretsManager

jonnyshaw89:
This adds the ability to load the Database credentials from AWS SecretsManager by passing an environment variable of `DB_SECRET_ID`

### How to review 
Launch on AWS and configure secrets manager with your RDS credentials


eq-author-api/db/index.js
```diff
@@ -1,3 +1,37 @@
...
+      const secret = await client
+        .getSecretValue({ SecretId: secretId })
+        .promise();
+      const dbCredentials = JSON.parse(secret.SecretString);
+      result.connection = {
+        host: dbCredentials.host,
+        port: dbCredentials.port,
+        user: dbCredentials.username,
+        password: dbCredentials.password,
+        database: dbCredentials.dbname
```
tgandrews: It looks like we are just naming dbname to datbase. Can just do that and spread the rest?

eq-author-api/app.js
```diff
@@ -20,38 +20,45 @@ const pino = pinoMiddleware();
...
-  app.post("/import", bodyParser.json({ limit: "50mb" }), importAction);
-}
-
-if (process.env.NODE_ENV === "development") {
+const db = require("./db");
+
+db(process.env.DB_SECRET_ID).then(async conf => {
+  let knex = require("knex")(conf);
+  logger.log("Running Migrate");
+  await knex.migrate.latest();
```
tgandrews: 👍

eq-author-api/db/index.js
```diff
@@ -1,3 +1,37 @@
...
+      };
+    } catch (error) {
+      logger.log("Failed to load DB credentials from AWS SecretsManager");
+      process.exit();
```
tgandrews: We should exit with a non-zero exit code to show it’s a process error. Can we also log the error details.

eq-author-api/app.js
```diff
@@ -20,38 +20,45 @@ const pino = pinoMiddleware();
...
+      pino,
+      cors(),
+      graphiqlExpress({ endpointURL: "/graphql" })
+    );
+  }
+
+  app.listen(PORT, "0.0.0.0", () => {
+    console.log(colors.green("Listening on port"), PORT);
+  });
 });
```
tgandrews: I think we should add a catch to and exit the process with the error otherwise IIRC they will end as unhandledRejections which currently don’t crash the process

eq-author-api/config/knexfile.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
     min: 10,
     max: 50
   },
+  acquireConnectionTimeout: 5000,
```
tgandrews: Can we remove the waitforpostgres stuff with this? I think it’s a dev and ci only issue so it should be straightforward to test

eq-author-api/db/index.js
jonnyshaw89: How do I do that?

samiwel:
```js
const { dbName, ...otherProps } = dbCredentials;

result.connection = {
  database: dbName,
  ...otherProps
}
```

eq-author-api/repositories/MetadataRepository.test.js
```diff
@@ -32,8 +28,20 @@ const buildMetadata = (questionnaireId, json = {}) => {
...
 describe("MetadataRepository", () => {
   let questionnaireId;
-
-  beforeAll(() => knex.migrate.latest());
+  let knex;
+  let QuestionnaireRepository;
+  let MetadataRepository;
+
+  beforeAll(async () => {
+    await db(process.env.DB_SECRET_ID).then(async conf => {
```
tgandrews:
```suggestion
    const conf = await db(process.env.DB_SECRET_ID);
```

eq-author-api/db/index.js
```diff
@@ -1,3 +1,36 @@
...
+        .promise();
+      const dbCredentials = JSON.parse(secret.SecretString);
+      const { dbName, ...otherProps } = dbCredentials;
+
+      result.connection = {
+        database: dbName,
+        ...otherProps
+      };
+    } catch (error) {
+      logger.log("Failed to load DB credentials from AWS SecretsManager");
```
tgandrews: Can we log the error

eq-author-api/config/knexfile.js
jonnyshaw89: This is a connection timeout rather than a "retry for"

jonnyshaw89:
@adammahmood Shouldn't have any impact locally as it defaults to running existing config when `process.env.DB_SECRET_ID` isn't set

jonnyshaw89:
@tgandrews can I get a re-review?

eq-author-api/app.js
```diff
@@ -20,38 +19,50 @@ const pino = pinoMiddleware();
...
+      app.use(
+        "/graphiql",
+        pino,
+        cors(),
+        graphiqlExpress({ endpointURL: "/graphql" })
+      );
+    }
+
+    app.listen(PORT, "0.0.0.0", () => {
+      logger.log("Listening on port", PORT);
```
tgandrews: It doesn't seem to log the `PORT` so the api to our previous logging is different.

eq-author-api/package.json
```diff
@@ -33,14 +31,12 @@
...
-    "jest": "^23.6.0",
     "js-yaml": "^3.12.0",
     "json-web-key": "^0.4.0",
     "jsrsasign": "^8.0.12",
     "knex": "^0.15.2",
     "lodash": "^4.17.4",
     "node-jose": "^1.0.0",
-    "nodemon": "^1.11.0",
     "pg": "^7.4.1",
     "pg-connection-string": "^2.0.0",
```
tgandrews: Can this dependency be removed as well?

eq-author-api/app.js
```diff
@@ -20,38 +19,50 @@ const pino = pinoMiddleware();
...
-    cors(),
-    graphiqlExpress({ endpointURL: "/graphql" })
-  );
-}
-
-app.listen(PORT, "0.0.0.0", () => {
-  console.log(colors.green("Listening on port"), PORT);
-});
+  .catch(error => {
+    logger.log(error);
```
tgandrews: Can we provide the log type?

eq-author-api/yarn.lock
```diff
@@ -101,74 +87,62 @@ ajv@^6.5.3:
...
 
 ansi-escapes@^1.0.0:
   version "1.4.0"
   resolved "https://registry.yarnpkg.com/ansi-escapes/-/ansi-escapes-1.4.0.tgz#d3a8a83b319aa67793662b13e761c7911422306e"
-  integrity sha1-06ioOzGapneTZisT52HHkRQiMG4=
 
 ansi-escapes@^3.0.0:
   version "3.1.0"
   resolved "https://registry.yarnpkg.com/ansi-escapes/-/ansi-escapes-3.1.0.tgz#f73207bb81207d75fd6c83f125af26eea378ca30"
-  integrity sha512-UgAb8H9D41AQnu/PbWlCofQVcnV4Gs2bBJi9eZPxfU/hgglFh3SMDMENRIqdr7H6XFnXdoknctFByVsCOotTVw==
```
tgandrews: All these integrity shas have been removed. I think you need a later version of yarn to get these back. `yarn global upgrade yarn` will upgrade it for you.

eq-author-api/middleware/import.js
```diff
@@ -1,3 +1,7 @@
+const pinoMiddleware = require("express-pino-logger");
+const pino = pinoMiddleware();
+const logger = pino.logger;
```
samiwel: Not sure if this is the best way to get a hold of the logger, even though it works. This seems to be instantiating express middleware, but the middleware is not being passed into the express handler functions, so it seems a bit wasteful. @tgandrews any ideas if it's possible to extract the logger from expresses middleware somehow? Otherwise feels like the logger is something that we should be exporting in the root `app.js` and passing down (or making available) to the other modules by means of a `getLogger` function for example?

tgandrews: Good spot, I didn't notice this change. According to the docs it looks like the middleware exposes `req.log` to get to the logger so we shouldn't have to import it here.

PR #78, opened 2018-12-07
Bug - Unable to pipe dates

tgandrews:
This was caused by us filtering out the date answers from the answers
you could pipe.

This change also fixes an issue where the get previous answers would
be returned in a non-deterministic order. So now we order in the db so
we can guarantee the order of the answers.

### How to review 
1. Ensure you can pipe dates anywhere you can pipe in the app.

### DIFF
--- a/eq-author-api/constants/pipingAnswerTypes.js
+++ b/eq-author-api/constants/pipingAnswerTypes.js
@@ -1,3 +1,9 @@
-const { TEXTFIELD, NUMBER, CURRENCY, DATE_RANGE } = require("./answerTypes");
+const type = require("./answerTypes");
 
-module.exports.PIPING_ANSWER_TYPES = [TEXTFIELD, NUMBER, CURRENCY, DATE_RANGE];
+module.exports.PIPING_ANSWER_TYPES = [
+  type.TEXTFIELD,
+  type.NUMBER,
+  type.CURRENCY,
+  type.DATE,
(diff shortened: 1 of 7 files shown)


eq-author-api/constants/pipingAnswerTypes.js
```diff
@@ -1,3 +1,8 @@
-const { TEXTFIELD, NUMBER, CURRENCY, DATE_RANGE } = require("./answerTypes");
+const type = require("./answerTypes");
```
adammahmood: Why not destructure the types?

adammahmood: Was the bug purely caused by wrong answer types being set here?

tgandrews: Yes, it was just the wrong type for piping. It should be `DATE` not `DATE_RANGE`.

I preferred not to destructure just to save typing as you have to state each type twice in this context. But it's just personal preference so I'm happy to change if others prefer the other way.

samiwel:
Seems like we can only pipe in the first answer on `DateRange` answers. I set up the date range with `From` and `To` labels. When piping, I only see `From`. Appreciate this PR was nothing to do with this, so maybe let's create another issue off the back of this one to fix?
image

tgandrews:
@samiwel I guess this is to with `displayName` calculation for DateRange answers?

PR #81, opened 2018-12-10
Authenticate users on the API using Firebase token

samiwel:
> This is still WIP but PR raised for early visibility.

This PR passes the user's token in the header of GraphQL API requests. Any request containing an invalid/non-existent token will get a 401 response from the API. Going forward, developers will need to create their own tokens (e.g. by setting up a Firebase project) in order to use the API.

A new query has been introduced on the root query type in order to get details about the currently logged in user. This can be queried using the `me` query which either returns a valid logged in user, or null (i.e. 401).

Finally, the createQuestionnaire mutation has been updated so that the `createdBy` field is not passed in with the query from the client, but rather the API assigns the createdBy field via the current authentication token.

### How to review 
Should be able to log in and use the app as normal. Please report any errors you encounter.

** GraphiQL does not seem to allow you to pass in authentication headers with the request so if you want to explore the API using an explorer client, I can recommend using Altair client which is available as a chrome extension and allows for setting the authentication header with your user's token. (Token may be found by examining the localStorage after logging in.) **

There is still some work to do updating the tests etc. I will continue working on this in the meantime.


eq-author/src/redux/auth/actions.js
```diff
@@ -4,14 +4,23 @@ export const SIGN_IN_USER = "SIGN_IN_USER";
 export const SIGN_OUT_USER = "SIGN_OUT_USER";
 export const SIGN_IN_ERROR = "SIGN_IN_ERROR";
 
-export const signInUser = ({ displayName, email, photoURL }) => {
+export const signInUser = ({
+  displayName,
+  email,
+  photoURL,
+  stsTokenManager: { accessToken },
```
jonnyshaw89: What is `stsTokenManager` used for?

jonnyshaw89: Oh this like makes the json onto parameters

jonnyshaw89: `getIdToken` may tie us in with Firebase so might be worth handling this at a later date in a more generic way.

eq-author/src/components/Header/index.js
```diff
@@ -117,6 +118,7 @@ export class UnconnectedHeader extends React.Component {
 
   render() {
     const { questionnaire } = this.props;
+    const currentUser = get("data.me", this.props);
```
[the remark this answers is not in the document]
samiwel: No particular reason other than `get` is safe in the event of a GraphQL error where `data` would be null/undefined.

Will destructing work in the event that a GraphQL error is received? If so, I'm happy to change it.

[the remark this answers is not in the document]
samiwel: Actually just realised that checking that we have no error already happens a level above this. I'll make the change.

samiwel:
[Use firebase token for determining identity of author.](https://trello.com/c/vM1r8ltz/753-use-firebase-token-for-determining-identity-of-author)

eq-author-api/middleware/auth.js
```diff
@@ -0,0 +1,29 @@
...
+  const accessToken = authHeader.replace("Bearer ", "");
+  if (isNil(accessToken)) {
+    logger.error("Request must contain a valid access token.");
+    res.send(401);
+    return;
+  }
+
+  const jwtToken = jwt.decode(accessToken);
+  if (isNil(jwtToken)) {
+    logger.error("Could not decode JWT token.");
```
tgandrews:
```suggestion
     logger.error("Could not decode token.");
```

The T in JWT is token so this says "token token"

eq-author-api/schema/resolvers.js
```diff
@@ -191,7 +193,10 @@ const Resolvers = {
   Questionnaire: {
     sections: (questionnaire, args, ctx) =>
       ctx.repositories.Section.findAll({ questionnaireId: questionnaire.id }),
-    createdBy: questionnaire => ({ name: questionnaire.createdBy }),
+    createdBy: questionnaire => ({
+      id: questionnaire.createdBy, // Temporary until next PR introduces users table.
```
tgandrews: Can't this use `ctx.auth.name` in this PR?

samiwel: It could I guess, but I'm not sure that makes sense and I don't think it would give you much. The auth token would contain a potentially different name to the one stored against the questionnaire. So you'd end up with something like:

```
createdBy: {
id: "Tom Andrews",
name: "Sam Thomas"
```

The next steps would be to actually store the user in a `users` table. When we get to that stage, we'd resolve the user details from that table and store a foreign key reference on the questionnaire.

samiwel: An `id` field is required for Apollo caching

tgandrews: Yes sorry this is wrong. I misread it as the create mutation its the field read. Sorry.

jonnyshaw89:
When I load Author while not logged in it sends a request to the API. Either it should not send this request as it will know it wont work as i have no token.
Or it should remove the header from the the request
screenshot 2018-12-19 at 12 02 55
My preference is to not send the request as we know it wont work, Otherwise we get 401 errors from the api all the time

eq-author/src/apollo/createApolloErrorLink.js
```diff
@@ -2,8 +2,8 @@ import { onError } from "apollo-link-error";
 import { apiDownError } from "../redux/saving/actions";
 
 export default getStore =>
-  onError(({ networkError }) => {
-    if (networkError) {
+  onError(({ networkError, operation: { operationName } }) => {
+    if (networkError && operationName !== "GetCurrentUser") {
```
jonnyshaw89: If we don't send the request if we have no token then we shouldn't need this

samiwel: The reason for sending the request is because of the network only fetch policy. This was to avoid displaying the users avatar even after they have logged off (since Apollo caches the user locally in its store). I think this could be avoided by using the presence of an Auth token to determine if the user avatar should be displayed, I'll take a look tomorrow.

samiwel:
Regarding the Snyk vulnerability, there is no fix at present. The Snyk check does not block merging, so I think we should be pragmatic and not allow the vulnerability report to block our work, but seek to upgrade to a version that fixes the vulnerability when a fix becomes available.

jonnyshaw89:
@samiwel so Pico logs out the token. Are we concerned about this? its not ideal really

eq-author-api/middleware/auth.test.js
```diff
@@ -0,0 +1,117 @@
...
+      res = {
+        send: jest.fn()
+      };
+      next = jest.fn();
+
+      middleware = createAuthMiddleware(logger, context);
+    });
+
+    it("should check the authorization header", () => {
+      req.header.mockImplementation(() => null);
```
tgandrews: You can use `mockReturnValue` instead but I think you'll get this by default with your empty `jest.fn()`

samiwel:
@jonnyshaw89 regarding your comment about logging the JWT token, I was reading the pino docs yesterday and think this can be avoided by passing redacted settings into the middleware's constructor function. You just need to pass the key to be redacted from the logs e.g `req.headers.authorization` or however it appears in the logs right now.

Default behaviour is to mask the value but you can pass `remove: true` to remove the value altogether. Happy if you want to make the change to this PR or raise a separate PR to address?

tgandrews:
@samiwel @jonnyshaw89  As this is only going to staging can we address this in a later PR?

PR #88, opened 2018-12-14
Fix error logging by using correct method

adammahmood:
GraphQL errors are no longer returning useful information as we are using a non-existent method from `pino` logger.

### How to review 
- Run tests
- Ensure invalid GraphQL request returns a useful message.

Eg.
```
query GetQuestionPage($id: ID!) {
  questionPage(id: $id) {
    foobar
  }
}
```

Should return:
```
{
  "errors": [
    {
      "message": "Cannot query field \"foobar\" on type \"QuestionPage\".",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ]
    }
  ]
}
```

Prior to this PR the following would be returned and we would see an error logged in the API:
```
{
  "errors": [
    {
      "message": "Internal server error"
    }
  ]
}
```

```
Error in formatError function: TypeError: Cannot read property 'Symbol(pino.write)' of undefined
web_1  |     at LOG (/app/node_modules/pino/lib/tools.js:39:11)
web_1  |     at /app/node_modules/apollo-server-core/dist/runQuery.js:31:24
web_1  |     at Array.map (<anonymous>)
web_1  |     at format (/app/node_modules/apollo-server-core/dist/runQuery.js:28:19)
web_1  |     at doRunQuery (/app/node_modules/apollo-server-core/dist/runQuery.js:115:25)
web_1  |     at /app/node_modules/apollo-server-core/dist/runQuery.js:21:56
web_1  |     at process._tickCallback (internal/process/next_tick.js:68:7)
```


jonnyshaw89:
Pino logger doesn't have a `.log` method on its api? 
https://getpino.io/#/docs/api

adammahmood:
@jonnyshaw89 Your right.  The method doesn't exist and we haven't setup it up as a custom level either (I assumed we had and was reverting back to the method we used a few commits ago).  Looks like I can just remove the `formatError` line completely and we achieve the same results.

PR #89, opened 2018-12-18
Add min and max duration validation to date range answers

adammahmood:
Adds min and max duration validation to date range answers.  This enforces users in `eq-survey-runner` to enter a date within the specified duration period. 

### How to review 
- Ensure test pass
- Create a date range answer and add a min/max duration validation rule and ensure `eq-survey-runner` shows appropriate validation messages when trying to submit data outside of the min/max duration rules set

### DIFF
--- a/eq-author-api/schema/typeDefs.js
+++ b/eq-author-api/schema/typeDefs.js
@@ -218,7 +218,7 @@ type NumberValue {
 
 union RoutingConditionValue = IDArrayValue | NumberValue
 
-union ValidationType = NumberValidation | DateValidation
+union ValidationType = NumberValidation | DateValidation | DateRangeValidation
 
 enum ValidationRuleEntityType {
   Custom
@@ -235,6 +235,13 @@ type NumberValidation {
 type DateValidation {
   earliestDate: EarliestDateValidationRule!
   latestDate: LatestDateValidationRule!
+}
+
+type DateRangeValidation {
+  earliestDate: EarliestDateValidationRule!
+  latestDate: LatestDateValidationRule!
+  minDuration: MinDurationValidationRule!
+  maxDuration: MaxDurationValidationRule!
 }
 
 interface ValidationRule {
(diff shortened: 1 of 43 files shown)


adammahmood:
[Validation for date range - Min duration and max duration (9)](https://trello.com/c/Qr9lDfQw/407-validation-for-date-range-min-duration-and-max-duration-9)

eq-author-api/schema/typeDefs.js
```diff
@@ -285,6 +292,18 @@ type LatestDateValidationRule implements ValidationRule {
   availableMetadata: [Metadata!]!
 }
 
+type MinDurationValidationRule implements ValidationRule {
+  id: ID!
+  enabled: Boolean!
+  offset: Duration!
```
tgandrews: I dislike using the an offset here. It isn't an offset. It is more like min or max value.

eq-author/src/components/Validation/Date/DatePreview.js
```diff
@@ -45,9 +44,10 @@ const DatePreview = ({
   return (
     <React.Fragment>
       <div>
-        {`${value} ${unit.toLowerCase()} ${relativePosition.toLowerCase()}:`}
+        {`${value} ${unit.toLowerCase()} `}
+        {relativePosition && `${relativePosition.toLowerCase()}:`}
       </div>
-      <div>{rule[entityType]}</div>
+      <div>{entityType && rule[entityType]}</div>
```
tgandrews: Do we want an empty div when its not needed?

eq-author/src/components/Validation/Date/builder.js
```diff
@@ -45,13 +45,13 @@ export const readToWriteMapper = outputKey => ({
   ...rest
 }) => ({
   id,
-  [outputKey]: {
+  [outputKey]: omitBy(isNil)({
```
tgandrews: :+1: Nice pattern.

eq-author-api/schema/typeDefs.js
adammahmood: It's definitely a `Duration` which we already described in our schema as:

```
type Duration {
  value: Int
  unit: DurationUnit!
}

enum DurationUnit {
  Days
  Months
  Years
}
```
What if we renamed `offset` to `duration`?

```
type MinDurationValidationRule implements ValidationRule {
  id: ID!
  enabled: Boolean!
  duration: Duration!
```

eq-author/src/components/Validation/Date/DatePreview.js
adammahmood: No. Ideally, we should keep DOM elements to a minimum.  I'll rework this.

eq-author-api/schema/typeDefs.js
tgandrews: :+1: Yes, that looks much better

PR #90, opened 2018-12-19
Separate implementation of Routing to support new structure on the API.

SamGodwin2:
This PR contains all the new modifiers, repositories, and resolvers for for the routing MK.2 implementation as inspired by the https://github.com/ONSdigital/eq-author-app/wiki/Routing,-MK2 wiki page. [Basic explanation of the structure explained](https://github.com/ONSdigital/eq-author-app/blob/755-should-have-a-parallel-and-isolated-routing-implementation/docs/architecture/0002-routing-mk2.md) 

In an attempt to simplify this PR we have reduced the scope to only match the required functionality of the current system.

### How to review 
Tests pass, and new endpoints all work as expected.


tgandrews:
[Should have a parallel and isolated routing implementation (16)](https://trello.com/c/mhQCfgUN/755-should-have-a-parallel-and-isolated-routing-implementation-16)

docs/architecture/0002-routing-mk2.md
```diff
@@ -0,0 +1,109 @@
...
+                condition: "Equals",
+                right: {
+                  custom: "Wales"
+                }
+              }
+            }
+          ]
+        },
+        destination: {
+          title: "Page 2"
```
rmccar: Shouldn't this be "Page 3"?

docs/architecture/0002-routing-mk2.md
```diff
@@ -0,0 +1,109 @@
...
+        }
+      },
+    ]
+    else {
+      page: {
+        title: "Page 3"
```
rmccar: And this be "Page 4"?

docs/architecture/0002-routing-mk2.md
```diff
@@ -0,0 +1,109 @@
+# 2. Routing Mk2
```
rmccar: Would maybe a table of contents would be good? With links to each section as I've done in other mds. I know these files aren't that long but could be useful if they became longer maybe? But if you don't think it worth it don't worry

tgandrews: The file format is generated by the adr tools so I would rather keep to that.

docs/architecture/0002-routing-mk2.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Decision
+
+Implement a new version of the routing in the API that can handle nested expressions.
+
+This is heavily inspied by https://github.com/ONSdigital/eq-author-app/wiki/Routing,-MK2
+
+### Terminology
+- Routing - all rules etc for a page with a fallback destination when no rules match.
+- Rule - a series of expressions that when evaluated to true will go to the destination specified.
```
samiwel: More of a question than anything else but do all rules have at least one `ExpressionGroup`? The example structure below has an ANDed group even though it's really a single expression?

SamGodwin2: Yeah a rule has to have at least one group, which is just set to AND by default it just won't have any effect. Saves having to do different things depending on whether there's one expression or two or more expressions.

eq-author-api/app.js
```diff
@@ -1,14 +1,16 @@
...
 const bodyParser = require("body-parser");
-const schema = require("./schema");
 const pinoMiddleware = require("express-pino-logger");
+
 const { PORT } = require("./config/settings");
 const status = require("./middleware/status");
 const { getLaunchUrl } = require("./middleware/launch");
 const createAuthMiddleware = require("./middleware/auth");
+const repositories = require("./repositories");
+const modifiers = require("./modifiers");
```
samiwel: Granted I haven't looked through them all yet, but from the first couple of examples that I've seen these modifiers don't appear to be "modifying" anything. Rather they appear to be orchestrating repositories, performing validation and calling other modifiers. I think `services` is a more conventional name for this type of entity, to me the name modifiers is a bit misleading and confusing. But I might be missing something that I haven't got to yet.

tgandrews: We did have a bit of a discussion about the name and it could probably be improved. However, they are always used for performing changes; creation, update and updating based on other events. They are never used for reading which is done directly against the repositories (at the moment). Where as a service I would view that as something that is completely isolated and completely handles that entity.

Also, a side note, service is an overloaded term that has multiple meaning including; separate network app or an OS managed daemon.

samiwel: Fair enough. I can live with it https://martinfowler.com/bliki/TwoHardThings.html

eq-author-api/schema/resolvers/routing2/binaryExpression2/index.js
```diff
@@ -0,0 +1,86 @@
+const answerTypes = require("../../../../constants/answerTypes");
```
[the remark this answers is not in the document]
SamGodwin2: Don't think we can use absolute paths on the api as there's no webpack building it?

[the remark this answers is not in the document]
samiwel: I think we can use NODE_PATH env var, but maybe let's look into this separately in another PR.

samiwel:
@SamGodwin2 LGTM can you squash the commits?

PR #98, opened 2018-12-28
Reorganise component folder structure

adammahmood:
Reorganises UI components folder structure as per discussion https://github.com/ONSdigital/eq-author-app/pull/58

### How to review 
- Ensure tests pass
- Ensure folder structure matches structure from discussions https://github.com/ONSdigital/eq-author-app/pull/58


adammahmood:
I tried sticking to the concept of having two main folder structures:
1) `src/components`- This contains app independent components (essentially these _could_ be pulled into any project and utilised)
2) `src/App` - This contains app specific components organised by feature.  Some components are shared and live in `src/App/components`.  I'm not too sure if this is the best name for these 'shared' components.  What do others think?

adammahmood:
[UI Structure (2 days)](https://trello.com/c/otdt0EY8/756-ui-structure-2-days)

adammahmood:
Codacy is failing due to:  `The numeric literal ''{0}'' will have a different value at runtime.` I think we can turn this rule off.  I don't think it protects us from much.  

The actual code it's failing on specifically tests how large / out of range values are handled. I can't seem to find a way to ignore just those lines and would rather not ignore the entire file.  

What does everyone else think?

hamishtaplin:
@adammahmood As discussed on Slack I don't think we need the App/components folder—everything in there would just go in /components.

tgandrews:
I have been having a look and it seems quite a bit different than discussed as it is not components. A lot of the time it seems to be flat folders that you have to know where they are imported. The idea was to be a parent folder would include all the children with the exception of component library folders.

Maybe we can discuss this offline?

Issue #92, opened 2018-12-20
Min and Max value validation input fields trigger requests onChange rather than onBlur


adammahmood:
### Expected behaviour
Min and Max value validation input fields should only trigger an API request onBlur.

### Actual behaviour
Currently, when entering values into these fields, characters are sometimes missed when typing quickly.

### Steps to reproduce the behaviour
Quickly type characters in the fields.

PR #105, opened 2019-01-07
Refactor answer validation to use a consistent pattern

adammahmood:
- Refactors  answer validation to use a consistent pattern and utilise `withEntityEditor` for state management for all validation types
- Removes duplicate methods by replacing with re-usable HOF's
- Fixes https://github.com/ONSdigital/eq-author-app/issues/92 

### How to review 
- Ensure tests pass
- Ensure https://github.com/ONSdigital/eq-author-app/issues/92 is fixed


adammahmood:
[Fix min and max value validation for number triggering requests onChange rather than onBlur](https://trello.com/c/1eLX0f6S/783-fix-min-and-max-value-validation-for-number-triggering-requests-onchange-rather-than-onblur)

adammahmood:
Codacy is failing due to: The numeric literal ''{0}'' will have a different value at runtime. I think we can turn this rule off. I don't think it protects us from much.

The actual code it's failing on specifically tests how large / out of range values are handled. I can't seem to find a way to ignore just those lines and would rather not ignore the entire file.

What does everyone else think? 

This codacy issue was also raised in https://github.com/ONSdigital/eq-author-app/pull/98 as well. I'm not too sure how that PR passed it's checks in the end 😕

tgandrews:
@adammahmood maybe we just remove codacy?

eq-author/src/App/questionPage/Design/Validation/AnswerValidation.js
```diff
@@ -9,19 +9,23 @@ import { gotoTab } from "redux/tabs/actions";
 
 import SidebarButton, { Title, Detail } from "components/buttons/SidebarButton";
 import ModalWithNav from "components/modals/ModalWithNav";
+import DurationValidation from "App/questionPage/Design/Validation/DurationValidation";
```
tgandrews: Could we make these relative imports.

eq-author/src/App/questionPage/Design/Validation/withCustomValueChange.js
```diff
@@ -0,0 +1,47 @@
+import PropTypes from "prop-types";
```
tgandrews: This could do with a better name as not all validation custom values are numeric and this is really specific to numeric fields.

eq-author/src/utils/withChangeUpdate.js
```diff
@@ -0,0 +1,21 @@
...
+import React from "react";
+
+const withChangeUpdate = WrappedComponent => {
+  return class extends React.Component {
+    static propTypes = {
+      onChange: PropTypes.func.isRequired,
+      onUpdate: PropTypes.func.isRequired
+    };
+
+    handleUpdate = update => this.props.onChange(update, this.props.onUpdate);
```
tgandrews: This looks remarkably to `withEntityTypeChange` is there any way we can reuse this?

adammahmood:
How do we feel about moving the following to `src/enhancers`?

`src/utils/enhancers.js` (Split into individual files or renamed maybe?)
`src/utils/withChangeUpdate.js` 

It feels a bit weird having these HOF in `utils` but placed here because originally we intended to completely remove the `enhancers` folder.

adammahmood:
@tgandrews I've tried to figure out a decent folder structure for the `src/App/questionPage/Design/Validation` components, however, a lot of the components are siblings.  I could maybe go down the path of grouping by type of components eg.

`enhancers`
`styles`
`validationTypes`

Do you think this would be beneficial?

adammahmood:
> > @adammahmood maybe we just remove codacy?
> 
> I think we should disable that particular rule.

I've disabled the rule. We can re-enable if anyone has any major objections.

adammahmood:
Failing due to: https://github.com/ONSdigital/eq-author-app/issues/109

eq-author/src/App/questionPage/Design/Validation/withCustomNumberValueChange.js
```diff
@@ -0,0 +1,47 @@
...
+import { inRange, isNaN } from "lodash";
+
+const withCustomNumberValueChange = WrappedComponent => {
+  return class extends React.Component {
+    static defaultProps = {
+      limit: 999999999
+    };
+
+    static propTypes = {
+      limit: PropTypes.number.isRequired,
```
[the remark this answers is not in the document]
adammahmood: Technically, you don't have to provide the `limit` prop when using the component as the `defaultProps` would be used. More explained here: https://reactjs.org/docs/typechecking-with-proptypes.html

> The defaultProps will be used to ensure that this.props.name will have a value if it was not specified by the parent component. The propTypes typechecking happens after defaultProps are resolved, so typechecking will also apply to the defaultProps.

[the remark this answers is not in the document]
SamGodwin2: Yeah just felt odd that we are saying this prop is required which suggests it should be passed into the component. Then also providing a default meaning that the prop can never not exist. It's a pretty minor thing tho.

Issue #109, opened 2019-01-07
Auth/duplication cypress flakiness


tgandrews:
### Expected behaviour
Cypress test for section duplication should always pass

### Actual behaviour
It is occasionally failing with
```
1) Duplicate Section duplication "before each" hook for "should display the new section in the sidebar":
     Uncaught TypeError: Cannot read property 'getItem' of null
This error originated from your application code, not from Cypress.
When Cypress detects uncaught errors originating from your application it will automatically fail the current test.
This behavior is configurable, and you can choose to turn this off by listening to the 'uncaught:exception' event.
https://on.cypress.io/uncaught-exception-from-application
Because this error occurred during a 'before each' hook we are skipping the remaining tests in the current suite: 'Section duplication'
      at G (http://localhost:13000/static/js/main.750ea0ef.chunk.js:1:153390)
      at Function.mapToProps (http://localhost:13000/static/js/main.750ea0ef.chunk.js:1:636452)
      at r (http://localhost:13000/static/js/2.7aafa1fb.chunk.js:1:138479)
      at http://localhost:13000/static/js/2.7aafa1fb.chunk.js:1:139747
      at d (http://localhost:13000/static/js/2.7aafa1fb.chunk.js:1:139793)
      at http://localhost:13000/static/js/2.7aafa1fb.chunk.js:1:139828
      at Object.run (http://localhost:13000/static/js/2.7aafa1fb.chunk.js:1:136398)
      at s.u.onStateChange (http://localhost:13000/static/js/2.7aafa1fb.chunk.js:1:136874)
      at p (http://localhost:13000/static/js/2.7aafa1fb.chunk.js:1:260154)
      at http://localhost:13000/static/js/2.7aafa1fb.chunk.js:1:867358
```

PR #110, opened 2019-01-08
Fix flakey cypress tests

SamGodwin2:
Thi Pr is required as for some reason when cypress tests in the CI environment the local storage is not always initialised in time meaning that when we try and set it within our code base we get Type Errors.

Fixes https://github.com/ONSdigital/eq-author-app/issues/109

Also fixes an issue in the routing_spec where the option label fields were not being blurred after input causing the data to occasionally not be saved.

### How to review 
Not convinced this is the most correct way to fix it but it does seem to do the job so any discussion would be good. Other than that make sure CI tests pass.

### DIFF
--- a/eq-author/cypress/utils/index.js
+++ b/eq-author/cypress/utils/index.js
@@ -102,6 +102,7 @@ export const buildMultipleChoiceAnswer = labelArray => {
       .get(testId("option-label"))
       .eq(index)
       .type(label)
+      .blur()
   );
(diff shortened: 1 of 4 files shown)


SamGodwin2:
[Fix flakey cypress test causing builds to fail](https://trello.com/c/8NSWxPZb/787-fix-flakey-cypress-test-causing-builds-to-fail)

PR #116, opened 2019-01-09
Fix broken storybook.

samiwel:
Storybook currently requires firebase project credentials to run which seems wrong. Storybook should primarily be there to test, develop and showcase our custom built components and should not have to rely on third party components such as Firebase.

This PR removes firebase as a dependency from the storybook launch.

### How to review 
Should be able to run storybook without supplying firebase project keys.


hamishtaplin:
@samiwel any idea why this PR is mostly code formatting changes?

samiwel:
Prettier ran and updated a bunch of files.

PR #123, opened 2019-01-11
Fix to stop description cut off

rmccar:
The question description was getting cut off after a line break. Originally it was just taking the first html tag and dropping the rest when it got into publisher. So this change has been to bypass the stripping of HTML tags that is done in the processPipedText for this field and then update the test concerning this. The users still cannot put line breaks in but the text will no longer be cut off because of it. This was deemed out of scope for this task and is something that will need to be looked at in the future.

### How to review 
Make sure tests pass

### DIFF
--- a/eq-author/src/components/RichTextEditor/index.js
+++ b/eq-author/src/components/RichTextEditor/index.js
@@ -76,9 +76,12 @@ const Input = styled.div`
   padding: 0;
 
   .header-two,
-  .unstyled,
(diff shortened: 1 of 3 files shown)


rmccar:
> I thought we were going to strip empty tags out?

We were but when looking into it, it turned into a bigger task where we would have to refactor a fair bit of code so we decided that was part of another piece of work and this piece should just be to stop the text being cut off

Issue #124, opened 2019-01-11
Partial error messages being thrown during unit tests in eq-author


adammahmood:
### Expected behaviour
To see full and useful error messages when a unit test fails

### Actual behaviour
Partial error message is displayed when running unit tests:

```
 console.error src/tests/setup/throwOnConsole.js:5
    Warning: Unknown event handler property `%s`. It will be ignored.%s
```

### Steps to reproduce the behaviour
- Add `onFoo={onUpdate}` to `<StyledWrappingInput/>` component in `src/components/AliasEditor.js`
- Run tests in `src/App/questionPage/Design/index.test.js`

### Screenshot
- Partial error message attached

PR #125, opened 2019-01-11
Fix partial error messages being thrown during unit tests

adammahmood:
- Fixes https://github.com/ONSdigital/eq-author-app/issues/124

### How to review 
- Ensure tests pass
- Ensure issue https://github.com/ONSdigital/eq-author-app/issues/124 is fixed

### DIFF
--- a/eq-author/src/tests/setup/throwOnConsole.js
+++ b/eq-author/src/tests/setup/throwOnConsole.js
@@ -1,7 +1,7 @@
 /* eslint-disable no-console, import/unambiguous */
 const actualConsoleError = console.error.bind(console);
 
-console.error = jest.fn(function(msg) {
-  actualConsoleError(msg);
+console.error = jest.fn(function(msg, ...rest) {
+  actualConsoleError(msg, ...rest);
   throw msg instanceof Error ? msg : new Error(msg);
 });


eq-author/src/tests/setup/throwOnConsole.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable no-console, import/unambiguous */
 const actualConsoleError = console.error.bind(console);
 
-console.error = jest.fn(function(msg) {
-  actualConsoleError(msg);
+console.error = jest.fn(function(msg, ...rest) {
+  actualConsoleError(msg, rest);
```
tgandrews: Should this spread as well?
```suggestion
  actualConsoleError(msg, ...rest);
```

rmccar:
Also tried adding Tom's suggested change but still got the same error

eq-author/src/tests/setup/throwOnConsole.js
adammahmood: Yes.  I think it's better to spread here as well.

adammahmood:
This PR fixes the `console.error` so it can utilise any available substitution parameters and log error messages with these parameters substituted as per the [Console API](https://developer.mozilla.org/en-US/docs/Web/API/console#Outputting_text_to_the_console)

Unfortunately, it does not resolve the message `Error` throws as this would involve complex string substitution.

PR #126, opened 2019-01-14
Added auth to graphql endpoint on Publisher

SamGodwin2:
After a recent change to add an authentication header to api requests the `/graphql/:questionnaireId` endpoint would fail authentication as the header was not being correctly applied. This PR adds the correct headers to that endpoint allowing the data to be retrieved correctly.

### How to review 
Start up the app and assert that a request to `/graphql/:questionnaireId` works as expected.


tgandrews:
Looks like the cypress test failure may be a re-occurance of the issue that should have been fixed by.
ONSdigital/eq-author-app/pull/100/files#diff-5390e6d64f724a77d37090bcb3e219e4R71

PR #127, opened 2019-01-14
Add question page definitions

adammahmood:
Adds ability to create question page definitions

### How to review 
- Ensure tests pass
- Ensure `eq-survey-runner` can render questionnaires with and without question page definitions

### DIFF
--- a/eq-author/src/App/questionPage/Preview/index.js
+++ b/eq-author/src/App/questionPage/Preview/index.js
@@ -1,19 +1,25 @@
+/* eslint-disable react/no-danger */
 import PropTypes from "prop-types";
 import React from "react";
 import { withApollo, Query } from "react-apollo";
 import gql from "graphql-tag";
 import { propType } from "graphql-anywhere";
 import styled from "styled-components";
+import { isEmpty } from "lodash";
 
-import EditorLayout from "App/questionPage/Design/EditorLayout";
 import IconText from "components/IconText";
 import Loading from "components/Loading";
 import Error from "components/preview/Error";
+import { Answer } from "components/preview/Answers";
+import PageTitle from "components/preview/elements/PageTitle";
+
+import EditorLayout from "App/questionPage/Design/EditorLayout";
 import QuestionPageEditor from "App/questionPage/Design/QuestionPageEditor";
 
-import { Answer } from "components/preview/Answers";
 import IconInfo from "./icon-info.svg?inline";
-import PageTitle from "components/preview/elements/PageTitle";
+import IconChevron from "./icon-chevron.svg";
+
+import { colors } from "constants/theme";
 
 const Container = styled.div`
   padding: 2em;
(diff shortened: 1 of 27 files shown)


adammahmood:
[Definitions (6)](https://trello.com/c/nuuM6bY9/695-definitions-6)

eq-author/src/App/questionPage/Preview/index.js
```diff
@@ -80,6 +119,24 @@ export const UnwrappedPreviewPageRoute = ({ loading, data }) => {
             <Panel dangerouslySetInnerHTML={{ __html: guidance }} />
           </Guidance>
         )}
+
+        {(!isEmpty(definitionLabel) || !isEmpty(definitionContent)) && (
```
tgandrews: I think this is the same - which I find easier to read.
```suggestion
        {!(isEmpty(definitionLabel) && isEmpty(definitionContent)) && (
```

eq-author/src/App/questionPage/Preview/index.js
```diff
@@ -80,6 +119,24 @@ export const UnwrappedPreviewPageRoute = ({ loading, data }) => {
...
+          <Details data-test="definition">
+            <DetailsTitle>
+              {definitionLabel || <Error small>Missing definition label</Error>}
+            </DetailsTitle>
+            <DetailsContent>
+              {definitionContent ? (
+                <span dangerouslySetInnerHTML={{ __html: definitionContent }} />
```
tgandrews: We should probably look at sanitising on the server as currently we don't check the values users are saving so we could have one user XSS another at the moment in these preview blocks. But it doesn't block this PR as this is an existing issue.

eq-author/src/App/questionPage/Preview/index.js
```diff
@@ -1,19 +1,25 @@
+/* eslint-disable react/no-danger */
```
tgandrews: How was the linting passing without this before?

eq-author/src/App/questionPage/Preview/index.js
adammahmood: Not too sure I agree.  It's better to have smaller blocks of logic which can be read without having to jump back.  Either way is fine with me 👍

eq-author/src/App/questionPage/Preview/index.js
adammahmood: Lets get a Technical Improvements card raised for this.

eq-author/src/App/questionPage/Preview/index.js
tgandrews: Yeah, I agree. I wasn't really sure. Maybe there would be some benefit in extracting this out to a separate function as it is starting to get a little large. I mean rendering the definition preview - not just the condition.

tgandrews: However, its just at the limit from my point of view where inline is ok.

PR #129, opened 2019-01-15
Jest coverage turned off as default

rmccar:
Jest coverage reports have become annoying when testing so we have taken the decision to turn them off as default. This should also speed up our tests. This can be overwritten and turned back on by adding ```--coverage``` after ```yarn test```. Coverage will continue to be collected on Travis.

Ive also removed the individual travis.yml files as these are no longer used after moving to a monotrepo

### How to review 
- Ensure tests pass
- Check that no coverage report is given if you run ```yarn test```
- Check that coverage report is given if you run ```yarn test --coverage```
- Check to see that coverage report is given in Travis


rmccar:
[jest collect coverage](https://trello.com/c/MWm7qw6b/811-jest-collect-coverage)

PR #130, opened 2019-01-15
Add response headers to improve security

adammahmood:
- Adds [Helmet](https://www.npmjs.com/package/helmet) package which adds security response headers.  This is recommended by [ExpressJS](https://expressjs.com/en/advanced/best-practice-security.html) as best practice.
- Some headers have been modified as per recommendations in https://trello.com/c/bIUdMXZ6/810-ensure-that-secure-headers-are-all-set

### How to review 
- Ensure tests pass
- Ensure headers listed in https://trello.com/c/bIUdMXZ6/810-ensure-that-secure-headers-are-all-set can be seen in response headers for all network requests.

### DIFF
--- a/eq-author-api/app.js
+++ b/eq-author-api/app.js
@@ -3,6 +3,7 @@ const { graphqlExpress } = require("graphql-server-express");
 const cors = require("cors");
 const bodyParser = require("body-parser");
 const pinoMiddleware = require("express-pino-logger");
+const helmet = require("helmet");
 
 const { PORT } = require("./settings");
 const status = require("./middleware/status");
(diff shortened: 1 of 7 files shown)


adammahmood:
[Ensure that secure headers are all set](https://trello.com/c/bIUdMXZ6/810-ensure-that-secure-headers-are-all-set)

jonnyshaw89:
@adammahmood do we want to add any tests to check the headers?

adammahmood:
> @adammahmood do we want to add any tests to check the headers?

@jonnyshaw89 I've added a e2e cypress test to check the correct headers are set.  It's quite difficult to unit test and makes much more sense as an e2e test.  What do you think?

jonnyshaw89:
@adammahmood yeh test looks ok. i just need to test it still loads in 3rd party JS

jonnyshaw89:
@adammahmood Does it make sense to add any of this to Author and Publisher as well?

adammahmood:
@jonnyshaw89 Yes. Definitely, worth adding to publisher as well.

adammahmood:
I've added the headers to `eq-publisher`.  Ideally, it would have been great to have a config in the root but this isn't possible as you can't `require` a module outside of a projects scope.  An alternative solution would be to have a separate package which could be re-used - this seems overkill considering we are about to merge API / Publisher.

PR #134, opened 2019-01-16
Bug - Earliest/Latest date range validations removed

tgandrews:
In the migration to create min/max duration for date range deleted all
other validations for date range. This lost the earliest/lates date
validations. This migration adds defaults back in for those date range
answers missing an earliest and latest date validation.

### How to review 
1. Create a date range answer
2. Delete the earliest/latest date validations for it
3. See the page fails to load
4. Run this migration
5. It now works


tgandrews:
[Bug - Pages with date range answers failing to load](https://trello.com/c/uhfsAwvE/819-bug-pages-with-date-range-answers-failing-to-load)

PR #136, opened 2019-01-17
Bug - Debounce was causing data to be lost

### LINKED ISSUE #135: Intermittently overwriting other entity's content on navigation
### Expected behaviour
1. Change a field on page 1
2. Navigate to page 2 (blurring the field)
3. Change should not update page 2

### Actual behaviour
When blurring the field to navigate this sometimes update page 2.

Hard to reproduce without making code changes. Might be easier with a slower machine/network

Can recreate by increasing the debounce in withEntityEditor.js:97 to something like 2000 (2 seconds).

tgandrews:
When a field is changed then the dirtyField property is set. On
blur of the field update is called where dirtyField is cleared as the
request to update is triggered.

When update was debounced and hence delayed by 20ms then the props to
which entity the dirtyField was for and which request to update would
be triggered could change between the dirtyField being set and the actual
update running.

This occurred when navigating to the same type of entity and blurring a
field at the same time.

Closes #135

### How to review 
Look at the recreation steps in #135


tgandrews:
[Bug - Intermittent data loss when navigating between same entity type](https://trello.com/c/9kyFFuAD/820-bug-intermittent-data-loss-when-navigating-between-same-entity-type)

PR #138, opened 2019-01-18
WIP - Speed up cypress by running in parallel

tgandrews:
Our plan has been upgraded so this change attempts to use cypress to run it in parallel

### How to review 
Don't yet


samiwel:
Not sure if I'm missing something but tests did not appear to run much faster with the parallel flag?

image

tgandrews:
@samiwel yeah looks like it requires more setup so trying to sort that now. We have to use multiple machines in travis.

tgandrews:
[Evaluate Cypress Parallel Capabilities](https://trello.com/c/ZOv8eZhP/821-evaluate-cypress-parallel-capabilities)

tgandrews:
This improves things quite a bit from a cypress test run point of view. However, spinning up the workers takes ~4 mins. So its not quick 😞 

screenshot 2019-01-21 at 15 52 01

PR #139, opened 2019-01-21
Support OIDC compliant auth provider

jonnyshaw89:
This allows us to disable firebase and trust that the infrastructure performs authentication for us using an OIDC compliant auth provider

### How to review 
With `AUTH_HEADER_KEY` set to `firebase` it should continue to work as before.
With `AUTH_HEADER_KEY` set to `external` the built in auth should be disabled so that the LoadBalancer can handle the auth for us.

### DIFF
--- a/eq-author-api/middleware/auth.js
+++ b/eq-author-api/middleware/auth.js
@@ -2,14 +2,14 @@ const { isNil, isEmpty } = require("lodash/fp");
 const jwt = require("jsonwebtoken");
 
 module.exports = (logger, context) => (req, res, next) => {
-  const authHeader = req.header("authorization");
+  const authHeader = req.header(process.env.AUTH_HEADER_KEY || "authorization");
   if (isNil(authHeader)) {
     logger.error("Request must contain a valid authorization header.");
     res.send(401);
     return;
   }
 
-  const accessToken = authHeader.replace("Bearer ", "");
+  const accessToken = authHeader.replace("Bearer ", "").replace(/=/g, "");
   if (isEmpty(accessToken)) {
     logger.error("Request must contain a valid access token.");
     res.send(401);
--- a/eq-author-api/middleware/auth.test.js
+++ b/eq-author-api/middleware/auth.test.js
@@ -88,10 +88,8 @@ describe("auth middleware", () => {
     });
 
     describe("valid token", () => {
-      let payload;
-
-      beforeEach(() => {
-        payload = {
+      it("should add token payload to context if valid token", () => {
+        let payload = {
           payload: {
             data: {
               some: "value",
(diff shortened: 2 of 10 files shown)


eq-author-api/middleware/auth.js
```diff
@@ -2,14 +2,14 @@ const { isNil, isEmpty } = require("lodash/fp");
...
-  const authHeader = req.header("authorization");
+  const authHeader = req.header(process.env.AUTH_HEADER_KEY || "authorization");
   if (isNil(authHeader)) {
     logger.error("Request must contain a valid authorization header.");
     res.send(401);
     return;
   }
 
-  const accessToken = authHeader.replace("Bearer ", "");
+  const accessToken = authHeader.replace("Bearer ", "").replace(/=/g, "");
```
tgandrews: Why is this necessary? Can we get a test around it as well.

jonnyshaw89: Looks like conflicting interpretations of the jwt spec

tgandrews: This seems wrong - its completely acceptable for a base64url encoded value to contain `=`. This needs more investigation.

eq-author/src/redux/auth/reducer.js
```diff
@@ -1,3 +1,4 @@
+import config from "../../config";
```
tgandrews: You don't need to use relative imports
 ```suggestion
import config from "config";
```

File: eq-author-api/middleware/auth.js (hunk 1 above)
Comment (jonnyshaw89) [reply eq-author-api/middleware/auth.js]: Yeh that seems to be the issue. each of the parts of a JWT is base64 encoded but some of the JWT libraries don't allow the `=`

Comment (jonnyshaw89) [reply eq-author-api/middleware/auth.js]: https://github.com/auth0/node-jsonwebtoken/issues/162

Comment (tgandrews) [reply eq-author-api/middleware/auth.js]: Can you add that into the code and add a test to cover this case. Just so its obvious. I don't mind if the link to the issue is only on the test.

File: eq-author-api/middleware/auth.test.js [hunk 3]
```diff
@@ -113,5 +113,38 @@ describe("auth middleware", () => {
         expect(next).toHaveBeenCalled();
       });
     });
+
+    describe("token containing =", () => {
```
tgandrews: I was expecting a small test with a token containing equals to be added to the previous describe - you can just redefine the mock to return the new bearer in the test. Thats all the logic that has been added AFAICT.

jonnyshaw89: it should be a new scenario if the mocks are changing shouldn't it?

PR #143, opened 2019-01-22
Travis perf - Remove build linting

tgandrews:
### What is the context of this PR
Linting costs 30 seconds - 1 minute. And we are performing it multiple times. Once separately and then once again as part of `yarn build`. This removes it from `yarn build` and `yarn start`.

### How to review 
1. Ensure times are lower than master
2. Ensure you are happy not having linting as part of `yarn start`.


tgandrews:
No obvious improvement at the moment.

PR #149, opened 2019-01-23
Bug - Unable to change routing else destination

tgandrews:
Graphql not importing the nested fragment as it was down two levels so this was causing a mutation read to fail

### How to review 
1. Ensure you can change the else destination
2. Tests pass


eq-author/cypress/integration/authenticated/routing_spec.js
```diff
@@ -437,4 +437,38 @@ describe("Routing", () => {
 
     cy.get(testId("deleted-answer-msg")).should("exist");
   });
+
+  it.only("can change the else destination", () => {
```
adammahmood:
```suggestion
  it("can change the else destination", () => {
```

PR #150, opened 2019-01-23
Travis Perf - Run two workers on one machine

tgandrews:
Trigger two cypress runs on one travis machine and have parallel split the tests between them

### How to review 
Ensure test run is fast.


tgandrews:
Once working this actually maintained the cypress run time at ~17 minutes as they shared resources on the travis build machine.

Issue #128, opened 2019-01-14
Section Introduction race condition


tgandrews:
### Expected behaviour
Section introduction cypress tests should pass successfully.

### Actual behaviour
Section introduction sometimes fails to be created

### Steps to reproduce the behaviour
Unknown other than on Travis and occasionally.

### Technical information
There seems to be a race condition between blurring a field and creating a section introduction as they both update the section. 

Debouncing the update in withEntityEditor was supposed to collect them so they happen at the same time but this doesn't seem to be true.
https://github.com/ONSdigital/eq-author-app/pull/100/files#diff-5390e6d64f724a77d37090bcb3e219e4R71

PR #151, opened 2019-01-24
Refactor section introduction into its own entity in both front-end and API.

SamGodwin2:
There was an occasional bug within the section introduction where you could get a race condition when changing the section title and also adding a section intro at the same time. This pr refactors introduction into its own graphql type and changes the returns of the update section and create sectionIntro mutations to ensure that there's no cache conflicts between the two.

### How to review 
Tests pass, changes make sense and it is not possible to get the cache into a weird state by changing the section title then blur'ing it by creating a section Intro. All other functionality should remain the same.

Should fix https://github.com/ONSdigital/eq-author-app/issues/128


eq-author-api/schema/resolvers/base.js
```diff
@@ -71,6 +71,34 @@ const Resolvers = {
       ctx.repositories.Section.remove(args.input.id),
     undeleteSection: (_, args, ctx) =>
       ctx.repositories.Section.undelete(args.input.id),
+    createSectionIntroduction: (
+      _,
+      { input: { sectionId, introductionContent, introductionTitle } },
+      ctx
+    ) =>
+      ctx.repositories.Section.update({
```
tgandrews: Updating the content and title seems pointless here.

eq-author/src/App/section/Design/SectionEditor/sectionIntroduction/index.js
```diff
@@ -0,0 +1,76 @@
...
+  };
+
+  hasIntro() {
+    return this.props.sectionIntro !== null;
+  }
+
+  render() {
+    return (
+      <StyledMainCanvas data-test="section-intro-canvas">
+        <TransitionGroup>
```
tgandrews: Do we want this `TransitionGroup` when there is no intro?

eq-author/src/App/section/Design/SectionEditor/sectionIntroduction/index.js
```diff
@@ -0,0 +1,76 @@
+import React from "react";
```
tgandrews: We capitalise a folder or file name if it is meant to be imported as a component. So  `sectionIntroduction/index` would become `SectionIntroduction/index`

eq-author/src/App/section/Design/SectionEditor/sectionIntroduction/index.test.js
```diff
@@ -0,0 +1,45 @@
...
+      createSectionIntro: jest.fn(),
+    };
+  });
+
+  it("should render section disabled", () => {
+    defaultProps.sectionIntro = null;
+
+    const wrapper = shallow(<UnwrappedSectionIntroduction {...defaultProps} />);
+
+    expect(wrapper).toMatchSnapshot();
```
tgandrews: Should this be the negative of the assertion on line 32?

eq-author-api/schema/resolvers/base.js
SamGodwin2: It's there to support the undo action without needing to make a new mutation for it. This way we can just re-create it with whatever's in the toast payload when they hit undo.

tgandrews: I guess we can remove this when we remove undo

eq-author/src/App/section/Design/SectionEditor/SectionIntroduction/updateSectionIntro.graphql
```diff
@@ -0,0 +1,13 @@
+mutation updateSectionIntroduction($input: UpdateSectionIntroductionInput!) {
+  updateSectionIntroduction(input: $input) {
+    id
+    introductionTitle
+    introductionContent
+    section {
```
tgandrews: I don't think we need to do this here. An update never affects the section does it?

eq-author/src/App/section/Preview/SectionIntroPreview.js
```diff
@@ -61,7 +61,7 @@ const SectionIntroPreview = ({
 );
 
 SectionIntroPreview.propTypes = {
-  section: PropTypes.shape({
+  introduction: PropTypes.shape({
```
adammahmood: Can we use the graphql fragment instead?

eq-author/src/App/section/Design/SectionEditor/SectionIntroduction/index.js
```diff
@@ -0,0 +1,77 @@
...
+
+export const IntroCanvas = styled.div`
+  padding: 0 1em;
+  border: 1px solid ${colors.bordersLight};
+  border-top: 0;
+  background-color: ${colors.white};
+  border-radius: 0 0 ${radius} ${radius};
+`;
+
+export class UnwrappedSectionIntroduction extends React.Component {
```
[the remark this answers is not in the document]
tgandrews: I guess it needs to be for the `handleCreateSectionIntro` otherwise we are invalidating the tree below for each re-render.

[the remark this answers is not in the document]
SamGodwin2: Yeah these were both classes just to prevent needing to define the handlers inline. Functional definitely looks better and without knowing whether or not it's a performance hit it's probably better to go with what looks nicer and is easier to follow.

PR #152, opened 2019-01-24
Travis perf - Dont pull two postgres containers

tgandrews:
Use the same container in the API tests and the integration tests.

### How to review 
1. Ensure we are only download the postgres container when running the API tests not when starting integration tests.


tgandrews:

PR #153, opened 2019-01-24
Fix pre deploy

tgandrews:
Whitespace matters and I recreate the error locally by running the following
```bash
$ export TRAVIS_BRANCH="master"
$ if ["$TRAVIS_BRANCH" == "master"]; then echo "true"; else echo "false"; fi
```
And then it is fixed by adding the spaces

### How to review 
1. Check that this script makes sense.


tgandrews:

PR #155, opened 2019-01-25
Master failing - Fix predeploy

tgandrews:
Travis does not reset directories between each stage so it was left in `eq-author` not at the root as I had assumed so `pushd` failed as there was nowhere to go back to.

This change couples the finishing directory of `script` and `before_deploy` but I think that is acceptable as I assume the next time we touch this will be re-organising the monorepo into one app.

### How to review 
1. Ensure this passes
2. Ensure this makes sense given this new information.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ script:
   - docker push onsdigital/eq-publisher:$TAG
 
 before_deploy:
-  - if [ "$TRAVIS_BRANCH" == "master" ]; then pushd eq-author; yarn storybook-build; popd; fi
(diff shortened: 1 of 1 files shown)


tgandrews:

jonnyshaw89:
Will this work? as `yarn storybook-build` wont work from the root?

tgandrews:
> Will this work? as `yarn storybook-build` wont work from the root?
@jonnyshaw89 
That was the problem. It isn't in the root hence the pushd/popd failed. So its actually still in `eq-author`

PR #156, opened 2019-01-25
Travis Perf - Stop spinning up new env just for e2e

tgandrews:
Stop spinning up new env especially for e2e tests

### How to review 
1. Ensure that we do not spin up a new env in travis for the e2e

### DIFF
--- a/eq-author/cypress/integration/e2e/e2e_spec.js
+++ b/eq-author/cypress/integration/e2e/e2e_spec.js
@@ -1,4 +1,5 @@
-import { question, questionnaire, section } from "../builders";
+import omitDeep from "omit-deep-lodash";
+import { question, questionnaire, section } from "../../builders";
 
 const assertSecurityHeaders = response => {
   expect(response.headers["referrer-policy"]).to.equal("no-referrer");
@@ -16,7 +17,7 @@ describe("End to end", () => {
   let questionnaireId;
   it("Has correct response headers", () => {
     cy.request({
-      url: "http://localhost:14000/graphql",
+      url: `${Cypress.env("API_URL") || "http://localhost:4000"}/graphql`,
       method: "POST",
       failOnStatusCode: false,
     }).then(response => {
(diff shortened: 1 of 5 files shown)


tgandrews:

eq-author/cypress/integration/e2e/e2e_spec.js
```diff
@@ -61,18 +62,22 @@ describe("End to end", () => {
...
+    const fieldsToIgnore = ["id", "form_type", "eq_id", "survey_id"];
+    cy.request(
+      `${Cypress.env("PUBLISHER_URL") ||
+        "http://localhost:9000"}/publish/${questionnaireId}`
+    ).then(response => {
+      // Use this to grab a copy of what was sent from publisher
+      // cy.writeFile(
+      //   "cypress/fixtures/publisher_new.json",
+      //   JSON.stringify(response.body, null, 2)
+      // );
```
jonnyshaw89: Should we just delete this code?

tgandrews: Its left so when people need to update the expected output the code is there to do it. Basically to make it as easy as possible to keep adding things to the build questionnaire and then not having to worry about this bit. I found it non-obvious that you had to use the cypress methods and not jut normal node methods.

eq-author/scripts/run_tests.sh
```diff
@@ -39,15 +39,13 @@ if [ "$1" == "integration" ]; then
...
-else
-  integration_folder="cypress/e2e";
 fi
 
 cypress_config="watchForFileChanges=false,integrationFolder=$integration_folder"
 
 # Run the tests
 if [ -z "${CYPRESS_RECORD_KEY-}" ]; then
-  yarn cypress run --browser chrome --config "$cypress_config"
+  CYPRESS_PUBLISHER_URL=http://localhost:19000 CYPRESS_API_URL=http://localhost:14000 yarn cypress run --browser chrome --config "$cypress_config"
```
samiwel: Shouldn't the CYPRESS_API_URL include the `/graphql`?

samiwel: Ignore comment, I misread the code.

PR #159, opened 2019-01-25
Add question page additional information

adammahmood:
Adds ability to create question page additional information.

### How to review 
- Ensure tests pass
- Ensure `eq-survey-runner` can render questionnaires with and without question page additional information


adammahmood:
[Additional information (3)](https://trello.com/c/JawkRJIM/814-additional-information-3)

eq-author/cypress/integration/authenticated/builder_spec.js
```diff
@@ -113,6 +113,27 @@ describe("builder", () => {
...
+      "contain",
+      additionalInfoLabel
+    );
+
+    const additionalInfoContent = "this is some additionalInfo content";
+    typeIntoDraftEditor(
+      testId("txt-question-additional-info-content", "testid"),
+      additionalInfoContent
+    );
+    cy.get(testId("txt-question-additional-info-content", "testid")).should(
```
tgandrews: I know we do this in a number of other places but, I've never checked if when the mutation fails this doesn't happen. I just want to know if all we are proving is that the RTE was updated which doesn't seem worthwhile as a cypress test.

eq-author/src/App/questionPage/Design/QuestionPageEditor/AdditionalInfo.js
```diff
@@ -0,0 +1,80 @@
...
+      onUpdate,
+      fetchAnswers,
+      onChangeUpdate,
+    } = this.props;
+    return (
+      <MultipleFieldEditor label="Additional information">
+        <Field>
+          <Label htmlFor="additional-info-label">Label</Label>
+          <WrappingInput
+            id="additional-info-label"
```
tgandrews: Do we still need this `id`? I thought we use `name`

eq-author/src/App/questionPage/Preview/index.js
```diff
@@ -137,6 +133,12 @@ export const UnwrappedPreviewPageRoute = ({ loading, data }) => {
           </Details>
         )}
 
+        {guidance && (
```
tgandrews: Was this in the wrong place?

eq-publisher/src/eq_schema/Question.js
```diff
@@ -92,6 +92,16 @@ class Question {
       this.type = "General";
       this.answers = this.buildAnswers(question.answers);
     }
+
+    if (question.additionalInfoLabel || question.additionalInfoContent) {
+      last(this.answers).guidance = {
+        /* eslint-disable-next-line camelcase */
+        show_guidance: question.additionalInfoLabel,
```
tgandrews: Are these boolean?

eq-publisher/src/eq_schema/Question.test.js
```diff
@@ -110,6 +110,59 @@ describe("Question", () => {
...
+            additionalInfoLabel: "additionalInfo label",
+            additionalInfoContent: "<p>additionalInfo content</p>",
+          })
+        );
+        expect(last(question.answers).guidance).toBeDefined();
+        expect(last(question.answers).guidance.show_guidance).toBeDefined();
+        expect(last(question.answers).guidance.hide_guidance).toBeDefined();
+        expect(last(question.answers).guidance.content).toBeDefined();
+      });
+      it("should be populated when label and no content", () => {
```
tgandrews: Do we ever want answer guidance with no content?

eq-author/src/App/questionPage/Design/QuestionPageEditor/AdditionalInfo.js
```diff
@@ -0,0 +1,80 @@
...
+  page: propType(pageFragment).isRequired,
+  onChangeUpdate: PropTypes.func.isRequired,
+};
+
+StatelessAdditionalInfo.fragments = {
+  Page: pageFragment,
+};
+
+export default flowRight(
+  withApollo,
```
tgandrews: I don't think this `withApollo` is needed or if it is for `fetchAnswers` could we push it down into there?

eq-author/src/App/questionPage/Design/QuestionPageEditor/AdditionalInfo.js
adammahmood: It's a required prop of `WrappingInput`.  It's not obvious why and I would be hesitant to make it optional. Do you want me to dig deeper?

eq-author/src/App/questionPage/Preview/index.js
adammahmood: Yes. Wrong place. It now matches `eq-survey-runner` layout.

eq-publisher/src/eq_schema/Question.js
adammahmood: No.  These are:

> The text that is used for the 'Show guidance' link

> The text that is used for the 'Hide guidance' link.

As per:

https://github.com/ONSdigital/eq-schema-validator/blob/master/schemas/common_definitions.json#L34

eq-publisher/src/eq_schema/Question.test.js
adammahmood: IMO, no it doesn't make sense and this came up during the development of question definitions as well.  It's valid schema so we decided not to introduce any ad-hoc validation.

eq-author/cypress/integration/authenticated/builder_spec.js
tgandrews: I think we do this all over the place so maybe we could raise a ticket for this and convert them to do that or remove them. Happy to leave it in this PR and deal with it later @samiwel

eq-author/src/App/questionPage/Design/QuestionPageEditor/AdditionalInfo.js
tgandrews: If its a 3rd party lib then don't bother otherwise yes please.

adammahmood: The `id` is required for `label` elements as described here:

> The id of a labelable form-related element in the same document as the <label> element. The first element in the document with an id matching the value of the for attribute is the labeled control for this label element, if it is a labelable element.

tgandrews: So its used on the label and input for accessibility. Makes sense.

eq-publisher/src/eq_schema/Question.test.js
tgandrews: I checked in runner and if there is no label it doesn't show no matter if there is content.

adammahmood: Ok.  Would need input from @hamishtaplin regarding this as it changes AC.  We should apply same concepts to definitions as well.  It would make sense to change the preview to match what runner does.

eq-author/src/App/questionPage/Design/QuestionPageEditor/AdditionalInfo.js
```diff
@@ -0,0 +1,78 @@
...
+
+import pageFragment from "graphql/fragments/page.graphql";
+
+const contentControls = {
+  bold: true,
+  emphasis: true,
+  piping: true,
+};
+
+export class StatelessAdditionalInfo extends React.Component {
```
SamGodwin2: Can this just be a functional component?

eq-author/src/App/questionPage/Preview/index.test.js
```diff
@@ -70,6 +77,55 @@ describe("PreviewPageRoute", () => {
     const wrapper = shallow(
       <PreviewPageRoute loading={false} data={{ questionPage }} />
     );
-    expect(wrapper.exists('[data-test="definition"]')).toBeFalsy();
+    expect(wrapper.exists(byTestAttr("definition"))).toBeFalsy();
+  });
+
+  it("should render definition label missing message", () => {
```
SamGodwin2: Should we be rendering the definition label if the message is missing? Runner seems to drop the whole thing if there's no message so the preview would be incorrect, not sure how to make it clear that the reason it's not displayed is due to the missing label though?

adammahmood: @hamishtaplin what are your thoughts on this?

PR #160, opened 2019-01-25
removal of undelete

rmccar:
Removal of the undelete functionality in Author. I have stripped the functionality out of the backend and removed the undelete button on the toast on the front end

### How to review 
Tests will not pass currently but test that the app functions


rmccar:
[Strip undelete functionality.](https://trello.com/c/EEZFnuBX/837-strip-undelete-functionality)

eq-author/src/components/ToastContainer/index.js
```diff
@@ -2,17 +2,10 @@ import React from "react";
 import { connect } from "react-redux";
```
tgandrews: Do we use toasts for anything more than undelete?

tgandrews: Is that still required or can we remove it completely?

samiwel: 🤷‍♂️ it's a UX change if we remove the Toast altogether, maybe @hamishtaplin has some thoughts?

PR #163, opened 2019-01-28
842 replace db question confirmation

adammahmood:
- Fixes question confirmation queries and mutations
- Adds a confirm modal when deleting a question confirmation

### How to review 
- Ensure questionnaire JSON is updated as expected
- Ensure queries return expected result
- Ensure confirm modal appears when deleting a question confirmation

### DIFF
--- a/eq-author-api/schema/resolvers/base.js
+++ b/eq-author-api/schema/resolvers/base.js
@@ -585,26 +585,23 @@ const Resolvers = {
       save(ctx.questionnaire);
       return deletedMetadata;
     },
-
     createQuestionConfirmation: (_, { input }, ctx) => {
       const section = findSectionByPageId(
         ctx.questionnaire.sections,
         input.pageId
       );
       const page = find(section.pages, { id: input.pageId });
-      const questionConfimation = {
+      const questionConfirmation = {
         id: uuid.v4(),
         title: "",
         positive: { label: null, description: null },
         negative: { label: null, description: null },
-        availablePipingAnswers: [],
-        availablePipingMetadata: [],
       };
-      set(page, "confirmation", questionConfimation);
+      set(page, "confirmation", questionConfirmation);
       save(ctx.questionnaire);
       return {
         pageId: input.pageId,
-        ...questionConfimation,
+        ...questionConfirmation,
       };
     },
     updateQuestionConfirmation: (
@@ -614,10 +611,8 @@ const Resolvers = {
     ) => {
       const newValues = {
         title,
-        positiveLabel: positive.label,
-        positiveDescription: positive.description,
-        negativeLabel: negative.label,
-        negativeDescription: negative.description,
+        positive,
+        negative,
       };
 
       const pages = flatMap(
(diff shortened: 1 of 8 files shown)


adammahmood:
[QuestionConfirmation](https://trello.com/c/VPm9wHx7/842-questionconfirmation)

eq-author-api/schema/resolvers/base.js
```diff
@@ -1076,18 +1071,9 @@ const Resolvers = {
...
-    negative: ({ negativeLabel, negativeDescription }) => ({
-      label: negativeLabel,
-      description: negativeDescription,
-    }),
     availablePipingAnswers: ({ id }, args, ctx) =>
-      ctx.repositories.QuestionConfirmation.getPipingAnswers(id),
-    availablePipingMetadata: ({ id }, args, ctx) =>
-      ctx.repositories.QuestionConfirmation.getPipingMetadata(id),
+      getPreviousAnswersForPage(ctx.questionnaire, id),
+    availablePipingMetadata: (page, args, ctx) => ctx.questionnaire.metadata,
```
tgandrews: Are these filtered by type? Or can we pipe all metadata?

eq-author/src/App/questionConfirmation/Design/index.js
```diff
@@ -27,14 +29,31 @@ export class UnwrappedQuestionConfirmationRoute extends React.Component {
...
+    showDeleteConfirmDialog: false,
+  };
+
+  handleOpenDeleteConfirmDialog = () =>
+    this.setState({ showDeleteConfirmDialog: true });
+
+  handleCloseDeleteConfirmDialog = cb =>
+    this.setState(
+      { showDeleteConfirmDialog: false },
+      isFunction(cb) ? cb : null
```
tgandrews: I think `setState` will handle this function check for us. I don't think we need to do it ourselves.

eq-author/src/App/questionConfirmation/Design/index.js
```diff
@@ -27,14 +29,31 @@ export class UnwrappedQuestionConfirmationRoute extends React.Component {
...
     );
+
+  handleDeletePageConfirm = () => {
+    this.handleCloseDeleteConfirmDialog(() => {
```
tgandrews: Could we just call setState as the function doesn't seem to do a lot.

eq-author-api/schema/resolvers/base.js
adammahmood: You can pipe any metadata into a question confirmation.

eq-author/src/App/questionConfirmation/Design/index.js
adammahmood: `cb` isn't always a function and `setState` will throw `Error: Invalid argument passed as callback. Expected a function. Instead received: ...` when an invalid type is passed as a callback.

eq-author/src/App/questionConfirmation/Design/index.js
adammahmood: I can set state in the separate functions. 👍

eq-author/src/App/questionConfirmation/Design/index.js
tgandrews: I was expecting to be `undefined` is that not the case?

tgandrews: No longer applicable.

adammahmood:
> Looks good just a couple of questions. Also, I was expecting a cypress change for undelete?

@tgandrews I've fixed the cypress tests to work with the latest UX.  I've decided not to touch anything related to "undo" / "undelete" as this is being stripped out in a separate PR.

PR #164, opened 2019-01-29
Fixed routing bug and added test

SamGodwin2:
There is currently a small bug in the api meaning that if you create a routing on a multiple choice answer, add a new option then set up routing based on that option the app will error when you then delete the option. This PR fixes the bug and ensures we do not return options that have been deleted. 

### How to review 
Above error is fixed and new test passes.

### DIFF
--- a/eq-author-api/schema/resolvers/routing2/binaryExpression2/integration.test.js
+++ b/eq-author-api/schema/resolvers/routing2/binaryExpression2/integration.test.js
@@ -207,6 +207,97 @@ describe("BinaryExpression Integration", () => {
     });
   });
 
+  it("should not return selected options that have been deleted", async () => {
+    const questionnaire = await buildTestQuestionnaire({
+      sections: [
+        {
+          pages: [
+            {
+              answers: [
+                {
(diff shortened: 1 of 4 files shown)


eq-author-api/schema/resolvers/routing2/binaryExpression2/integration.test.js
```diff
@@ -207,6 +207,97 @@ describe("BinaryExpression Integration", () => {
...
+
+    const deleteOptionQuery = `
+    mutation DeleteOption($input: DeleteOptionInput!) {
+      deleteOption(input: $input) {
+        id
+        __typename
+      }
+    }
+    `;
+    executeQuery(deleteOptionQuery, { input: { id: option1.id } }, ctx);
```
tgandrews: Should this be awaited?

SamGodwin2:
```suggestion
    await executeQuery(deleteOptionQuery, { input: { id: option1.id } }, ctx);
```

PR #165, opened 2019-01-29
Percentage answer type

tgandrews:
Adds percentage answer type which is available for routing and can be validated using min and max value validations.

---
Implement percentage answer type in creation, routing and validation.
We also found some inconsistency in formatting preview validation and
inputting min/max value so changed the number input to allow for a type
to be passed this is now used consistently in properties, validation
and routing.

### How to review 
1. Ensure you can route with it and see it in runner
2. Ensure you can apply validations and see them in runner
3. Tests all pass.


tgandrews:
[Percentage (9)](https://trello.com/c/CT2KWNQW/761-percentage-9)

hamishtaplin:
screenshot 2019-01-31 13 42 51

@tgandrews I think this `null` has been introduced here? Tested on pre-prod and it didn't happen.

PR #166, opened 2019-01-29
Previous answer minimum validation

SamGodwin2:
When first adding min value validation it was thought that it was unnecessary to allow you to validate the min value against a previous answer. Recently however it has come up as a user need and so this PR implements the functionality. 

### How to review 
Can now validate min value against a previous answer and new tests pass.


eq-author-api/schema/resolversTests/validation.test.js
```diff
@@ -244,6 +245,30 @@ describe("resolvers", () => {
       });
     });
 
+    it("can update inclusive and previous answer min values", async () => {
+      const previousAnswer = await createNewAnswer(firstPage, "Number");
```
tgandrews: I think this is wrong. We restrict it on the content picker to only type.

eq-author/cypress/integration/authenticated/validation_spec.js
```diff
@@ -99,20 +99,23 @@ describe("Answer Validation", () => {
       });
       it("Can set input value", () => {
         toggleCheckboxOn("@minValueToggle");
-        cy.get(testId("min-value-input"))
+        cy.get(testId("numeric-value-input"))
```
tgandrews: We are going to have a merge conflict with #165 but it should be ok to resolve.

SamGodwin2:
@adammahmood Been talking with Tom and Sam about that schema issue doesn't look like we have a suitable fix for it at the moment and it already exists in Max Value validation so gonna just ignore it for now see if it actually causes any problems and could be fixed by author validation.

eq-author-api/schema/resolversTests/validation.test.js
tgandrews: From the business rule (as implemented in the content picker) we restrict the previous answers to being the same type. This is setting the min value of a currency to a number.

SamGodwin2:
[QPSES previous answer minimum validation](https://trello.com/c/iPsyu30h/853-qpses-previous-answer-minimum-validation)

PR #171, opened 2019-01-30
Update routing logic to use JSON backend storage.

samiwel:
This PR is still somewhat of a work in progress. Whilst most of the routing logic has been switched over, there are a couple of side affects that still need to be taken care of (see couple of TODO statements in the base resolver).

### How to review 
Test the routing logic on this branch and highlight anything that I may have missed on this PR so I can fix.

### DIFF
--- a/eq-author-api/schema/resolvers/routing2/binaryExpression2/index.js
+++ b/eq-author-api/schema/resolvers/routing2/binaryExpression2/index.js
@@ -1,7 +1,34 @@
 const answerTypes = require("../../../../constants/answerTypes");
-const { find, flatMap, some, intersectionBy } = require("lodash/fp");
+const {
+  find,
+  flatMap,
+  some,
+  intersectionBy,
+  first,
+  getOr,
+  reject,
+  map,
+} = require("lodash/fp");
 
+const {
+  createExpression,
+  createLeftSide,
+} = require("../../../../src/businessLogic");
+
+const save = require("../../../../utils/saveQuestionnaire");
 const Resolvers = {};
+
+const answerTypeToConditions = require("../../../../modifiers/BinaryExpression/answerTypeToConditions");
+
+const isLeftSideAnswerTypeCompatible = (leftSideType, rightSideType) => {
+  const AnswerTypesToRightTypes = {
+    [answerTypes.CURRENCY]: "Custom",
+    [answerTypes.NUMBER]: "Custom",
+    [answerTypes.RADIO]: "SelectedOptions",
+  };
+
+  return AnswerTypesToRightTypes[leftSideType] === rightSideType;
+};
 
 Resolvers.BinaryExpression2 = {
   left: async ({ left }, args, ctx) => {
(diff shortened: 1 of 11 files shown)


eq-author-api/schema/resolvers/routing2/routingRule2/index.js
```diff
@@ -15,25 +25,103 @@ Resolvers.RoutingRule2 = {
...
+      )
+    );
+
+    const routingRule = createRoutingRule({
+      expressionGroup: createExpressionGroup({
+        expressions: [
+          createExpression({
+            left: createLeftSide({
+              type: "Null",
+              nullReason: "NoRoutableAnswerOnPage",
```
SamGodwin2: Needs to default to the first routable answer on page, falling back to this if there isn't one.

eq-author-api/schema/typeDefs.js
```diff
@@ -577,7 +577,7 @@ type Mutation {
   updateBinaryExpression2(input: UpdateBinaryExpression2Input!): BinaryExpression2!
   updateLeftSide2(input: UpdateLeftSide2Input!): BinaryExpression2!
   updateRightSide2(input: UpdateRightSide2Input!): BinaryExpression2!  
-  deleteBinaryExpression2(input: DeleteBinaryExpression2Input!): BinaryExpression2!
+  deleteBinaryExpression2(input: DeleteBinaryExpression2Input!): ExpressionGroup2!
```
SamGodwin2: How come we need to return the parent here?

samiwel: Due to not storing a back-ref to the parent the only way of determining the correct expression group is to look for the one that contains the expression. Since the expression is deleted you cannot resolve the parent.

It's only an issue when you delete something and want to get back to the parent from the returned deleted entity. In this case it is simpler to just return the parent directly.

Possibly there might be a nicer way of doing this.

samiwel:
[Routing](https://trello.com/c/SuRm4NAQ/840-routing)

samiwel:
@tgandrews agree I think this would be good to discuss in next tech session?

tgandrews:
@samiwel yeah, can do. I think the last point might actually be a bug in this PR as I think we were checking before.

eq-author-api/schema/resolvers/routing2/binaryExpression2/index.js
```diff
@@ -78,28 +105,239 @@ Resolvers.SelectedOptions2 = {
...
-    ctx.modifiers.LeftSide.update(input),
+    const expressionGroup = find(expressionGroup => {
+      if (some({ id: input.id }, expressionGroup.expressions)) {
+        return expressionGroup;
+      }
+    }, flatMap(rule => rule.expressionGroup, rules));
+
+    const expression = find({ id: expressionId }, expressionGroup.expressions);
+
+    const answer = find({ id: answerId }, flatMap(page => page.answers, pages));
```
tgandrews: This doesn't guarantee type or location in the document of the answer. So this might be wrong. I think this is a regression.

samiwel: It didn't seem to be validating this before?

https://github.com/ONSdigital/eq-author-app/blob/e58b720b9afb93d4745daec1ed3c05d34018739e/eq-author-api/modifiers/LeftSide/update.js

But you're right in that it probably should.

tgandrews: Ignore this then as its not a regression. Happy to resolve this later.

PR #173, opened 2019-01-31
Tech Debt - Code cov

tgandrews:
Deal with projects separately so we can easily see where coverage is good/bad.
https://docs.codecov.io/docs/flags

### How to review 
Don't yet.


.codecov.yml
```diff
@@ -0,0 +1,23 @@
...
+      default: off
+      author:
+        flags: author
+      api:
+        flags: api
+      publisher:
+        flags: publisher
+
+flags:
+  # filter the folder(s) you wish to measure by that flag
```
adammahmood: I think these comments are unnecessary

tgandrews: Yeah, this is just a copy and paste from the docs at the moment to see what we can get it to do at the moment.

adammahmood:
Definitely prefer this 👍

PR #174, opened 2019-01-31
Refactor and remove CompositeAnswers to use JSON backend store

adammahmood:
- Removes `CompositeAnswer` type and utilises `BasicAnswer` type for `DateRange` answers

### How to review 
- Ensure you can add/modify/delete `DateRange` answers
- Ensure you can add/modify/delete `DateRange` validations

### DIFF
--- a/eq-author-api/schema/resolvers/base.js
+++ b/eq-author-api/schema/resolvers/base.js
@@ -885,8 +885,6 @@ const Resolvers = {
     __resolveType: ({ type }) => {
       if (includes(["Checkbox", "Radio"], type)) {
         return "MultipleChoiceAnswer";
-      } else if (includes(["DateRange"], type)) {
-        return "CompositeAnswer";
       } else {
         return "BasicAnswer";
       }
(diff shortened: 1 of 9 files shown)


adammahmood:
[CompositeAnswers](https://trello.com/c/Q8oGjKx2/841-compositeanswers)

eq-author-api/schema/resolvers/base.js
```diff
@@ -907,22 +905,12 @@ const Resolvers = {
       return parentPage;
     },
     validation: answer =>
-      ["number", "date"].includes(getValidationEntity(answer.type))
+      ["number", "date", "dateRange"].includes(getValidationEntity(answer.type))
```
tgandrews: We should probably make these constants

eq-author-api/schema/typeDefs.js
```diff
@@ -138,27 +139,14 @@ type MultipleChoiceAnswer implements Answer {
...
   label: String
+  secondaryLabel: String
   type: AnswerType!
   options: [Option]
   mutuallyExclusiveOption: Option
   page: QuestionPage
   properties: JSON
 }
 
-type CompositeAnswer implements Answer {
```
tgandrews: If this doesn't exist can we get rid of `BasicAnswer` and just have `Answer`?

adammahmood: Attempting to keep changes to a minimum for the moment. Also I'm pretty sure we would get a naming conflict with `Answer` interface.

adammahmood:
@tgandrews Linting has a few failures.  We'll pick this up when we do a 2nd pass on everything to tidy it up.  I'm going to tackle all of the publishing changes as part of https://trello.com/b/GOqjEepx/eq-author

eq-author-api/schema/typeDefs.js
tgandrews: Can we just remove that interface?

adammahmood: It results in a lot of changes on the UI (which we're trying to leave as untouched as possible).

eq-author-api/schema/resolvers/base.js
adammahmood: This references values defined here: https://github.com/ONSdigital/eq-author-app/blob/39144650e141d176b0e66b37ab15c0cc0eb314ea/eq-author-api/utils/defaultAnswerValidations.js#L15 

Should we still convert to constants?

PR #175, opened 2019-01-31
Add function to deeply map over nested objects and transform nested id's

SamGodwin2:
Added a dependency on the deep-map function that allows us to easily recursively map over deeply nested objects.  

### How to review 
Changes make sense and when duplicating a page or section all nested ids should be updated to new consistent values.

### DIFF
--- a/eq-author-api/package.json
+++ b/eq-author-api/package.json
@@ -16,6 +16,7 @@
     "body-parser": "latest",
     "cheerio": "latest",
     "cors": "latest",
+    "deep-map": "^1.5",
     "dotenv": "latest",
     "express": "latest",
     "express-pino-logger": "latest",
--- a/eq-author-api/schema/resolvers/base.js
+++ b/eq-author-api/schema/resolvers/base.js
@@ -28,6 +28,7 @@ const {
 const fs = require("fs");
 const uuid = require("uuid");
 const stringify = require("json-stable-stringify");
+const deepMap = require("deep-map");
 
 const createAnswer = require("../../src/businessLogic/createAnswer");
 const updateMetadata = require("../../src/businessLogic/updateMetadata");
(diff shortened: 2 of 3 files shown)


eq-author-api/package.json
```diff
@@ -16,7 +16,9 @@
     "body-parser": "latest",
     "cheerio": "latest",
     "cors": "latest",
+    "deep-map": "latest",
```
adammahmood: I'm unsure about this library.  It's untested and hasn't been updated for 3 years.  Can we implement this algorithm ourselves?

eq-author-api/package.json
```diff
@@ -16,7 +16,9 @@
...
     "dotenv": "latest",
+    "es6-weak-map": "^2.0.2",
```
adammahmood: Why is this needed?

eq-author-api/package.json
tgandrews: Looks like it was updated 2 days ago to version 2? https://www.npmjs.com/package/deep-map

SamGodwin2: https://www.npmjs.com/package/deep-map seems to be plenty of weekly downloads and was updated 2 days ago?

adammahmood: My bad. Was looking at a different package: https://www.npmjs.com/package/deepmap

eq-author-api/schema/resolvers/base.js
```diff
@@ -153,6 +154,24 @@ const createQuestionnaire = input => ({
   ...input,
 });
 
+const remapAllNestedIds = entity => {
```
tgandrews: Can we put this in utils with a test around it?

adammahmood: We need to relocate loads of methods which are located in `base.js`.  I think it's best to save this until we restructure and figure out the final folder structure of the API.

SamGodwin2:
@adammahmood think my last comment got lost somewhere but as for the weak-map package it seems to be an issue that appeared when deep-map was updated from 1.5 to 2 so for now it might be best to pin it to 1.5 until that gets looked at.

tgandrews:
> @adammahmood think my last comment got lost somewhere but as for the weak-map package it seems to be an issue that appeared when deep-map was updated from 1.5 to 2 so for now it might be best to pin it to 1.5 until that gets looked at.

Can we open an issue with them as it looks like it happened recently. https://github.com/mcmath/deep-map/commit/1a98c4e93e16f2dd1cbb758835271b6205568c41

PR #176, opened 2019-01-31
Bug - Editor column changing width

tgandrews:
Fix issue where editing RTE could cause editing column to change width. This was caused by the max allowed width being 100% so it would grow with the RTE when it wasn't supposed to.

I appreciate the recreation steps are very specific but this is a guaranteed way of creating it but the causes are more common than just this case.

### How to review 
1. Set your viewport width to 1280px (use Chrome dev tools mobile button to make this easy)
1. Create a question with the title `Afafafafafaf afhoaaadad`
1. Add an answer with properties to the page (e.g. Currency)
1. Refresh the page.
1. Start updating the title and keep typing - at 10-12 characters (character width matters) into the second line the width of the editing column starts to grow.
1. Change to this branch and this no longer happens.

### DIFF
--- a/eq-author/src/App/QuestionnaireDesignPage/__snapshots__/index.test.js.snap
+++ b/eq-author/src/App/QuestionnaireDesignPage/__snapshots__/index.test.js.snap
@@ -94,7 +94,7 @@ exports[`QuestionnaireDesignPage should render 1`] = `
         />
       </Column>
       <Column
-        cols={12}
+        cols={9}
       >
         <Switch>
           <Route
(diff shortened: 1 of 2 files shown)


tgandrews:
[Width of central editing column is affected by content (2)](https://trello.com/c/Z3ieAGi5/831-width-of-central-editing-column-is-affected-by-content-2)

samiwel:
I can't seem to reproduce the issue locally, but the change looks reasonable.

tgandrews:
@samiwel even setting the width to 1280px?

samiwel:
Assuming I'm doing it right?

image

tgandrews:
@samiwel I think I've cracked it. You also need something in the properties panel. I've update the recreation steps.

PR #178, opened 2019-02-01
Update publisher to use latest schema for JSON datastore updates

adammahmood:
- Updates publisher to use latest schema for JSON datastore updates

### How to review 
- Ensure tests pass
- Ensure you can publish date range answers

### DIFF
--- a/eq-publisher/src/eq_schema/Questionnaire.js
+++ b/eq-publisher/src/eq_schema/Questionnaire.js
@@ -2,7 +2,7 @@
 const Section = require("./Section");
 const Summary = require("./block-types/Summary");
 const Confirmation = require("./block-types/Confirmation");
-const { last } = require("lodash");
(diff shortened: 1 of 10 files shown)


eq-publisher/src/eq_schema/Questionnaire.js
```diff
@@ -29,7 +29,8 @@ class Questionnaire {
     this.mime_type = "application/json/ons/eq";
     this.schema_version = "0.0.1";
     this.data_version = "0.0.2";
-    this.survey_id = questionnaireJson.surveyId || questionnaireId;
+    this.survey_id =
+      questionnaireJson.surveyId || kebabCase(questionnaireJson.title);
```
samiwel: This is what probably breaks preview. For now I would suggest

```js
this.survey_id = kebabCase(questionnaireJson.title);
```

PR #179, opened 2019-02-01
JSON DB - Cypress failures

tgandrews:
~Fix cypress failures apart from:~
1. ~Routing ~(still in active dev #171)~ Now in progress~
2. ~e2e as publisher isn't working yet. But works when rebased on #178~

No local cypress failures

### How to review 
1. That all cypress tests ~(apart from routing and e2e)~ pass successfully


tgandrews:
I’ll have a look at them as it was passing locally before the rebase

tgandrews:
@samiwel Header seems to be passing for me do you have any more details? I can definitely see the issue with validation; after the rebase the min value validation against previous answer came in so I need to get that working.

jonnyshaw89:
How are we going to ensure that these tests keep passing? as the CI builds don't currently get this far.

tgandrews:
Maybe I make a change to travis to run these first? Just for the moment?

jonnyshaw89:
@tgandrews Could we make travis not fail instantly on test failures? but fail after everything runs?

tgandrews:
@jonnyshaw89 I can't see anything obvious. But this is a definite concern. Maybe we just work on getting the tests passing on author and api. 

I'm working on the author failures at the moment and I think @samiwel is looking at the API.

samiwel:
[Fix JSON DB Cypress tests](https://trello.com/c/HAUf0J5L/864-fix-json-db-cypress-tests)

samiwel:
Retested this locally. Initially I had the same failures as before when running all tests, but upon running the failing tests in isolation they worked correctly. Happy to approve this PR for now as it's a big step in the right direction even if there is some worked required later to improve flakiness of the test.

Getting these tests running on travis would be a good indication.

PR #183, opened 2019-02-04
Use DynamoDB for Questionnaire persistence

jonnyshaw89:
Replace writing to file with writing to DynamoDB
Each revision writes a new revision
Saves only occur if the state has changed

### How to review 
Have a play and see if it all works.


eq-author-api/db/models/DynamoDB.js
```diff
@@ -0,0 +1,93 @@
...
+  },
+};
+
+const questionnanaireSchema = new Schema(baseQuestionnaireSchema, {
+  throughput: throughput,
+});
+
+const questionnaireVersionsSchema = new Schema(
+  {
+    ...baseQuestionnaireSchema,
```
tgandrews: I presume these versions have some shared Id to link them back to the original questionnaire? They all have `id` but I can't see a questionnaireId for them

eq-author-api/db/models/DynamoDB.js
```diff
@@ -0,0 +1,93 @@
+const dynamoose = require("dynamoose");
+
+let throughput = "ON_DEMAND";
+let questionnanaireTableName = "author-questionnaires";
+let questionnanaireVersionsTableName = "author-questionnaire-versions";
+
+if (process.env.DYNAMO_ENDPOINT !== "") {
```
tgandrews: It assumes local if you provide a `DYNAMO_ENDPOINT` that seems a little weird. How does it derive the DYNAMO_ENDPOINT otherwise? Maybe this needs a better name - something like `DYNAMO_ENDPOINT_DEV`

eq-author-api/db/models/DynamoDB.js
```diff
@@ -0,0 +1,93 @@
...
+if (process.env.DYNAMO_QUESTIONNAIRE_TABLE_NAME !== "") {
+  questionnanaireTableName = process.env.DYNAMO_QUESTIONNAIRE_TABLE_NAME;
+}
+
+if (process.env.DYNAMO_QUESTIONNAIRE_VERSION_TABLE_NAME !== "") {
+  questionnanaireVersionsTableName =
+    process.env.DYNAMO_QUESTIONNAIRE_VERSION_TABLE_NAME;
+}
+
+const Schema = dynamoose.Schema;
```
tgandrews: This var seems pointless. You can go `new dynamoose.Schema` without losing anything.

eq-author-api/db/models/DynamoDB.js
```diff
@@ -0,0 +1,93 @@
...
+  },
+  surveyId: {
+    type: String,
+  },
+  theme: {
+    type: String,
+  },
+};
+
+const questionnanaireSchema = new Schema(baseQuestionnaireSchema, {
```
tgandrews: Could we extract these schemas to a separate file to split them so each model/table deals with its setup.

eq-author-api/docker-compose.yml
```diff
@@ -5,6 +5,7 @@ services:
       context: .
     depends_on:
       - db
```
tgandrews: I'd like to see the PR that drops postgres ASAP

eq-author-api/app.js
```diff
@@ -14,8 +14,12 @@ const repositories = require("./repositories");
 const modifiers = require("./modifiers");
 const schema = require("./schema");
 
+const noir = require("pino-noir");
+
 const app = express();
-const pino = pinoMiddleware();
+const pino = pinoMiddleware({
+  serializers: noir(["req.headers.authorization"], "[Redacted]"),
```
tgandrews: This looks unrelated but a good change to stop logging the token. Do you want to do it against master instead?

eq-author-api/middleware/loadQuestionnaire.js
```diff
@@ -1,9 +1,21 @@
-const loadQuestionnaire = require("../utils/loadQuestionnaire");
+const { QuestionnanaireVersionsModel } = require("../db/models/DynamoDB");
```
tgandrews: I'd like for the models to be imported without the rest of the app knowing its Dynamo backing them and hence get it out of the path.

eq-author-api/middleware/loadQuestionnaire.js
```diff
@@ -1,9 +1,21 @@
...
+            return console.log(err);
+          }
+          if (!questionnaire.metadata) {
+            questionnaire.metadata = [];
+          }
+          resolve(questionnaire);
+        });
+    });
   }
   next();
```
tgandrews: No need for the `return` below

eq-author-api/middleware/loadQuestionnaire.js
```diff
@@ -1,9 +1,21 @@
...
+module.exports = (logger, context) => async (req, res, next) => {
   const questionnaireId = req.header("questionnaireId");
   if (questionnaireId) {
-    context.questionnaire = JSON.parse(loadQuestionnaire(questionnaireId));
+    context.questionnaire = await new Promise((resolve, _) => {
+      QuestionnanaireVersionsModel.queryOne({ id: { eq: questionnaireId } })
+        .descending()
+        .exec((err, questionnaire) => {
+          if (err) {
+            return console.log(err);
```
tgandrews: We should reject the promise not just log.

eq-author-api/middleware/loadQuestionnaire.js
```diff
@@ -1,9 +1,21 @@
...
+          }
+          if (!questionnaire.metadata) {
```
tgandrews: Why are we doing this? Don't all questionnaires have metadata?

eq-author-api/package.json
```diff
@@ -36,6 +37,7 @@
     "node-jose": "latest",
     "pg": "latest",
     "pg-hstore": "latest",
+    "pino-noir": "^2.2.1",
```
tgandrews: Do we need to pin?

eq-author-api/schema/resolvers/base.js
```diff
@@ -225,63 +218,49 @@ const Resolvers = {
...
       });
-      save(questionnaire);
-      const questionnaireList = getQuestionnaireList();
-      questionnaireList.push({
-        ...omit(questionnaire, "sections", "metadata"),
-        createdAt: questionnaire.createdAt.toString().split("T")[0],
-      });
-      saveQuestionnaireList(questionnaireList);
+
+      createQuestionnaire(questionnaire);
```
tgandrews: I presume we should await here?

eq-author-api/schema/resolvers/base.js
```diff
@@ -225,63 +218,49 @@ const Resolvers = {
...
+    duplicateQuestionnaire: async (_, { input }, ctx) => {
+      const questionnaire = await getQuestionnaire(input.id);
+      const newQuestionnaire = {
+        ...questionnaire,
+        title: addPrefix(questionnaire.title),
+        id: uuid.v4(),
+      };
 
-      save(newQuestionnaire);
+      createQuestionnaire(newQuestionnaire);
```
tgandrews: We should await here as well

eq-author-api/schema/resolvers/base.js
```diff
@@ -225,63 +218,49 @@ const Resolvers = {
...
-        createdAt: newQuestionnaire.createdAt.toString().split("T")[0],
-      });
-      saveQuestionnaireList(questionnaireList);
       return newQuestionnaire;
     },
     createSection: async (root, { input }, ctx) => {
       const section = createSection(input);
       ctx.questionnaire.sections.push(section);
-      save(ctx.questionnaire);
+      saveQuestionnaire(ctx.questionnaire);
```
tgandrews: We should await all of these `saveQuestionnaire`

eq-author-api/schema/resolvers/routing2/routing2/index.js
```diff
@@ -7,7 +7,7 @@ const isMutuallyExclusiveDestination = isMutuallyExclusive([
 ]);
 
 const { flatMap, find, first } = require("lodash/fp");
-const save = require("../../../../utils/saveQuestionnaire");
+const { save } = require("../../../../utils/datastore");
```
tgandrews: More awaits

eq-author-api/schema/resolvers/routing2/expressionGroup2/index.js
```diff
@@ -23,7 +23,7 @@ Resolvers.Mutation = {
       )
     );
     expressionGroup.operator = operator;
-    save(ctx.questionnaire);
+    saveQuestionnaire(ctx.questionnaire);
```
tgandrews: More awaits

eq-author-api/utils/datastore.js
```diff
@@ -0,0 +1,104 @@
...
+} = require("../db/models/DynamoDB");
+const { omit, isEqual } = require("lodash");
+
+const createQuestionnaire = async questionnaire => {
+  await new Promise((resolve, _) => {
+    new QuestionnanaireModel({
+      ...omit(questionnaire, "sections", "metadata"),
+    }).save({}, err => {
+      if (err) {
+        return console.log(err);
```
tgandrews: We should reject this.

eq-author-api/utils/datastore.js
```diff
@@ -0,0 +1,104 @@
...
+      }
+      new QuestionnanaireVersionsModel({
+        ...questionnaire,
+        metadata: questionnaire.metadata,
+        updatedAt: Date.now(),
+      }).save({}, err => {
+        if (err) {
+          return console.log(err);
```
tgandrews: Another rejection

eq-author-api/utils/datastore.js
```diff
@@ -0,0 +1,104 @@
...
+  if (equal) {
+    return questionnaire;
+  }
+  await new Promise((resolve, _) => {
+    if (questionnaire.updatedAt) {
+      questionnaire.updatedAt = Date.now();
+    }
+    questionnaire.save({}, err => {
+      if (err) {
+        return console.log(err);
```
tgandrews: We should reject here. There are more below as well

eq-author-api/utils/datastore.js
```diff
@@ -0,0 +1,104 @@
...
+      }
+      resolve();
+    });
+  });
+};
+
+const getQuestionnaire = async id => {
+  let questionnaire = await new Promise((resolve, _) => {
+    QuestionnanaireVersionsModel.queryOne({ id: { eq: id } })
+      .descending()
```
tgandrews: Ordered by what?

eq-author-api/db/models/DynamoDB.js
jonnyshaw89: The `id` for both tables is the questionnaireId

eq-author-api/db/models/DynamoDB.js
jonnyshaw89: Its more of an override, when you run in AWS you don't need to provide one ill rename it to `DYNAMO_ENDPOINT_OVERRIDE`

eq-author-api/docker-compose.yml
jonnyshaw89: I think there is some other stuff still using the DB so this needs to be changed 1st

eq-author-api/db/models/DynamoDB.js
tgandrews: As discussed this makes sense with the updated at making it unique.

eq-author-api/middleware/loadQuestionnaire.js
jonnyshaw89: Yeh I refactored most of this out but missed this one.

eq-author-api/middleware/loadQuestionnaire.js
jonnyshaw89: For some reason the empty Metadata arrays were not being persisted. I need to look at this

eq-author-api/schema/resolvers/base.js
jonnyshaw89: even though there is an await inside `createQuestionnaire`?

eq-author-api/utils/datastore.js
jonnyshaw89: By the `rangeKey` defined on the table

eq-author-api/app.js
jonnyshaw89: https://github.com/ONSdigital/eq-author-app/pull/184

eq-author-api/schema/resolvers/base.js
tgandrews: Yeah, `async` functions return a `Promise` so to wait for them to be done you have to `await` them otherwise you'll start the outer one but not await on its result even if internally it was waiting on something else to finish. I hope that makes sense.

.travis.yml
```diff
@@ -45,3 +45,4 @@ deploy:
 branches:
   only:
     - master
+    - 797-replace-relational-database-with-json-document
```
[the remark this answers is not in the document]
adammahmood: Completed via https://github.com/ONSdigital/eq-author-app/pull/193

adammahmood:
Can we rebase this? Getting errors when running locally which should be resolved once rebased.

eq-author-api/docker-compose.yml
```diff
@@ -28,3 +35,7 @@ services:
       - POSTGRES_DB=postgres
     ports:
       - 5432:5432
+  dynamo:
+    image: amazon/dynamodb-local
+    ports:
+      - 8000
```
adammahmood: This port conflicts with `go-launch-a-survey`.

```suggestion
      - 8050:8000
```

eq-author/src/utils/appendAuthHeader.js
```diff
@@ -1,3 +1,5 @@
+/*eslint-disable no-useless-escape */
```
tgandrews: I don't think this is needed any more.

eq-author-api/schema/resolvers/routing2/routing2/index.js
tgandrews: Does this import even exist?

eq-author-api/utils/datastore.js
```diff
@@ -0,0 +1,97 @@
...
+
+const createQuestionnaire = questionnaire => {
+  return new Promise((resolve, reject) => {
+    new QuestionnaireModel({
+      ...omit(questionnaire, "sections", "metadata"),
+    }).save({}, err => {
+      if (err) {
+        reject(err);
+      }
+      new QuestionnaireVersionsModel({
```
tgandrews: To avoid the call back inside a call back you could turn these into two methods each wrapping the functions in a promise. You might even be able to do this at the import easily with `util.promisify` https://nodejs.org/dist/latest-v10.x/docs/api/util.html#util_util_promisify_original

jonnyshaw89: so like this?
```
return util.promisify(new QuestionnaireModel({
    ...omit(questionnaire, "sections", "metadata"),
  }).save({})).then(() => {
    util.promisify(new QuestionnaireVersionsModel({
      ...questionnaire,
      metadata: questionnaire.metadata,
      updatedAt: Date.now(),
    }).save()).then(() => {
      return questionnaire
    })
  })
```

tgandrews: No, you'd normally wrap the function but as it is an instance method this might not be so easy.

PR #188, opened 2019-02-05
Remove Undelete

rmccar:
This is to remove the undelete functionality for the app as we will no longer support undeleting as we will now be implementing a versioning system instead. This is a replacement for a previous PR that included questionnaire deletion confirmation but that is now in a separate PR

### How to review 
- Check the app works as it should
- Check you can delete all types of things (questionnaire, section, question, answer and section introduction)
- Check that when deleting these things the toast is still shown but now there is no undo button


rmccar:
[Strip undelete functionality.](https://trello.com/c/EEZFnuBX/837-strip-undelete-functionality)

eq-author/cypress/integration/authenticated/builder_spec.js
```diff
@@ -264,42 +264,6 @@ describe("builder", () => {
     cy.get(testId("btn-add-intro"));
   });
 
-  it("can undelete a section introduction", () => {
```
tgandrews: I don't think this is the only undelete cypress test

rmccar: Ive done a search for undelete and cant seem to find any others

rmccar: You're right Ive found some that don't use the word 'undelete'

eq-author/src/components/ToastContainer/index.js
```diff
@@ -55,36 +38,15 @@ const ToastWrapper = styled.div`
   justify-content: center;
 `;
 
-export const ToastArea = ({
-  toasts,
-  dismissToast,
-  undoToast,
-  ...otherProps
-}) => {
```
SamGodwin2: Don't think you've updated the tests for this component probably a couple broken snapshot tests in there somewhere.

rmccar: Fixed

PR #190, opened 2019-02-05
JSON DB - Fix author linting and unit tests

tgandrews:
All author tests and linting should pass.

The biggest change was around the url to ensure we differentiate between ids and the tab name. This re-introduces the regex pattern back into the URL to do this but had the knock on effect of requiring everywhere that built URLs to create "valid" (according to the regex pattern) UUIDs this is done in tests now using `src/tests/utils/fakeId` which generates a 36 long string from what was given as the fill. This could probably be improved.

### How to review 
1. Ensure all linting and tests (unit and cypress) pass in author.


tgandrews:
[Fix author linting and tests](https://trello.com/c/9Ajk21jg/866-fix-author-linting-and-tests)

PR #196, opened 2019-02-07
Bug - Fix rendering of nested flexboxes in Chrome 72

### LINKED ISSUE #189: Content picker overflowing when large number of options.
### Expected behaviour
The content picker modal should scroll when there is a large number of options inside the accordion.
 
### Actual behaviour
The accordion panel just expands until it starts overflowing the modal.

Make a questionnaire with <10 pages, on the first page go to routing and open the destination content picker. Select the destinations in this section and observe...

### Screenshot

screenshot 2019-02-05 at 14 47 01

tgandrews:
This resolves the issue where Chrome was not matching the web standard and defaulting the `min-height` of a nested flexbox item to `0`. To improve compatibility they changed the default to `auto`. This broke the places where we had nested flex boxes

There is more information and references in #189 

Closes #189 

### How to review 
1. Ensure that you can scroll the main app
2. Ensure that the ContentPicker does not overflow as described in #189

### DIFF
--- a/eq-author/src/components/BaseLayout/index.js
+++ b/eq-author/src/components/BaseLayout/index.js
@@ -26,7 +26,7 @@ const Main = styled.main`
   flex: 1 1 auto;
   display: flex;
   flex-direction: column;
-  height: calc(100vh - 4em);
+  min-height: 0;
 `;
 
 const Title = styled.h1`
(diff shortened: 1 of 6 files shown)


tgandrews:
[Chrome 72 layout bugs](https://trello.com/c/nBzWIEQk/920-chrome-72-layout-bugs)

tgandrews:
@rmccar - good spot. I missed the routing content picker. This should be resolved now.

PR #200, opened 2019-02-08
Added functionality to give content picker a selected state.

SamGodwin2:
This Pr implements an improvement to the content picker in the UI. It allows it to have a selected state on opening meaning that if a answer has already been selected the picker should open onto that answer instead of being at the top of the tree.

### How to review 
On either Routing or Validation previous answer select, set a previous answer then when you close and re-open the content picker the selected state should still be selected. Also ne tests pass.

### DIFF
--- a/eq-author/src/components/ContentPicker/ContentPicker.js
+++ b/eq-author/src/components/ContentPicker/ContentPicker.js
@@ -5,8 +5,9 @@ import { isNil } from "lodash";
 
 import ButtonGroup from "components/buttons/ButtonGroup";
 import Button from "components/buttons/Button";
-
 import ContentPickerSingle from "components/ContentPicker/ContentPickerSingle";
+
+import setSelectedElement from "./setSelectedElement";
 
 const ContentWrapper = styled.div`
   display: flex;
@@ -19,6 +20,7 @@ const ActionButtons = styled(ButtonGroup)`
   padding: 1em 0;
   flex: 0 0 auto;
 `;
+
 export default class ContentPicker extends React.Component {
   static propTypes = {
     onSubmit: PropTypes.func,
@@ -28,6 +30,7 @@ export default class ContentPicker extends React.Component {
         id: PropTypes.string.isRequired,
       })
     ).isRequired,
+    selectedId: PropTypes.string,
     config: PropTypes.arrayOf(
       PropTypes.shape({
         id: PropTypes.string.isRequired,
@@ -38,7 +41,16 @@ export default class ContentPicker extends React.Component {
   };
 
   state = {
-    selectedItems: [],
+    selectedItems: this.props.selectedId
+      ? [
+          ...setSelectedElement(
+            this.props.config,
+            this.props.selectedId,
+            this.props.data
+          ),
+        ]
+      : [],
+    openLevel: this.props.selectedId ? this.props.config.length - 1 : null,
   };
 
   getDataAtLevel(level) {
@@ -83,7 +95,7 @@ export default class ContentPicker extends React.Component {
       const data = this.getDataAtLevel(level);
 
       const isDisabled = level > selectedItems.length || !data.length;
-      const isHidden = level > openLevel;
+      const isHidden = openLevel !== null && level > openLevel;
       const isOpen = level === openLevel || config.length === 1;
 
       const selectedItem = selectedItems[level];
@@ -118,6 +130,7 @@ export default class ContentPicker extends React.Component {
     const { config } = this.props;
     const { selectedItems } = this.state;
     const selectedItem = selectedItems[config.length - 1];
+
     return (
       <React.Fragment>
         <ContentWrapper>{this.renderPickers()}</ContentWrapper>
--- a/eq-author/src/components/ContentPicker/GroupContentPicker.js
+++ b/eq-author/src/components/ContentPicker/GroupContentPicker.js
@@ -1,7 +1,7 @@
 import React from "react";
 import styled from "styled-components";
 import PropTypes from "prop-types";
-import { isNil } from "lodash";
+import { isNil, findIndex, forEach } from "lodash";
 
 import ButtonGroup from "components/buttons/ButtonGroup";
 import Button from "components/buttons/Button";
@@ -21,6 +21,33 @@ const ActionButtons = styled(ButtonGroup)`
   flex: 0 0 auto;
 `;
 
+const generateDefaultState = (selectedObj, config) => {
+  if (!selectedObj) {
+    return {
+      openLevel: null,
+      selectedItem: null,
+    };
+  }
+  let openLevel;
+  let selectedItem;
+  forEach(selectedObj, value => {
+    if (value === null) {
+      return;
+    }
+    if (typeof value !== "object") {
+      openLevel = null;
+    } else {
+      openLevel = findIndex(config, ({ id }) => id === value.__typename);
+    }
+    selectedItem = value;
+    return false;
+  });
+  return {
+    openLevel,
+    selectedItem,
+  };
+};
+
 export default class GroupContentPicker extends React.Component {
   static propTypes = {
     onSubmit: PropTypes.func,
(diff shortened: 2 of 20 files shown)


hamishtaplin:
@SamGodwin2 possible to add `aria-pressed` to the selected button?

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

eq-author/src/components/ContentPicker/ContentPicker.js
```diff
@@ -82,7 +151,7 @@ export default class ContentPicker extends React.Component {
       const data = this.getDataAtLevel(level);
 
       const isDisabled = level > selectedItems.length || !data.length;
-      const isHidden = level > openLevel;
+      const isHidden = openLevel != null && level > openLevel;
```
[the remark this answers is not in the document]
SamGodwin2: ❗️

eq-author/src/components/ContentPicker/ContentPicker.js
```diff
@@ -37,7 +97,16 @@ export default class ContentPicker extends React.Component {
   };
 
   state = {
-    selectedItems: [],
+    selectedItems: this.props.selectedId
```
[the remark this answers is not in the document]
SamGodwin2: I destructure them in the functions, can I actually destructure them here without creating a constructor on the class?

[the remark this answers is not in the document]
hamishtaplin: Ah yeah sorry I didn't realise you weren't in a place it could be done 👍

eq-author/src/components/ContentPicker/ContentPicker.js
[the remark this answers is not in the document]
hamishtaplin: Were you intentionally not using the strict operator? It's generally considered best-practise to default to the strict ones.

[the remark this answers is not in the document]
tgandrews: We should probably be linting this

eq-author/src/components/ContentPicker/ContentPicker.js
```diff
@@ -18,6 +18,65 @@ const ActionButtons = styled(ButtonGroup)`
...
+        const path = Object.keys(cur)[0];
+        const indexAtPath = Object.values(cur)[0];
+        acc.push(acc[idx - 1][path][indexAtPath]);
+        return acc;
+      }
+    },
+    []
+  );
+};
+
```
samiwel: The logic in `getDesiredPath` and `setSelectedElement` is pretty complex (lots of array index calculations going on). It clearly works, but it might be an idea to add some unit tests around it to document the behaviour more than anything else.

Failing that, a useful comment explaining what the expected inputs and outputs of the function will look like might be justified. Future you will thank you for it! 😄

eq-author/src/components/ContentPicker/ContentPicker.js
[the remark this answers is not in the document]
SamGodwin2: Nah it should be strict, a linting rule would be a good idea, I always forget to do strict on `not equals`'s.

eq-author/src/components/ContentPicker/ContentPicker.js
```diff
@@ -18,6 +18,65 @@ const ActionButtons = styled(ButtonGroup)`
...
+    compact(
+      data.map((test, index) => {
+        const result = getDesiredPath(test, selectedId, config, 0);
+        return result ? [index, ...result] : null;
+      })
+    )
+  );
+  return reduce(
+    dataPath,
+    (acc = [], cur, idx) => {
```
tgandrews: You don't need this default - you have provided the starting accumulator below.

eq-author/src/components/ContentPicker/ContentPicker.js
```diff
@@ -18,6 +18,65 @@ const ActionButtons = styled(ButtonGroup)`
...
+const setSelectedElement = (config, selectedId, data) => {
+  const dataPath = flatten(
+    compact(
+      data.map((test, index) => {
+        const result = getDesiredPath(test, selectedId, config, 0);
+        return result ? [index, ...result] : null;
+      })
+    )
+  );
+  return reduce(
```
tgandrews: You can just use dataPath.reduce

eq-author/src/components/ContentPicker/ContentPicker.js
```diff
@@ -18,6 +18,65 @@ const ActionButtons = styled(ButtonGroup)`
...
+      }
+    } else if (
+      dataset[config[currentIndex].childKey][dataIndex].id === selectedId
+    ) {
+      return [...path, { [config[currentIndex].childKey]: dataIndex }];
+    }
+  }
+};
+
+const setSelectedElement = (config, selectedId, data) => {
```
tgandrews: This looks like it could be extracted to a separate file which would make it easy to test independently.

eq-author/src/components/ContentPicker/ContentPicker.js
```diff
@@ -18,6 +18,65 @@ const ActionButtons = styled(ButtonGroup)`
...
+  const dataPath = flatten(
```
tgandrews: Could we use `flow` here instead of lots of nested wrapping.

eq-author/src/components/ContentPicker/GroupContentPicker.js
```diff
@@ -20,6 +20,31 @@ const ActionButtons = styled(ButtonGroup)`
...
 `;
 
+const generateDefaultState = (selectedDestination, config) => {
+  if (!selectedDestination) {
+    return {
+      openLevel: null,
+      selectedItem: null,
+    };
+  }
+  Object.keys(selectedDestination).forEach(key => {
```
tgandrews: Instead of clearing things out of the object. Could we instead search through the possible locations?
so 35 becomes something like: `destinationType = selectedDestination.page || ...`

tgandrews: Also this is very routing specific for a `GroupContentPicker`

eq-author/src/components/ContentPicker/GroupContentPicker.test.js
```diff
@@ -4,8 +4,8 @@ import ContentPickerButton from "components/ContentPicker/ContentPickerButton";
 import GroupContentPicker from "components/ContentPicker/GroupContentPicker";
 import { byTestAttr } from "tests/utils/selectors";
 
-const PAGES_PICKER = byTestAttr("pages-picker");
-const SECTIONS_PICKER = byTestAttr("sections-picker");
+const PAGES_PICKER = byTestAttr("page-picker");
```
tgandrews: Change the constants as well just for consistency

eq-author/src/components/ContentPicker/ContentPicker.js
SamGodwin2: Is this what you had in mind, not sure it makes it more readable? 
```
  const noisyDataPath = data.map((data, index) => {
    const result = getDesiredPath(data, selectedId, config, 0);
    return result ? [index, ...result] : null;
  });
  const cleanPath = flow(
    compact,
    flatten
  );
  const dataPath = cleanPath(noisyDataPath);
```

eq-author/src/components/ContentPicker/GroupContentPicker.js
```diff
@@ -20,6 +20,33 @@ const ActionButtons = styled(ButtonGroup)`
...
+  let openLevel;
+  let selectedItem;
+  forEach(selectedObj, value => {
+    if (value === null) {
+      return;
+    }
+    if (typeof value !== "object") {
+      openLevel = null;
+    } else {
+      openLevel = findIndex(config, ({ id }) => id === value.__typename);
```
tgandrews: If `id === typename`? That looks weird - either its an id or a typename.

SamGodwin2: Yeah the id of the config dictates what entity the config refers to, could add another field to the config with the same value? This whole group content picker feels very awkward here but this seemed to be the best way to set it up without making it specific to routing?

tgandrews: Yeah, that might be a reasonable solution. Remove the idea of the group content picker and move it all into routing. But that can be looked at later.

PR #202, opened 2019-02-08
Break out of formatting in RTE

tgandrews:
This implements breaking out of formatting in rich text editors.

We have switched to using a wrapper around DraftJS called [DraftJS Plugins](https://www.draft-js-plugins.com/) which provides a nice interface for building plugins on top of DraftJS. This allowed us to use the [breakout plugin](https://github.com/icelab/draft-js-block-breakout-plugin/).

We have taken the [defaults of the breakout plugin](https://github.com/icelab/draft-js-block-breakout-plugin/#options). This includes breaking out of headings on one enter and breaking out of lists on two enters. 

### Note
~~It seems to require hard refreshing after switching to the branch - don't rely on simple browser refresh or hot reload.~~
This has a bug at the moment so its back in progress.

### How to review 
1. Ensure that RTE breaks out of formatting as expected.
2. Ensure that other RTE functionality is maintained.

### DIFF
--- a/eq-author/package.json
+++ b/eq-author/package.json
@@ -116,6 +116,8 @@
     "downshift": "latest",
     "draft-convert": "latest",
     "draft-js": "latest",
+    "draft-js-block-breakout-plugin": "latest",
+    "draft-js-plugins-editor": "latest",
     "draft-js-raw-content-state": "latest",
     "draftjs-filters": "^1.0.0",
     "draftjs-to-html": "^0.7.6",
--- a/eq-author/src/components/RichTextEditor/RichTextEditor.test.js
+++ b/eq-author/src/components/RichTextEditor/RichTextEditor.test.js
@@ -1,9 +1,12 @@
 import React from "react";
 import { shallow } from "enzyme";
 import { SynchronousPromise } from "synchronous-promise";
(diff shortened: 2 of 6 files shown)


tgandrews:
[Break out of list (RTE) (8)](https://trello.com/c/XJDafWcV/256-break-out-of-list-rte-8)

hamishtaplin:
@tgandrews Seems to break piping?

screenshot 2019-02-08 13 50 18

tgandrews:
@hamishtaplin Good spot. It doesn't seem to work in multiline but it does in single line I'll have a look.

hamishtaplin:
@tgandrews Piping a value puts the cursor back to start of the RTE. I've checked on pre-prod and it doesn't do that there. Has this been introduced here?

tgandrews:
No it wasn’t but I thought it had been happening for a while. I’ll have a
look into this further. Maybe the hack isn’t enough 😩

tgandrews:
@hamishtaplin I think this *should* be resolved now - following the pattern in the emoji plugin. https://github.com/draft-js-plugins/draft-js-plugins/blob/9975f41d7f0533843882e5bf4e483081c17f7d12/draft-js-emoji-plugin/src/modifiers/addEmoji.js#L97

PR #203, opened 2019-02-12
Fix for no name in firebase

rmccar:
Firebase seems to have an intermittent issue with sending the name in the token. This will resort to using the email if no name is supplied.

### How to review 
- Run tests
- Try using author without a name on account


rmccar:
[Unknown users in pre-prod.](https://trello.com/c/TfngcrdX/924-unknown-users-in-pre-prod)

PR #205, opened 2019-02-12
Fix wording on delete questionnaire

rmccar:
Fix for the wording on deleting questionnaire to be changed to make more sense

### How to review 
Make sure tests pass


rmccar:
[Add questionnaire deletion confirmation](https://trello.com/c/IKz0lNtD/919-add-questionnaire-deletion-confirmation)

PR #206, opened 2019-02-13
Fix - Context being shared across requests

tgandrews:
We were sharing the context & therefore questionnaire across requests which was causing dirty reads.

For section intro there was a race condition where the mutation for createSectionIntro (cSI) was running interlaced with updateSection (uS).

There are 3 stages to each request: 
1. the read of the questionnaire from the db (in middleware)
2. the mutation - running the mutation resolver and triggering the write
3. the read of the values on return (using the return value in the resolvers) 

They are broken up like this due as between these are the async points where other things that were queued up can start in-between and run until their async point.

The problem we had was that all parts were relying on `ctx.questionnaire` which was a static object and so `ctx.questionnaire` could change when requests happened at the same time.

For example a breaking scenario:

| Order of events | Notes |
|-----------------|-------|
| cSI-read | Good - got the right questionnaire |
| uS-read-start| Kicked off async action and then queued |
| cSI-write | Kicked off write let node do more |
| uS-read-run | Returned the value it got (pre-write). Context now wrong |
| cSI-return | Context returning old value in resolvers |

### Fix
1. Update middleware to store values on req
2. Update to apollo-server 2 so we can read req to generate a context object for each request.

### Visible Recreation steps
1. Check out https://github.com/ONSdigital/eq-author-app/tree/logging-static
1. Run the builder spec and see the SectionIntro tests fail (it doesn't always happen)
1. You should be able to see that is a log of the write updating the created at (e.g. `updating from 123 to 125`) and then a read of the old updated at time (e.g. `read 123`). Also you will see the section intro create resolver log showing the old updated at (e.g. `Read section from section intro 123`)

### How to review 
1. Ensure that section introduction cypress tests (a part of build spec) always pass.

### DIFF
--- a/eq-author-api/middleware/loadQuestionnaire.js
+++ b/eq-author-api/middleware/loadQuestionnaire.js
@@ -1,14 +1,14 @@
 const { getQuestionnaire } = require("../utils/datastore");
 
-module.exports = (logger, context) => async (req, res, next) => {
+module.exports = async (req, res, next) => {
   const questionnaireId = req.header("questionnaireId");
   if (questionnaireId) {
-    context.questionnaire = await getQuestionnaire(questionnaireId);
-    if (!context.questionnaire.metadata) {
-      context.questionnaire.metadata = [];
+    req.questionnaire = await getQuestionnaire(questionnaireId);
+    if (!req.questionnaire.metadata) {
+      req.questionnaire.metadata = [];
     }
-    if (!context.questionnaire.sections) {
-      context.questionnaire.sections = [];
+    if (!req.questionnaire.sections) {
+      req.questionnaire.sections = [];
     }
   }
   next();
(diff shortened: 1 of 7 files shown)


tgandrews:
[Solve issue of concurrent editing](https://trello.com/c/GLPlLgQx/929-solve-issue-of-concurrent-editing)

eq-author-api/middleware/loadQuestionnaire.js
```diff
@@ -1,14 +1,14 @@
 const { getQuestionnaire } = require("../utils/datastore");
 
-module.exports = (logger, context) => async (req, res, next) => {
+module.exports = () => async (req, res, next) => {
```
SamGodwin2: Don't think we need to pass in logger or ctx on line 38 in app.js with this change.

tgandrews:
@SamGodwin2 please re-review.

PR #207, opened 2019-02-13
Add env to switch between DynamoDB and local filesystem.

rmccar:
Currently without a decent way of interacting with dynamoDB this is a workaround so that an environment variable can be set so that in dev we can manage the questionnaires as json files as before DynamoDB. This makes testing questionnaires a lot easier.

### How to review 
- Make sure tests pass
- Add 'DATASTORE=filesystem' to your environment variables in the api docker-compose.yml and ensure that author now goes back to writing to json files in the data folder

### DIFF
--- a/eq-author-api/utils/datastore.js
+++ b/eq-author-api/utils/datastore.js
@@ -1,90 +1,17 @@
+let datastore;
+if (process.env.DATASTORE === "filesystem") {
+  datastore = require("./datastoreFileSystem");
+} else {
+  datastore = require("./datastoreDynamo");
+}
+
 const {
-  QuestionnaireModel,
-  QuestionnaireVersionsModel,
-} = require("../db/models/DynamoDB");
-const { omit, isEqual } = require("lodash");
-
-const saveModel = model =>
-  new Promise((resolve, reject) => {
-    model.save(err => {
-      if (err) {
-        reject(err);
-      }
-      resolve();
-    });
-  });
-
-const createQuestionnaire = async questionnaire => {
-  await saveModel(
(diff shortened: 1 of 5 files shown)


eq-author-api/utils/datastoreFileSystem.js
```diff
@@ -0,0 +1,55 @@
+const { omit, reject } = require("lodash");
+const fs = require("fs");
+const stringify = require("json-stable-stringify");
+
+const createQuestionnaire = async questionnaire => {
+  fs.writeFileSync(
```
tgandrews: Can we use `async` versions so the differences aren't too big. https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_promises_api

eq-author-api/utils/datastore.js
```diff
@@ -1,90 +1,12 @@
...
-        resolve(questionnaires);
-      });
-  });
-};
+  createQuestionnaire,
+  saveQuestionnaire,
+  deleteQuestionnaire,
+  getQuestionnaire,
+  listQuestionnaires,
+} = process.env.DATASTORE_FILESYSTEM
```
jonnyshaw89: Could better option to make this non-binary?
so have `process.env.DATASTORE` that could equal `filesystem`, `dynamodb`, `some googleness`?

jonnyshaw89: Should also add documentation to the Readme

rmccar: Yep ill do that

rmccar:
[Extract data from pre-prod questionnaires](https://trello.com/c/PiKibqbr/923-extract-data-from-pre-prod-questionnaires)

PR #211, opened 2019-02-18
Fix UUID regex to be backwards compatable

rmccar:
This PR is to change the current UUID regex to allow it to be backwards compatible with the old ID and URLs. This will then be changed at some point in a refactor.

### How to review 
- Check tests pass
- Check that current survey ids are accepted


rmccar:

PR #212, opened 2019-02-19
Log when client and API versions are different

SamGodwin2:

There seems to currently be a possibility for the api and front-end to get into version mismatches when the front end gets cached by the user. This PR adds the cache control headers to nginx and also adds logging to help alert us when this happens and if it's an issue that needs handling. 

### How to review 
Changes make sense and if the api is running in a non-dev NODE_ENV you should see logs alerting you that there is a "Application version mismatch".

### DIFF
--- a/eq-author-api/app.js
+++ b/eq-author-api/app.js
@@ -9,6 +9,7 @@ const { PORT } = require("./settings");
 const status = require("./middleware/status");
 const { getLaunchUrl } = require("./middleware/launch");
 const createAuthMiddleware = require("./middleware/auth");
+const versionMiddleware = require("./middleware/version");
 const repositories = require("./repositories");
 const modifiers = require("./modifiers");
 const schema = require("./schema");
@@ -60,6 +61,7 @@ db(process.env.DB_SECRET_ID)
       pino,
       cors(),
       authMiddleware,
+      versionMiddleware,
       bodyParser.json(),
       graphqlExpress({
         schema,
--- a/eq-author-api/middleware/version.js
+++ b/eq-author-api/middleware/version.js
@@ -1,0 +1,16 @@
+module.exports = (req, res, next) => {
+  if (
+    req.headers.clientversion &&
+    process.env.EQ_AUTHOR_API_VERSION &&
+    req.headers.clientversion !== process.env.EQ_AUTHOR_API_VERSION
+  ) {
+    req.log.warn(
+      {
+        clientversion: req.headers.clientversion,
+        apiversion: process.env.EQ_AUTHOR_API_VERSION,
+      },
+      "Application version mismatch"
+    );
+  }
+  next();
+};
--- a/eq-author/src/index.js
+++ b/eq-author/src/index.js
@@ -11,7 +11,7 @@ import { setContext } from "apollo-link-context";
 import config from "config";
 import getIdForObject from "utils/getIdForObject";
 import render from "utils/render";
-import appendAuthHeader from "utils/appendAuthHeader";
+import getHeaders from "middleware/headers";
 
 import App from "App";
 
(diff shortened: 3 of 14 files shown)


eq-author-api/app.js
```diff
@@ -35,6 +35,22 @@ db(process.env.DB_SECRET_ID)
 
     const authMiddleware = createAuthMiddleware(logger, context);
 
+    const versionMiddleware = (req, res, next) => {
```
tgandrews: Could we extract this to a separate file with tests?

eq-author/src/index.js
```diff
@@ -43,11 +43,20 @@ const history = createHistory();
 
 const httpLink = createHttpLink(config.REACT_APP_API_URL);
 
+const versionLink = setContext((_, { headers }) => ({
```
tgandrews: This might be worth extracting as well so we can test it? Maybe create a middleware folder so the questionnaire id and auth headers are handled there as well?

eq-author-api/app.js
```diff
@@ -35,6 +35,22 @@ db(process.env.DB_SECRET_ID)
...
+      if (
+        process.env.NODE_ENV !== "development" &&
+        req.headers.clientversion !== process.env.EQ_AUTHOR_API_VERSION
+      ) {
+        logger.info(
+          {
+            clientversion: req.headers.clientversion,
+            apiversion: process.env.EQ_AUTHOR_API_VERSION,
```
tgandrews: Should we send this in all responses? Or is the status page enough. I think it probably is.

eq-author-api/middleware/version.js
```diff
@@ -0,0 +1,15 @@
+module.exports = (req, res, next) => {
+  if (
+    process.env.NODE_ENV !== "development" &&
+    req.headers.clientversion !== process.env.EQ_AUTHOR_API_VERSION
+  ) {
+    req.log.info(
```
tgandrews: Should it be `warn`?

eq-author/src/middleware/headers/versionHeader.js
```diff
@@ -0,0 +1,4 @@
+export default headers => ({
+  ...headers,
+  clientVersion: process.env.REACT_APP_EQ_AUTHOR_VERSION,
```
samiwel: Probably best to only set the request header if `REACT_APP_EQ_AUTHOR_VERSION` has a value.

samiwel: image

eq-author-api/middleware/version.js
```diff
@@ -0,0 +1,15 @@
+module.exports = (req, res, next) => {
+  if (
+    process.env.NODE_ENV !== "development" &&
```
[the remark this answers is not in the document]
SamGodwin2: Yeah the reason for this is because it just causes a bunch of additional logging noise every time you make a request to the API as while in dev the version numbers aren't present, this seemed like a better solution than putting dummy version numbers in the envs?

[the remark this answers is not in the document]
samiwel: If we change to something like this it won't log in the event that versions aren't set. No need for dummy versions to be entered in dev.

```js
if (req.headers.clientversion && process.env.EQ_AUTHOR_API_VERSION && req.headers.clientversion !== process.env.EQ_AUTHOR_API_VERSION) {
    req.log.warn(
      {
        clientversion: req.headers.clientversion,
        apiversion: process.env.EQ_AUTHOR_API_VERSION,
      },
      "Application version mismatch"
    );
  }
  next();
};
```

samiwel:
[Fix browser caching old version of HTML/JS (2)](https://trello.com/c/mAVwAURP/848-fix-browser-caching-old-version-of-html-js-2)

samiwel:
Coverage dropped, I think you just need to test that the `clientversion` header if env var is not set.

PR #214, opened 2019-02-19
Fix validation on update

rmccar:
The front end wasn't being updated when the validation rule using a previous answer was being changed. It was because the validation type could not be resolved. This was causing the app to get in an odd state if you did not refresh after each change. This change adds the validation type on to the object when a validation rule is updated

### How to review 
- Tests pass
- Make sure changing validation works using a previous answer and the page updates without a refresh


rmccar:

PR #215, opened 2019-02-20
Create a questionnaire builder and schema tests

adammahmood:
- Adds schema tests to API
- This is still a WIP but all tests in schema/tests pass

### How to review 
- Ensure tests pass


adammahmood:

eq-author-api/utils/loadQuestionnaire.js
```diff
@@ -0,0 +1,4 @@
+const fs = require("fs");
```
tgandrews: I don't think this file is used.

eq-author-api/utils/saveQuestionnaire.js
```diff
@@ -0,0 +1,10 @@
+const fs = require("fs");
```
tgandrews: I don't think this file is used.

PR #216, opened 2019-02-20
Add import questionnaire script

rmccar:
This is a new script to be used to import questionnaires from local to dynamoDB. It also adds a switch to the filesystem script because when running this new script it will be run from a different location so needs slightly different paths

### How to review 
- Tests pass
- Script can be run and duplicates the files in the data folder

### DIFF
--- a/eq-author-api/scripts/import-json-to-dynamo.js
+++ b/eq-author-api/scripts/import-json-to-dynamo.js
@@ -1,0 +1,26 @@
+#!/usr/bin/env node
+/* eslint-disable import/unambiguous,no-console */
+const fs = require("fs");
+const jsonPath = process.env.DATA_DIR;
+const { createQuestionnaire } = require("../utils/datastore");
+
+if (!jsonPath) {
+  throw new Error("Set DATA_DIR environment variable");
+}
+
+const importQuestionnaire = async contents => {
+  return createQuestionnaire({ ...contents, id: contents.id + "_copy" });
+};
+
+const loadQuestionnaireJSON = async path => {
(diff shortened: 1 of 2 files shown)


rmccar:

eq-author-api/scripts/import-json-to-dynamo.js
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env node
```
tgandrews: The filename is dynamo but the script imports to file system?

eq-author-api/scripts/import-json-to-dynamo.js
```diff
@@ -0,0 +1,26 @@
...
+const fs = require("fs");
+const jsonPath = process.env.DATA_DIR;
+const { createQuestionnaire } = require("../utils/datastoreFileSystem");
+
+if (!jsonPath) {
+  throw new Error("Set DATA_DIR environment variable");
+}
+
+const importQuestionnaire = async contents => {
+  return createQuestionnaire({ ...contents, id: contents.id + "_copy" });
```
tgandrews: Is that a valid questionnaire id? Or will it clash with our UI routing regex?

eq-author-api/scripts/import-json-to-dynamo.js
rmccar: At the moment it will import to file system for testing but will eventually import to Dynamo

tgandrews: Can we not call dynamo yet?

rmccar: Yes ive made a change now to allow this

PR #219, opened 2019-02-21
Fix travis build

samiwel:
We had an issue where travis was timing out because of the docker-compose up command not producing any output within a specified time.

docker-compose up attempts to pull, build and start the services. At the point where the integration tests are being run on travis the containers should already have been built.
This change explicitly tells docker-compose not to attempt to re-create and build the containers and pulls only the database container.

### How to review 
Should reach integration tests and start the cypress testing on travis.


samiwel:

PR #220, opened 2019-02-21
Fix validation, duplication and properties test

adammahmood:
- Fixes validation, duplication and properties tests

### How to review 
- Ensure tests pass


adammahmood:

eq-author-api/schema/tests/duplication.test.js
```diff
@@ -0,0 +1,158 @@
...
+    let page, pageCopy;
+
+    beforeEach(async () => {
+      page = await queryQuestionPage(questionnaire, last(section.pages).id);
+      let { id } = await duplicatePage(questionnaire, page);
+      pageCopy = await queryQuestionPage(questionnaire, id);
+    });
+
+    afterEach(async () => {
+      await deleteQuestionPage(questionnaire, pageCopy.id);
```
tgandrews: I'm not sure I like sharing questionnaires across tests. I think we should discuss this after. My worry is that we could have state leaking between tests

eq-author-api/schema/tests/duplication.test.js
```diff
@@ -0,0 +1,158 @@
...
+      duplicatedQuestionnaire = await duplicateQuestionnaire(questionnaire);
+      questionnaireCopy = await queryQuestionnaire(duplicatedQuestionnaire);
+    });
+
+    afterEach(async () => {
+      await deleteQuestionnaire(questionnaireCopy.id);
+    });
+
+    it("should copy questionnaire with sections and pages", () => {
+      //@todo - Some sort of issue with createdBy resolving correctly
```
tgandrews: I fixed this so that it would work for create questionnaire. You can see how that works. The auth value is currently not configurable though

eq-author-api/schema/tests/properties.test.js
```diff
@@ -0,0 +1,129 @@
...
+      type: DATE,
+    });
+    const properties = { format: "mm/yy" };
+    await updateAnswer(questionnaire, { id, properties });
+    const { properties: updatedProperties } = await queryAnswer(
+      questionnaire,
+      id
+    );
+
+    expect(updatedProperties).toEqual(expect.objectContaining(properties));
```
tgandrews: Instead of `toEqual(expect.objectContaining` can we use `toMatchObject`?

eq-author-api/schema/tests/validation.test.js
```diff
@@ -166,13 +92,23 @@ describe("validation", () => {
 
   describe("Number and Currency", () => {
     it("should create min and max validation db entries for Currency and Number answers", async () => {
```
tgandrews: Can we drop `db entries` here?

PR #225, opened 2019-02-25
Don't rely on buildQuestionnaire to run create mutation

tgandrews:
Test creation instead of relying buildTestQuestionnaire to do it. This had a knock on effect which resulted in removing any reliance on one questionnaire for the test file and instead mostly creating one per test. The cost seems relatively minimal at the moment ~20-30ms on my machine.

Also, fixed some todos such as mutually exclusive options.

One minor thing was a change to `executeQuery` so it will throw if it reads any graphql errors. We might want to make this configurable later but is probably a good catch for now.

### How to review 
1. Ensure we are red for the right reasons
2. Ensure the test suite does not take more than ~10 seconds locally.

### DIFF
--- a/eq-author-api/schema/resolvers/base.js
+++ b/eq-author-api/schema/resolvers/base.js
@@ -207,6 +207,10 @@ const Resolvers = {
         }
       });
 
+      if (!confirmationPage) {
+        return null;
+      }
+
       return { pageId, ...confirmationPage };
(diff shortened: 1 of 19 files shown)


tgandrews:

tgandrews:
@adammahmood this issue is resolved in #226

PR #226, opened 2019-02-25
Fix validation bug

SamGodwin2:
There was a bug found that was introduced in https://github.com/ONSdigital/eq-author-app/pull/214 where trying to toggle a validation or update it would not update the questionnaires data. This was because the function changed in that PR caused a new object to be returned instead of the reference of the entity within the questionnaire. This PR reverts this change and allows you to toggle and update the questionnaire.

### How to review 
You can update the validations on a questionnaire and tests pass.

### DIFF
--- a/eq-author-api/constants/validationTypes.js
+++ b/eq-author-api/constants/validationTypes.js
@@ -14,23 +14,6 @@ const VALIDATION_TYPES = [
   MAX_DURATION,
 ];
 
-const MIN_VALUE_INPUT = "minValueInput";
-const MAX_VALUE_INPUT = "maxValueInput";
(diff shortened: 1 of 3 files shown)


SamGodwin2:

SamGodwin2:
Yeah was planning on talking it over on Wednesday I've got an idea on how we could neaten this stuff up a bit but I don't think it's worth doing till we're in a good state.

tgandrews:
@SamGodwin2 @samiwel why is travis not running on this?

SamGodwin2:
Looks like Travis isn't running on any of the PR's atm?

Issue #208, opened 2019-02-14
Metadata dialogue permits spaces in Key field


jchapman68:
### Expected behaviour
Key values entries should be checked to ensure they match regex '^[a-zA-Z0-9-_]+$'

### Actual behaviour
Any text value can be entered including punctuation

### Steps to reproduce the behaviour
Select Metadata dialogue
Select "Add metadata"
Enter any 'Test??  Value[]"

#### Browser
Chrome

#### Operating System
Mac OS

### Screenshot
![image](https://user-images.githubusercontent.com/30293265/52779731-a7a29100-3040-11e9-8801-6a153d72c051.png)

PR #228, opened 2019-02-26
Fixed metadata key field to only accept desired inputs

SamGodwin2:
Set up the key select field on the metadata modal on only accept input that matches ^[a-zA-Z0-9-_]+$.

### How to review 
Tests pass and yo can't enter spaces or punctuation into the key select for metadata.

fixes https://github.com/ONSdigital/eq-author-app/issues/208

### DIFF
--- a/eq-author/src/App/MetadataModal/MetadataTable/Controls/KeySelect.js
+++ b/eq-author/src/App/MetadataModal/MetadataTable/Controls/KeySelect.js
@@ -1,6 +1,5 @@
 import PropTypes from "prop-types";
 import React, { Component } from "react";
-import Downshift from "downshift";
 import { reject, includes, isUndefined } from "lodash";
 
 import Typeahead from "components/Forms/Typeahead";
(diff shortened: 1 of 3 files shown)


samiwel:
[Metadata dialogue permits spaces in Key field](https://trello.com/c/iux1tMM7/969-metadata-dialogue-permits-spaces-in-key-field)

eq-author/src/App/MetadataModal/MetadataTable/Controls/KeySelect.js
```diff
@@ -34,9 +34,14 @@ class KeySelect extends Component {
   };
 
   getChangeValue = changes => {
+    const regex = /^[a-z0-9-_]+$/i;
```
samiwel: Not sure we should have the ignore case flag. Metadata keys should ideally be lower case.

eq-author/src/App/MetadataModal/MetadataTable/Controls/KeySelect.js
```diff
@@ -34,9 +34,14 @@ class KeySelect extends Component {
...
     if (changes.hasOwnProperty("selectedItem")) {
       return changes.selectedItem.value;
-    } else if (changes.hasOwnProperty("inputValue")) {
+    } else if (
+      (changes.hasOwnProperty("inputValue") &&
```
samiwel: Will this work if it was just

```js
else if (changes.inputValue) {
 ...
}
```

eq-author/src/App/MetadataModal/MetadataTable/Controls/KeySelect.js
```diff
@@ -34,9 +34,14 @@ class KeySelect extends Component {
...
+        regex.test(changes.inputValue)) ||
+      !changes.inputValue
```
samiwel: This feels a bit weird. In the case of `inputValue` being `undefined`, the `else if` will still evaluate to true so we return `undefined`?

eq-author/src/App/MetadataModal/MetadataTable/Controls/KeySelect.js
samiwel: Also possibly move creating this regex down to where it's being used in the else if?

PR #224, opened 2019-02-22
Run tests against DynamoDB.

samiwel:
### What is the context of this PR?
Until now the feature branch has been running tests against the filesystem. The preferred way of running the tests are against DynamoDB since this is what will be used initially within AWS.

This change spins up a containerised DynamoDB local instance when running the tests and makes slight changes to the application in order to get the tests passing.

### How to test
The API tests should pass consistently when running `yarn test`.
Should observe that the tests are running against DynamoDB.

PR #229, opened 2019-02-26
Run cypress against dynamo in CI

tgandrews:
This builds on #224 to get the cypress tests to pass

### How to review 
1. It should all go green 🤞


tgandrews:

Issue #209, opened 2019-02-14
Cursor jumps to end of field when editing Key field name 


jchapman68:
### Expected behaviour
The cursor should remain at the edit point when editing the text of the Key field

### Actual behaviour
When editing the middle of a Key name the cursor jumps to the end of the field after each character inserted  or backspace/delete key pressed

### Steps to reproduce the behaviour
Open the Metadata dialogue box
Select a Key value
Place cursor in middle of Key value either using mouse or arrow keys
Insert an alphanumeric character or press backspace or delete key to edit the Key value

PR #231, opened 2019-02-26
Cursor no longer jumps to end of input on input

SamGodwin2:
As shown in the below issue, when editing a key field on metadata the cursor would jump to the end of the input whenever the input was changed.

### How to review 
Tests continue to pass and the bug has been observed to be fixed.

fixes https://github.com/ONSdigital/eq-author-app/issues/209


SamGodwin2:
[Cursor jumps to end of field when editing Key field name](https://trello.com/c/zfJYGUGO/971-cursor-jumps-to-end-of-field-when-editing-key-field-name)

jchapman68:
I have tested the change and works as expected, but why does removing the lines in the files setting the value fix the issue?

PR #233, opened 2019-02-27
913 remove unused tests

adammahmood:
- Removes unused tests
- <strike>Intentionally ignoring `db/index.test.js` due to strange behaviour with `jest` mocking / unmocking.  We will solve this problem separately. </strike> This is being removed as part of https://github.com/ONSdigital/eq-author-app/pull/232 and will no longer be a problem.
- Gives us a clearer idea of our coverage

### How to review 
- Run tests
- Review coverage


adammahmood:

tgandrews:
Woah! 16k down! 🎉 

Looks like cypress isn't starting. I presume a container isn't starting up. This might be because it is still referencing postgres? If so, I guess that would be resolved by #232

PR #236, opened 2019-02-28
Fix CI pushing docker images with incorrect tag

samiwel:
We noticed that the 797 branch pushed its container image to docker hub with the `latest` tag. This is incorrect and could lead to some undesirable behaviour when trying to run Author using docker.

The branch should push up an image tagged with the branch name.

### How to review 
Observe in the travis logs that the branch name and tag are outputted in the console.
If the branch is master then the tag should be "latest".
When the branch is some other branch then the tag should reflect the branch name.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,11 @@ cache:
 
 before_script:
   - set -e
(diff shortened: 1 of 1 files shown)


samiwel:
Codacy seems to have stalled so I'm going to merge this PR on the basis it has 3 approvals and passing CI.

samiwel:
[Fix pushing docker images with incorrect tag](https://trello.com/c/JYkVIurj/980-fix-pushing-docker-images-with-incorrect-tag)

PR #238, opened 2019-03-01
Fix for import script to work with Dynamo

rmccar:
This is a fix to have the import script work with DynamoDB. The "_copy" needed to be removed from testing and the updated at needed to be removed for it to line up with the DB schema.

### How to review 
- Tests Pass
- Script runs and imports files into Dynamo

### DIFF
--- a/eq-author-api/scripts/import-json-to-dynamo.js
+++ b/eq-author-api/scripts/import-json-to-dynamo.js
@@ -9,7 +9,7 @@ if (!jsonPath) {
 }
 
 const importQuestionnaire = async contents => {
-  return createQuestionnaire({ ...contents, id: contents.id + "_copy" });
+  return createQuestionnaire({ ...contents, id: contents.id });
 };
 
 const loadQuestionnaireJSON = async path => {
--- a/eq-author-api/db/models/DynamoDB.js
+++ b/eq-author-api/db/models/DynamoDB.js
@@ -9,11 +9,11 @@ if (process.env.DYNAMO_ENDPOINT_OVERRIDE) {
   throughput = { read: 10, write: 10 }; // DynamoDB local doesn't yet support on-demand
 }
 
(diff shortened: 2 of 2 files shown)


eq-author-api/scripts/import-json-to-dynamo.js
```diff
@@ -9,7 +9,7 @@ if (!jsonPath) {
 }
 
 const importQuestionnaire = async contents => {
-  return createQuestionnaire({ ...contents, id: contents.id + "_copy" });
+  return createQuestionnaire({ ...contents, id: contents.id });
```
tgandrews: So this was the only bit of the problem yesterday? It can't have `_`? Or was that it wasn't a hex value?

rmccar: It was the `_` the regex creating the url was ignoring it

rmccar: That plus the updated at stuff was also causing another issue

tgandrews:
Do we have some files to test?

rmccar:
Ive added the files I have been using to test to a zip file on the trello card https://trello.com/c/a7kUASvc/797-replace-relational-database-with-json-schema

Also to run the script I have been using this line:
`AWS_REGION=eu-west-1 AWS_ACCESS_KEY_ID=dummy AWS_SECRET_ACCESS_KEY=dummy DYNAMO_QUESTIONNAIRE_TABLE_NAME=dev-author-questionnaires DYNAMO_QUESTIONNAIRE_VERSION_TABLE_NAME=dev-author-questionnaire-versions DYNAMO_ENDPOINT_OVERRIDE=http://localhost:8050 DATA_DIR=../data  node import-json-to-dynamo.js`

tgandrews:
Could we use this to seed cypress? That would be a decent way of proving it works as well?

tgandrews:
It looks like this needs rebasing @rmccar

PR #239, opened 2019-03-01
Fix bug where empty tag set if not on master branch

samiwel:
There was a bug in the existing logic where the tag used when pushing up docker images would be empty if a branch other than master was built. This was observed in a [recent build of the 797 feature branch on Travis](https://travis-ci.com/ONSdigital/eq-author-app/builds/102790810#L476).

This change sets the tag to be the branch name if the branch being built is not master and it is not a pull request.

### How to review 
Change should look sensible. When master is built, tag should be "latest".
On this PR the tag should be "980-fix-pushing-docker-images-with-incorrect-tag"
When merged into 797 the tag should be "797-replace-relational-database-with-json-document"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,8 @@ cache:
 
 before_script:
   - set -e
-  - export TAG=`if [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$TRAVIS_BRANCH" == "master" ]; then echo "latest"; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi`
+  - if [ "$TRAVIS_BRANCH" == "latest" ]; then echo "Building branches named latest not allowed as it would overwrite the stable docker image. Please rename your branch." && exit 1; fi;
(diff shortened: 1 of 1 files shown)


samiwel:
This fix works for now, but raises the question of what happens if someone pushes up a branch called "latest". I am trusting the team not to do that!

tgandrews:
Should we error in that `latest` case?

samiwel:
@tgandrews Added a guard against this possibility in my latest commit.

.travis.yml
```diff
@@ -16,7 +16,8 @@ cache:
 
 before_script:
   - set -e
-  - export TAG=`if [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$TRAVIS_BRANCH" == "master" ]; then echo "latest"; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi`
+  - if [ "$TRAVIS_BRANCH" == "latest" ]; then echo "Building branches named latest not allowed. Please rename your branch." && exit 1; fi;
```
tgandrews:
```suggestion
  - if [ "$TRAVIS_BRANCH" == "latest" ]; then echo "Building branches named latest not allowed as it would overwrite the stable docker image. Please rename your branch." && exit 1; fi;
```

samiwel:
[Fix pushing docker images with incorrect tag](https://trello.com/c/JYkVIurj/980-fix-pushing-docker-images-with-incorrect-tag)

samiwel:
I'm merging this as codacy seems to have stalled and it has 2 reviews.

PR #240, opened 2019-03-01
Fix routing error

rmccar:
There is an error in pre-prod that is unable to load a routing rule if it has been changed, navigated to a different tab and back again. We found that it wasn't returning the destination ID so this PR fixes that.

### How to review 
- Tests pass
- Can change a routing rule else destination, navigate to the design tab and back again and the routing rule is loaded correctly


tgandrews:
@rmccar can you provide detailed recreation steps. I'm not sure I understand.

SamGodwin2:
@tgandrews 

1. Create a questionnaire with two pages
2. Make a routing on the first question
3. Change the else from "Next Page" to either the second page or the EOQ
4. Navigate to Design then back to Routing, should see the "something went wrong page"

PR #243, opened 2019-03-04
Bug - Unable to update other options

tgandrews:
Fix issue where not able to update other answer labels or description as it tried to send the answer type of the additional answer which is no longer allowed.

### Recreation steps
1. Create a questionnaire
2. Add a checkbox answer
3. Add an "other" option
4. Change the other label or description
5. Blur the field - the red bar should show. And the change is not persisted (you can check with a refresh)

### How to review 
1. That this PR resolves the issue.


tgandrews:
[New DB - Unable to other option labels and description](https://trello.com/c/oET7TEAs/984-new-db-unable-to-other-option-labels-and-description)

PR #245, opened 2019-03-05
WIP - Blow up cypress when graphql errors

tgandrews:
This change is to throw an error when a network or server error occurs. This in turn causes cypress to blow up.

This is to catch errors that have been occurring in cypress but not causing our tests to fail.

### How to review 
Don't yet but help to fix all the failing tests would be appreciated.


tgandrews:
[Fail Cypress when an API error occurs](https://trello.com/c/lMfCy7Im/982-fail-cypress-when-an-api-error-occurs)

tgandrews:
This will be looked at later.

PR #246, opened 2019-03-05
Stop sending empty updates

tgandrews:
This fixes an issue where we would send updates unnecessarily because of a bug in wEE. 

### How to review 
1. Create a page
1. Set the title to some value
1. Blur the field
1. Focus and blur the field
1. A second update should not be sent. With this PR it isn't


tgandrews:
[Sending updates when nothing is changing](https://trello.com/c/iJVN0kXx/981-sending-updates-when-nothing-is-changing)

PR #247, opened 2019-03-05
Add Hotjar

tgandrews:
Adds hotjar tracking script which is enabled when a hot jar environment variable is provided as part of the docker launch.

This also tidies up the `script` tags that were sitting between the `head` and the `body` tags and being interpreted as being in the head by Chrome. So they now all correctly sit in the `head`.

### How to review 
1. Ensure that the tracking code is as per hotjar documentation. (see me for account details)
1. Ensure the hotjar script runs when the hot jar id is provided as `REACT_APP_HOT_JAR_ID`. The easiest way to see this is in `author` to run `yarn build && docker build .`. Then you can run the container: `docker run -p 3003:3000 -e REACT_APP_HOT_JAR_ID=123 [[container id from docker build]]`. If you now hit http://localhost:3003 it should make calls to request hotjar sources.

### DIFF
--- a/eq-author/public/index.html
+++ b/eq-author/public/index.html
@@ -25,42 +25,55 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>Author</title>
-  </head>
-  <script type="text/javascript">
-    window.config = {};
-    window.config.REACT_APP_API_URL="";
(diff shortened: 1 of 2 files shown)


tgandrews:
[Add Hotjar (3)](https://trello.com/c/ZIrVAwlq/927-add-hotjar-3)

PR #250, opened 2019-03-06
Add Codeowners

tgandrews:
I'm getting a little frustrated having to add everyone as a reviewer each time. I hope this resolves this. 

It follows the documentation: https://help.github.com/en/articles/about-code-owners

### How to review 
1. Ensure that everyone is included.
2. Ensure that it matches the format specified in the documentation.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,2 @@
+*             @samiwel @rmccar @jchapman68 @heofs @samgodwin2 @adammahmood @tgandrews
+eq-author/*   @samiwel @rmccar @jchapman68 @heofs @samgodwin2 @adammahmood @tgandrews @hamishtaplin


tgandrews:
@jonnyshaw89 so that get it in their list on github to review and an email.

samiwel:
Worth noting that I don't usually get email notification when someone requests a review. I have turned email notifications off for the most part or I would just get spammed with tons of email all the time.

samiwel:
Also this is another file that now has to be managed when someone joins/leaves the team.

adammahmood:
I usually add reviewers based on the type of work completed. eg. If it's UI based then always request @hamishtaplin.  I'm not strongly opinionated enough on this change so will let others in the team decide if this is beneficial enough to merge. I think having to maintain the code owners will be a pain!

tgandrews:
I tend to use the github view as the list of PRs I need to review so require being assigned to pick it up there. So maybe its only a thing for my workflow. Happy to not include as I understand it requires some extra effort to maintain when people join/leave.

tgandrews:
Only for my workflow so no need to burden the team

PR #252, opened 2019-03-08
Migrate JSON data versions as application changes

adammahmood:
- Migrates JSON data versions as application changes.
- Allows us to transform a questionnaire over time.
- Migrations are run (if needed) on every request via middleware

### How to review 
- Ensure tests pass
- Ensure the correct migrations are run ie. already executed migrations are not run again
- Ensure you can create migrations as per README instructions

### DIFF
--- a/eq-author-api/README.md
+++ b/eq-author-api/README.md
@@ -9,6 +9,7 @@ application.
 - [Environment Variables](#environment-variables)
 - [Run using Docker](#run-using-docker)
 - [Tests](#tests)
+- [Migrations](#migrations)
 - [Debugging (with VS Code)](#debugging-with-vs-code)
 - [Importing Questionnaires)](#importing-questionnaires)
 
@@ -104,6 +105,14 @@ There are [queries](tests/fixtures/queries.gql) and [example data](tests/fixture
 `yarn test` will start a single run of unit and integration tests.
 
 `yarn test --watch` will start unit and integration tests in watch mode.
+
+## Migrations
+
+`runQuestionnaireMigrations` middleware is responsible for updating the schema version and running any necessary migrations.
+
+-  `yarn create-migration [name]` will create a new migration in the `/migrations` directory.
+-  Add the created migration to the bottom of the `migrations/index.js` array.
+-  `runQuestionnaireMigrations` will execute necessary migrations on every request providing the schema has not already been migrated.
 
 ## Debugging (with VS Code)
 
--- a/eq-author-api/middleware/runQuestionnaireMigrations.test.js
+++ b/eq-author-api/middleware/runQuestionnaireMigrations.test.js
@@ -1,0 +1,100 @@
+const runQuestionnaireMigrations = require("./runQuestionnaireMigrations");
+const { buildQuestionnaire } = require("../tests/utils/questionnaireBuilder");
+
+describe("runQuestionnaireMigrations", () => {
+  let res, req, next, migrations, logger;
+
+  beforeEach(async () => {
+    req = {
+      questionnaire: await buildQuestionnaire({
+        metadata: [{}],
+      }),
+    };
+    res = {};
+    next = jest.fn();
+    migrations = [];
+    logger = {
+      info: jest.fn(),
+    };
+  });
+
+  it("should correctly handle no questionnaire", async () => {
+    req = {};
+    let migrationOne = jest.fn(() => req.questionnaire);
+    migrations = [migrationOne];
+    await runQuestionnaireMigrations(logger)(migrations)(req, res, next);
+    expect(next).toHaveBeenCalled();
+    expect(migrationOne).not.toHaveBeenCalled();
+  });
+
+  it("should correctly handle an undefined version", async () => {
+    req.questionnaire.version = undefined;
+    let migrationOne = jest.fn(() => req.questionnaire);
+    migrations = [migrationOne];
+
+    await runQuestionnaireMigrations(logger)(migrations)(req, res, next);
+
+    expect(next).toHaveBeenCalled();
+    expect(migrationOne).toHaveBeenCalledWith(req.questionnaire);
+    expect(req.questionnaire.version).toEqual(1);
+  });
+
+  it("should not migrate data if version up to date", async () => {
+    req = {
+      questionnaire: {
+        version: 2,
+      },
+    };
+    let migrationOne = jest.fn(() => req.questionnaire);
+    let migrationTwo = jest.fn(() => req.questionnaire);
+    migrations = [migrationOne, migrationTwo];
+    await runQuestionnaireMigrations(logger)(migrations)(req, res, next);
+    expect(next).toHaveBeenCalled();
+    expect(migrationOne).not.toHaveBeenCalled();
+    expect(migrationTwo).not.toHaveBeenCalled();
+  });
+
+  it("should not update version if migration fails", async () => {
+    req.questionnaire.version = 1;
+    let migrationOne = jest.fn(() => req.questionnaire);
+    let migrationTwo = jest.fn(() => req.questionnaire);
+    let migrationThree = jest.fn(() => {
+      throw new Error("foobar");
+    });
+    migrations = [migrationOne, migrationTwo, migrationThree];
+
+    await runQuestionnaireMigrations(logger)(migrations)(req, res, next);
+    expect(next).toHaveBeenCalled();
+    expect(migrationOne).not.toHaveBeenCalled();
+    expect(migrationTwo).toHaveBeenCalled();
+    expect(migrationThree).toThrowError();
+    expect(logger.info).toHaveBeenCalledTimes(3);
+    expect(req.questionnaire.version).toEqual(1);
+  });
+
+  it("should migrate data and set version correctly", async () => {
+    req.questionnaire.version = 1;
+
+    let migrationOne = jest.fn(() => req.questionnaire);
+    let migrationTwo = jest.fn(questionnaire => {
+      questionnaire.title = "foo";
+      return questionnaire;
+    });
+    let migrationThree = jest.fn(questionnaire => {
+      questionnaire.description = "bar";
+      return questionnaire;
+    });
+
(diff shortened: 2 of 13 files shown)


adammahmood:
[Migrate JSON data versions as application changes. (9)](https://trello.com/c/k9aPKt4X/836-migrate-json-data-versions-as-application-changes-9)

eq-author-api/package.json
```diff
@@ -53,7 +54,8 @@
       "!scripts/*",
       "!middleware/import.js",
       "!app.js",
-      "!utils/datastoreFileSystem.js"
+      "!utils/datastoreFileSystem.js",
+      "!utils/migrations/*"
```
tgandrews: I would say we should cover this

eq-author-api/utils/migrations/createMigration.js
```diff
@@ -0,0 +1,21 @@
+const fs = require("fs").promises;
+
+if (process.argv.length < 3) {
+  /* eslint-disable-next-line no-console */
+  console.info(
+    "\x1b[33m",
```
tgandrews: Could we use [chalk[(https://www.npmjs.com/package/chalk) instead

eq-author-api/package.json
```diff
@@ -5,11 +5,12 @@
...
   "scripts": {
     "start": "node app.js",
-    "start:dev": "NODE_ENV=development nodemon --ignore 'data/' --inspect=0.0.0.0:5858",
+    "start:dev": "NODE_ENV=development nodemon --ignore 'data/ migrations/' --inspect=0.0.0.0:5858",
     "lint": "eslint .",
     "test": "./scripts/test.sh",
     "test:breakingChanges": "node scripts/checkForBreakingChanges.js",
-    "dynamodb-admin": "AWS_ACCESS_KEY_ID=dummy AWS_SECRET_ACCESS_KEY=dummy DYNAMO_ENDPOINT=http://localhost:8050 dynamodb-admin"
+    "dynamodb-admin": "AWS_ACCESS_KEY_ID=dummy AWS_SECRET_ACCESS_KEY=dummy DYNAMO_ENDPOINT=http://localhost:8050 dynamodb-admin",
+    "create-migration": "node utils/migrations/createMigration.js"
```
tgandrews: Maybe this should live in scripts?

eq-author-api/README.md
```diff
@@ -9,6 +9,7 @@ application.
 - [Environment Variables](#environment-variables)
 - [Run using Docker](#run-using-docker)
 - [Tests](#tests)
+- [Migrations](#migrations)
```
tgandrews: I guess this will impact the docs from @jchapman68 so there may be some merging or a different format to follow

eq-author-api/utils/migrations/createMigration.js
adammahmood: I considered this but refrained from adding another dependancy. I'm happy to add Chalk if others are in agreement.

tgandrews: As it won't be bundled I don't think its a big deal and I personally have no idea what that escape sequence does.

eq-author-api/package.json
adammahmood: Once I move this to `/scripts`, it won't be covered due the existing ignore rule.  We don't test other scripts, why would we test this?

tgandrews: Yeah, sorry I missed that migrations end up as `/migrations` and not `utils/migrations` so I wanted to make sure those were covered.

eq-author-api/scripts/createMigration.js
```diff
@@ -0,0 +1,22 @@
+const fs = require("fs").promises;
+const chalk = require("chalk");
+
+if (process.argv.length < 3) {
+  /* eslint-disable-next-line no-console */
+  console.info(
+    chalk.green(
```
tgandrews: Maybe red instead so that its an obvious error?

eq-author-api/scripts/createMigration.js
```diff
@@ -0,0 +1,22 @@
...
+      "Not enough arguments provided to createMigration. Maybe the migration name is missing? Please refer to README."
+    )
+  );
+  return;
```
tgandrews: Can we use `process.exit(1)` so the terminal picks it up that the command failed

eq-author-api/scripts/createMigration.js
tgandrews: Also can we use `console.error`

eq-author-api/middleware/runQuestionnaireMigrations.test.js
```diff
@@ -0,0 +1,77 @@
...
+      },
+    };
+    migrations = [jest.fn()];
+    let result = await runQuestionnaireMigrations(logger)(migrations)(
+      req,
+      res,
+      next
+    );
+    expect(next).toHaveBeenCalled();
+    expect(result).toBeFalsy();
```
tgandrews: Can we assert on no migrations being run instead of result which does not require this functionality for it to be falsy?

eq-author-api/middleware/runQuestionnaireMigrations.test.js
```diff
@@ -0,0 +1,77 @@
...
+    let migrationTwo = jest.fn(() => req.questionnaire);
+    let migrationThree = jest.fn(() => req.questionnaire);
+
+    migrations = [migrationOne, migrationTwo, migrationThree];
+
+    await runQuestionnaireMigrations(logger)(migrations)(req, res, next);
+    expect(next).toHaveBeenCalled();
+    expect(migrationOne).not.toHaveBeenCalled();
+    expect(migrationTwo).toHaveBeenCalledWith(req.questionnaire);
+    expect(migrationThree).toHaveBeenCalledWith(req.questionnaire);
```
tgandrews: Could we have this prove it was the output of `migrationTwo`?

eq-author-api/utils/datastoreDynamo.js
```diff
@@ -56,7 +56,7 @@ const getQuestionnaire = id => {
   });
 };
 
-const MAX_UPDATE_TIMES = 2;
+const MAX_UPDATE_TIMES = 5;
```
tgandrews: Why is this necessary?
Does removing the equal fix solve the issue?

adammahmood: Unsure at the moment.  Having cypress issues and giving a few things a go!

tgandrews: If the isEqual is to blame. Maybe move it to a separate PR?

adammahmood: There's UI errors as well 😕 
screenshot 2019-03-08 at 15 59 20

tgandrews: Weird.

eq-author-api/middleware/runQuestionnaireMigrations.test.js
```diff
@@ -0,0 +1,82 @@
...
+    expect(req.questionnaire.title).toEqual("foo");
+    expect(req.questionnaire.description).toEqual("bar");
+    expect(logger.info).toHaveBeenCalledTimes(2);
+    expect(req.questionnaire.version).toEqual(3);
+  });
+});
```
samiwel: I wonder if it's worth adding a test where one of the migrations throws an error. Just to prove that the version does not get incremented and the saveQuestionnaire function should never be called?

eq-author-api/migrations/addVersion.js
```diff
@@ -0,0 +1,6 @@
+//This is an auto-generated file.  Do NOT modify the method signature.
+
+module.exports = questionnaire => {
+  questionnaire.version = 1;
```
samiwel: Do we want to be defensive about this? Or are we happy making the assumption that the version number is the source of truth? e.g. should this be `questionnaire.version = questionnaire.version || 1` or should we error/log a warning if it already has a version?

This is more of an open question that a request to change the above BTW.

eq-author-api/scripts/createMigration.js
```diff
@@ -0,0 +1,22 @@
...
+
+const name = process.argv[2];
+const filename = `${name}.js`;
+
+fs.readFile("scripts/migration.template.js").then(data => {
+  fs.writeFile(`migrations/${filename}`, data);
+});
+
+/* eslint-disable-next-line no-console */
+console.info(chalk.green(`${name} migration successfully created.`));
```
samiwel: Maybe add a hint here to remember to add it to the `migrations/index.js` array?

eq-author-api/migrations/addVersion.js
tgandrews: I think we have to assume each migration moves from one state to another. The state before (now) has no version so this adds it. I would be worried where the one with a version got it from.

adammahmood:
Removed the following Codacy Markdown check as it conflicts with Prettier:

>Warn when the spacing between a list item’s bullet and its content

eq-author-api/migrations/addVersion.js
tgandrews: When is it not undefined? Surely this migration would not have been run in that instance?

tgandrews:
> Removed the following Codacy Markdown check as it conflicts with Prettier:
> 
> > Warn when the spacing between a list item’s bullet and its content

prettier isn't running on markdown is it? So how do these conflict?

adammahmood:
> > Removed the following Codacy Markdown check as it conflicts with Prettier:
> > > Warn when the spacing between a list item’s bullet and its content
> 
> prettier isn't running on markdown is it? So how do these conflict?

Prettier started supporting Markdown from version 1.8:
https://prettier.io/blog/2017/11/07/1.8.0.html

As discussed we should update our root `package.json` to run on all supported files (not just `*.js`)

eq-author-api/migrations/addVersion.js
adammahmood: As discussed, we will leave this untouched and always set to version 1.

PR #256, opened 2019-03-12
Map our user type to a table

SamGodwin2:
This Pr adds the concept of a user table and is part of the inroductionary work for permissions.

### How to review 
New user table is correctly created and on first sign in user accounts are created for all users. Tests also pass.


eq-author-api/db/models/DynamoDB.js
```diff
@@ -79,6 +84,33 @@ const questionnaireVersionsSchema = new dynamoose.Schema(
...
+    sub: {
+      type: String,
+      required: true,
+    },
+    picture: {
+      type: String,
+    },
+  },
+  {
+    throughput: throughput,
```
tgandrews:
```suggestion
    throughput,
```

eq-author-api/middleware/addUserToRequest.js
```diff
@@ -0,0 +1,13 @@
+const { getUserBySub } = require("../utils/datastore");
+
+module.exports = async (req, res, next) => {
+  let user = await getUserBySub(req.auth.sub);
+
+  if (!user) {
+    res.status(401).send("User does not exist");
```
tgandrews: The rest of the middleware will keep running. So you should return here.

eq-author-api/middleware/upsertUser.js
```diff
@@ -0,0 +1,21 @@
...
+module.exports = async (req, res) => {
+  let user = await getUserBySub(req.auth.sub);
+
+  if (!user) {
+    user = await createUser(req.auth);
+  } else if (!checkForUpdates(user, req.auth)) {
+    Object.assign(user, req.auth);
+    user = await updateUser(user);
+  }
+  res.send(200);
```
tgandrews: You should send a json response here:
`res.json({ status: "OK" });`
It by default will have a status of 200.

eq-author-api/middleware/upsertUser.js
```diff
@@ -0,0 +1,21 @@
...
+  return user;
```
tgandrews: I don't think we should be returning this.
```suggestion
  return;
```

eq-author-api/middleware/upsertUser.test.js
```diff
@@ -0,0 +1,29 @@
...
+  });
+  it("should create a user is one doesn't exist", async () => {
+    const user = await upsertUser(req, res);
+    expect(user).toMatchObject({
+      sub: "new_user",
+      id: expect.any(String),
+    });
+    expect(res.send).toHaveBeenCalledWith(200);
+  });
+  it("should update the user if any changes", async () => {
```
tgandrews: We are also missing a test for when the user is already created and not changed.

eq-author-api/schema/resolvers/base.js
```diff
@@ -219,17 +219,17 @@ const Resolvers = {
       return { pageId, ...confirmationPage };
     },
     me: (root, args, ctx) => ({
-      id: ctx.auth.sub,
-      ...pick(ctx.auth, ["name", "email", "picture"]),
-      name: ctx.auth.name || ctx.auth.email,
+      id: ctx.user.id,
+      ...pick(ctx.user, ["name", "email", "picture"]),
```
tgandrews: We pick name and then overwrite it?

eq-author-api/schema/resolvers/base.js
```diff
@@ -219,17 +219,17 @@ const Resolvers = {
...
+      name: ctx.user.name || ctx.user.email,
```
tgandrews: Should this be `displayName`?

eq-author-api/utils/datastoreDynamo.js
```diff
@@ -21,6 +23,41 @@ const saveModel = (model, options = {}) =>
     });
   });
 
+const createUser = user => {
+  const { id, email, name, sub, picture } = user;
+  return saveModel(
+    new UserModel({
+      id: id ? id : uuid.v4(),
```
tgandrews: Sometimes it is created with a known id?

```suggestion
      id: id || uuid.v4(),
```

eq-author-api/utils/datastoreFileSystem.js
```diff
@@ -1,5 +1,6 @@
-const { omit, reject } = require("lodash");
+const { omit, reject, merge } = require("lodash");
```
tgandrews: @samiwel are we going to keep this around or should we just drop this filesystem support. It seems an unnecessary burden to maintain.

eq-author/.env.development
```diff
@@ -1,2 +1,3 @@
 REACT_APP_API_URL="http://localhost:4000/graphql"
```
tgandrews: We can drop the server location for all of these now?

eq-author/src/App/SignInForm/index.js
```diff
@@ -43,7 +43,13 @@ SignInForm.defaultProps = {
...
     credentialHelper,
     callbacks: {
-      signInSuccessWithAuthResult: () => false, // Avoid redirects after sign-in.
+      signInSuccessWithAuthResult: authResult => {
+        return window
+          .fetch(process.env.REACT_APP_SIGN_IN_URL, {
+            headers: { authorization: `Bearer ${authResult.user.ra}` },
+          })
+          .then(() => false);
+      }, // Avoid redirects after sign-in.
```
tgandrews: Is this `false` and comment still necessary?

SamGodwin2:
[Map our User type to a table. (6)](https://trello.com/c/b5tDb8j3/976-map-our-user-type-to-a-table-6)

eq-author-api/utils/datastoreFileSystem.js
samiwel: Now that we have some kind of GUI for local dynamodb and an import script to seed the tables, maintaining a filesystem datastore is not as important.

eq-author-api/app.js
```diff
@@ -45,14 +49,17 @@ app.use(
   pino,
   cors(),
   createAuthMiddleware(logger),
-  loadQuestionnaire,
-  runQuestionnaireMigrations(logger)(require("./migrations"))
+  runQuestionnaireMigrations(logger)(require("./migrations")),
+  userSetup,
```
tgandrews: Should this also be set for launch? We could do as part of the permissions changes.

PR #257, opened 2019-03-12
Author validation proposal

tgandrews:
Description of the proposed API validation architecture discovered during the spike.

### How to review 
1. Ensure you are happy with approach.

### DIFF
--- a/docs/architecture/0003-author-validation.md
+++ b/docs/architecture/0003-author-validation.md
@@ -13,7 +13,7 @@ they are building. This change aims to provide a way for the users to see feedba
 they are making but exposing the validation errors to them.
 
 There are multiple types of validations:
-- Intra-entity. These are validations within a field itself e.g. a question page must
+- Intra-entity. These are validations on fields within an entity itself e.g. a question page must
 have a populated title.
 - Inter-entity validations. These are validations across entities. e.g. an answer validation that
 references an answer can only reference a previous answer.
--- a/docs/architecture/0003-author-validation.md
+++ b/docs/architecture/0003-author-validation.md
@@ -1,0 +1,77 @@
+# 3. Author validation
+
+Date: 2019-03-12
+
+## Status
+
+Accepted
+
+## Context
+
+Currently users are provided with very little useful feedback as to any errors in the questionnaire
+they are building. This change aims to provide a way for the users to see feedback on the changes
+they are making but exposing the validation errors to them.
+
+There are multiple types of validations:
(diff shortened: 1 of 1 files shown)


tgandrews:
[Spike: Validation for authors (2-3 days)](https://trello.com/c/kXNOMR6a/987-spike-validation-for-authors-2-3-days)

docs/architecture/0003-author-validation.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Currently users are provided with very little useful feedback as to any errors in the questionnaire
+they are building. This change aims to provide a way for the users to see feedback on the changes
+they are making but exposing the validation errors to them.
```
samiwel:
```suggestion
they are making by exposing the validation errors to them.
```

docs/architecture/0003-author-validation.md
```diff
@@ -0,0 +1,77 @@
...
+
+There are multiple types of validations:
+- Intra-entity. These are validations within a field itself e.g. a question page must
```
samiwel:
```suggestion
- Intra-entity. These are validations on fields within an entity itself e.g. a question page must
```

docs/architecture/0003-author-validation.md
```diff
@@ -0,0 +1,77 @@
...
+### 3. Re-validate the questionnaire after every mutation
+The validation errors can change afer a mutation so we need to update the errors held in the context
+once this has happened so that the result being read on the response of the mutation contain the
+correct validation errors.
+
+### 4. Provide the validation errors at the entiy level
+All entities will have a validation errors array field in their graphql type which will contain a 
+list of all the errors. Each error will contain:
+- location - The field/property name that has the error
+- errorCode - The code representing the type of error
```
samiwel: Two questions:

1. Is location alone enough context to target a single field? Just thinking of a case where, say, an answer and an option are both missing labels. Will location need to be parsed to determine the correct field on the UI to display the validation rule? Or do we need more fields in the result to give context?

2. Should we also return the underlying JSON Schema/AJV error for additional information about the validation?

tgandrews: 1. I had expected the option itself to have its own validation errors. So the location should just be the field name in all cases I would have thought.

2. My worry is that it would couple the UI to API implementation so if we chose for some reason to use a new method for validation the interface to the UI could still be maintained.

samiwel: Okay, good points 👍

PR #258, opened 2019-03-12
Ignore timestamps in patches

adammahmood:
Improvements to https://github.com/ONSdigital/eq-author-app/pull/252:
 - Reduces number of database requests by ignoring timestamps in patches
 - Sets current version when creating a new questionnaire 

### How to review 
- Ensure tests pass
- Ensure version is updated as expected

### DIFF
--- a/eq-author-api/migrations/index.js
+++ b/eq-author-api/migrations/index.js
@@ -1,3 +1,10 @@
 const addVersion = require("./addVersion");
 
-module.exports = [addVersion];
+const migrations = [addVersion];
+
+const currentVersion = migrations.length;
+
+module.exports = {
+  migrations,
+  currentVersion,
+};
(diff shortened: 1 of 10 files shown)


eq-author-api/migrations/index.js
```diff
@@ -1,3 +1,10 @@
 const addVersion = require("./addVersion");
 
-module.exports = [addVersion];
+const migrations = [addVersion];
+
+const version = migrations.length;
+
+module.exports = {
+  migrations,
+  version,
```
samiwel: Can we rename version to be clearer about what we mean e.g. `latestVersion` or `currentVersion` or similar?

eq-author-api/utils/datastoreDynamo.js
```diff
@@ -56,7 +59,7 @@ const getQuestionnaire = id => {
   });
 };
 
-const MAX_UPDATE_TIMES = 3;
+const MAX_UPDATE_TIMES = 2;
```
samiwel: Nice 👍

eq-author-api/migrations/index.js
adammahmood: I named it `version` so it was easier to add to objects but happy to rename for clarity 👍

PR #260, opened 2019-03-15
734 move existing optional fields to the properties panel

adammahmood:
- Allows you to hide and show optional fields on a question page 
- All optional fields are disabled by default
- Any populated optional fields in existing questionnaires should be shown

### How to review 
- Ensure tests pass
- Ensure migrations run correctly


adammahmood:
[Move existing optional fields to the properties panel (9)](https://trello.com/c/m7c0m75w/734-move-existing-optional-fields-to-the-properties-panel-9)

eq-author/src/App/questionPage/Design/QuestionPageEditor/withHandleTransitionEnd.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+
+const withHandleTransitionEnd = WrappedComponent => {
```
tgandrews: This feels more of a util than a HoC.

eq-author/src/App/questionPage/Design/QuestionPageEditor/withHandleTransitionEnd.js
```diff
@@ -0,0 +1,27 @@
...
+  return class extends React.Component {
+    handleTransitionEnd = selector => {
+      document.querySelector(`[id="${selector}"`).focus();
```
tgandrews: This feels like a quick fix instead of having all the field inputs work the same way. IMO it should all work like the RTE.

eq-author/src/App/questionPage/Design/QuestionPageEditor/withHandleTransitionEnd.js
adammahmood: Why a util rather than HOC? I guess it doesn't manage any state...?

eq-author/src/components/buttons/ToggleSwitch/__snapshots__/index.test.js.snap
```diff
@@ -365,7 +365,7 @@ exports[`ToggleSwitch should render when checked and mounted 1`] = `
 .c4 {
   height: 1em;
   width: 1.5em;
-  top: 1em;
+  top: 0.6000000000000001em;
```
tgandrews: WAT?

eq-author/src/App/questionPage/Design/QuestionPageEditor/withHandleTransitionEnd.js
adammahmood: I spent some time trying to get a consistent method to work for both RTE and normal fields. Feel free to have a go too! The API's are too inconsistent but I may have missed something 👍

eq-author/src/components/buttons/ToggleSwitch/__snapshots__/index.test.js.snap
adammahmood: Some styles have changed in the `ToggleSwitch` as per @hamishtaplin prototype.  This is probably calculated or something.

eq-author/src/App/questionPage/Design/QuestionPageEditor/withHandleTransitionEnd.js
tgandrews: Maybe findDOMNode is better - but is still discouraged. 
https://reactjs.org/docs/react-dom.html#finddomnode

I think our only option is to add a focus method to all of the inputs types so we can call it with the ref like RTE.

eq-author/src/App/questionPage/Design/QuestionPageEditor/withHandleTransitionEnd.js
tgandrews: Yeah, it doesn't create any other components, add lifecycle methods etc and so it's not really react specific at the moment. Other than its given a ref.

eq-author/src/App/questionPage/Design/QuestionPageEditor/withHandleTransitionEnd.js
hamishtaplin: Yeah not a fan of this either and `finddomnode` not ideal either. 

Will have a look at some options...

eq-author/src/App/questionPage/Design/QuestionPageEditor/withHandleTransitionEnd.js
tgandrews: I view HoCs as awkward ways of sharing logic and so are only really justified when they are necessary. Nothing this is doing requires it to be a HoC. If it was doing something with context etc that required it then it would seem fine otherwise I would prefer a simpler system of sharing logic - a util file.

Additionally this component exposes some very generally named functions that are true for how they are used but do not explain what they do.

For example we don't expose `onButtonClick` from `withCreatePage` we expose `onCreatePage` as that is what it is doing and how it gets called is the components responsibility.

The fact it has this smell as well makes me feel that a HoC is not the right way of sharing this logic.

eq-author/src/components/buttons/ToggleSwitch/__snapshots__/index.test.js.snap
tgandrews: What is the calculation? Is it a css `calc` or something done in JS that we could move into the code? Or maybe I've miss understood this

hamishtaplin: https://github.com/ONSdigital/eq-author-app/blob/master/eq-author/src/components/buttons/ToggleSwitch/index.js#L34

tgandrews: Ah, no its me being stupid. This is a snaphshot! I'm an idiot. The calculation is there.

eq-author/src/App/questionPage/Design/QuestionPageEditor/withHandleTransitionEnd.js
adammahmood: As discussed we will move this to a `util`.

eq-author/src/App/questionPage/Design/QuestionPageEditor/withHandleTransitionEnd.js
adammahmood: As discussed we will leave as-is and tackle this as tech debt (alongside other areas which handle `focus` in a similar way)

eq-author/cypress/integration/authenticated/validation_spec.js
```diff
@@ -162,6 +172,7 @@ describe("Answer Validation", () => {
           toggleCheckboxOff("@maxValueToggle");
           toggleCheckboxOn("@maxValueToggle");
           cy.get(numberInputId).should("have.value", "3");
+          closeValidationModal();
```
tgandrews: Can this be moved to the `afterEach` on line 169?

eq-author/cypress/integration/authenticated/validation_spec.js
```diff
@@ -504,11 +545,13 @@ describe("Answer Validation", () => {
           .type("1985-09-14")
           .blur()
           .should("have.value", "1985-09-14");
+        closeValidationModal();
       });
 
       it("should update metadata", () => {
         toggleCheckboxOn("@latestDateToggle");
         setMetadata("@latestDate", METADATA_KEY);
+        closeValidationModal();
```
tgandrews: Can this also move to an `afterEach`?

eq-author/src/App/questionPage/Design/QuestionProperties/index.test.js
```diff
@@ -0,0 +1,42 @@
...
+    };
+
+    wrapper = render(props);
+  });
+
+  it("should render", () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it.each([
```
tgandrews: Cool - I've not seen this before :+1:

eq-author/src/App/questionPage/Preview/__snapshots__/index.test.js.snap
```diff
@@ -153,3 +148,14 @@ exports[`PreviewPageRoute should render definition label missing message 1`] = `
   </Error>
 </Preview__DetailsTitle>
 `;
+
+exports[`PreviewPageRoute should render description missing message 1`] = `
```
tgandrews: This looks wrong? Does the snapshot need updating?

eq-author/src/utils/focusOnElement.js
```diff
@@ -0,0 +1 @@
+export default selector => document.querySelector(`[id="${selector}"`).focus();
```
tgandrews: Can we do `document.getElementById`? More of a nit than actual problem.

eq-author/cypress/integration/authenticated/validation_spec.js
adammahmood: Unfortunately, the `...set previous answer` test does not require `closeValidationModal()`.

PR #262, opened 2019-03-15
Fix issue where unable to add page at correct position

tgandrews:
The position provided should be used to determine where to add the page
it was being ignored and persisted to the question page incorrectly.

This change uses it to insert the page at the right position and inserts
at the end if no position is provided.

### How to review 
1. Starting from a section with one page.
2. Add a page.
3. It should be added as the first page of the section and persist in that position after refresh.

In #182 this is not true but should be after this change.

### DIFF
--- a/eq-author-api/schema/tests/page.test.js
+++ b/eq-author-api/schema/tests/page.test.js
@@ -50,6 +50,32 @@ describe("page", () => {
         position: 0,
       });
     });
+
+    it("should create at a given position", async () => {
+      questionnaire = await buildQuestionnaire({
+        sections: [
+          {
+            pages: [{}],
+          },
+        ],
+      });
+      const section = questionnaire.sections[0];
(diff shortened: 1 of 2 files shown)


tgandrews:
Travis failing because of #263

eq-author-api/schema/tests/page.test.js
```diff
@@ -50,6 +50,29 @@ describe("page", () => {
...
+        title: "Title",
+        description: "Description",
+        sectionId: section.id,
+        position: 0,
+      });
+      expect(createdPage).toMatchObject({
+        title: "Title",
+        description: "Description",
+        position: 0,
+      });
```
SamGodwin2: Instead of asserting on the object that's created, just wondering if the test could be a little more explicit and query the section and then assert that the first page is correct?

PR #264, opened 2019-03-15
Create a new version of questionnaire on each mutation

samiwel:
Calculates the difference between either:
    a) the original version, or
    b) the latest version, if a newer version exists.
Patches the latest version with diff created by current mutation.

e.g. 
v1 - originalItem
v2 - latest
v3 - questionnaire

In the above case it should diff questionnaire with latest, if a new version has been written since the original read operation.

### How to review 
Tests are green. Changes look sensible.

### DIFF
--- a/eq-author-api/utils/datastoreDynamo.js
+++ b/eq-author-api/utils/datastoreDynamo.js
@@ -109,7 +109,7 @@ const saveQuestionnaire = async (versionModel, count = 0, patch) => {
 
     const newVersion = new Date();
 
-    const result = await dynamoose.transaction([
+    await dynamoose.transaction([
       QuestionnaireVersionsModel.transaction.create({
         ...versionModel,
         updatedAt: newVersion,
--- a/eq-author-api/utils/datastoreDynamo.js
+++ b/eq-author-api/utils/datastoreDynamo.js
@@ -68,21 +68,6 @@ const getQuestionnaire = id => {
   });
 };
 
-const getLatestVersion = id => {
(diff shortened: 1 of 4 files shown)


rmccar:
[Write new version of the questionnaire on each mutation (6)](https://trello.com/c/quiNtA2y/974-write-new-version-of-the-questionnaire-on-each-mutation-6)

eq-author-api/utils/datastoreDynamo.js
```diff
@@ -59,49 +68,78 @@ const getQuestionnaire = id => {
...
-  questionnaire.updatedAt = newUpdatedAt;
   try {
-    await saveModel(questionnaire, {
-      updateTimestamps: false,
-      condition: "updatedAt = :updatedAt",
-      conditionValues: {
-        updatedAt: oldUpdatedAt,
-      },
-    });
+    const questionnaireModel = await getLatestVersion(versionModel.id);
```
tgandrews: This seems unnecessary. Can't we use `originalItem`?

eq-author-api/utils/datastoreDynamo.js
```diff
@@ -59,49 +68,78 @@ const getQuestionnaire = id => {
...
+    const originalLatestVersion = questionnaireModel.latestVersion;
+
+    const newVersion = new Date();
+
+    await dynamoose.transaction([
```
tgandrews: Nice one. I didn't know about this api :+1:

PR #270, opened 2019-03-21
Make the versions a shadow table

tgandrews:
Change the way we use the versions table so we the latest version of any questionnaire on the questionnaires table and all versions of a questionnaire on the versions table.

This means we read and save all data from the questionnaire table (no title bug) and we no longer have to perform a scan to get the latest questionnaire. We can just look it up by id.

This change makes adding questionnaire type and short name easier as 

### How to review 
1. Ensure that the questionnaires are being saved as expected - latest in the questionnaires table and all versions (including latest) in the versions table.

### DIFF
--- a/eq-author-api/db/models/DynamoDB.js
+++ b/eq-author-api/db/models/DynamoDB.js
@@ -33,57 +33,58 @@ const baseQuestionnaireSchema = {
   version: {
     type: Number,
   },
+  description: {
+    type: String,
+  },
+  legalBasis: {
+    type: String,
+  },
+  navigation: {
+    type: Boolean,
+  },
+  surveyId: {
+    type: String,
+  },
+  theme: {
+    type: String,
+  },
+  summary: {
+    type: Boolean,
+  },
+  sections: {
+    type: Array,
+    required: true,
+  },
+  metadata: {
+    type: Array,
+    required: true,
+  },
+  createdAt: {
+    type: Date,
+    required: true,
+  },
 };
 
 const questionnanaireSchema = new dynamoose.Schema(
   {
     ...baseQuestionnaireSchema,
-    latestVersion: {
+    updatedAt: {
       type: Date,
       required: true,
     },
   },
   {
     throughput: throughput,
-    timestamps: true,
   }
 );
 
 const questionnaireVersionsSchema = new dynamoose.Schema(
   {
     ...baseQuestionnaireSchema,
-    description: {
-      type: String,
-    },
-    legalBasis: {
-      type: String,
-    },
-    navigation: {
-      type: Boolean,
-    },
-    surveyId: {
-      type: String,
-    },
-    theme: {
-      type: String,
-    },
-    summary: {
-      type: Boolean,
-    },
-    createdAt: {
-      type: Date,
-    },
     updatedAt: {
       type: Date,
       required: true,
       rangeKey: true,
-    },
-    sections: {
-      type: Array,
-      required: true,
-    },
-    metadata: {
-      type: Array,
     },
   },
   {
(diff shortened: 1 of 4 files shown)


eq-author-api/db/models/DynamoDB.js
```diff
@@ -33,58 +33,59 @@ const baseQuestionnaireSchema = {
...
+    type: Array,
+    required: true,
+  },
+  metadata: {
+    type: Array,
+    required: true,
+  },
+  createdAt: {
+    type: Date,
+    required: true,
```
samiwel: We can probably go back to using Dynamoose `timestamps:true`

eq-author-api/utils/datastoreDynamo.js
```diff
@@ -80,40 +50,39 @@ const saveQuestionnaire = async (versionModel, count = 0, patch) => {
...
     omitTimestamps(versionModel)
   );
-
   if (!diff) {
     return versionModel;
   }
 
   try {
-    const originalLatestVersion = originalQuestionnaireVersion.updatedAt;
+    const originalUpdatedAt = originalQuestionnaireVersion.updatedAt;
```
samiwel: Is this now `originalQuestionnaire` rather than the version?

tgandrews:
builder_spec passes but looks like there are some unit test failures. I'm going to have a look now.

tgandrews:
@samiwel and @adammahmood I had to deal with some issues where the tests weren't working due to dynamoose not updating the originalItem which is fine for the app but caused an issue with our tests so there is a change in `executeQuery` to deal with this. Please re-review.

eq-author-api/tests/utils/executeQuery.js
```diff
@@ -18,6 +18,18 @@ Resulted in:
 ${response.errors.map(e => e.message).join("\n----\n")}
     `);
   }
+  // Required because dynamoose transaction does not update the questionnaireModel
+  // originalItem is out of sync
+  if (
+    ctx.questionnaire &&
+    ctx.questionnaire.originalItem &&
```
SamGodwin2: This questionnaire is the dynamoose model right? When would it not have an originalItem property?

tgandrews: yes, good point. I'll remove this

tgandrews:
```suggestion
```

tgandrews: Ah, I just checked and it has no original item when a questionnaire is duplicated

eq-author-api/db/models/DynamoDB.js
tgandrews: I just tried and it seems to clash with our values causing it to merge. Maybe we can look at doing it later. Unless you think its worth digging into more?

eq-author-api/middleware/loadQuestionnaire.test.js
```diff
@@ -33,6 +33,6 @@ describe("loadQuestionnaire", () => {
     await new Promise(resolve => {
       loadQuestionnaire(req, res, resolve);
     });
-    expect(req.questionnaire).toEqual(null);
+    expect(req.questionnaire).toEqual(undefined);
```
samiwel: This could become `toBeUndefined` but not a big deal.

eq-author-api/utils/datastoreDynamo.js
```diff
@@ -26,110 +26,87 @@ const saveModel = (model, options = {}) =>
...
 
     const patchToApply = patch || diff;
     logger.warn(
-      `Dynamoose merging on save id: ${versionModel.id}`,
+      `Dynamoose merging on save id: ${questionnaireModel.id}`,
       patchToApply
     );
 
-    const dbQuestionnaire = await getQuestionnaire(versionModel.id);
+    const dbQuestionnaire = await getQuestionnaire(questionnaireModel.id);
```
samiwel: Can we perhaps rename `dbQuestionnaire` to something like `latestVersion` or `refreshedLatestVersion`? I think it would help to make it clearer what's going on here.

PR #271, opened 2019-03-21
Make RTE toolbar permanently visible

adammahmood:
- Makes RTE toolbar permanently visible
- Prior to this change the toolbar would become visible when focusing on the field

### How to review 
- Ensure tests pass
- Ensure all RTE toolbars are permanently visible (not just on focus)


adammahmood:
[Make editing toolbar permanently visible](https://trello.com/c/o6CU7LpI/949-make-editing-toolbar-permanently-visible)

adammahmood:
The build is failing on the RTE storybook.  Do we intend to maintain this storybook?

adammahmood:
@hamishtaplin @tgandrews Shall we delete this storybook entry? I can't identify a quick fix and just wondering if it's worth spending time to resolve it. 

Personally, I think we should get rid of storybook entirely as we don't maintain it and it seems like more of a chore than useful in the current stage of our product.

tgandrews:
We have a tech session. Maybe we can discuss removing storybook entirely
which can be done as a PR which would unblock this one?

hamishtaplin:
@adammahmood agree it should be removed now, just do whatever needed to get this PR to pass and we can remove it all later

with regards the formatting toolbar you can see I [made some changes to Toolbar](src/components/RichTextEditor/Toolbar.js), adding a `canFocus` prop passed to the `ToolbarButton` which [prevents it being focussed](src/components/RichTextEditor/ToolbarButton.js) until the Toolbar is visible.

adammahmood:
Storybook for RTE will be removed so we can continue with this PR.   We will remove Storybook entirely via https://github.com/ONSdigital/eq-author-app/issues/277

eq-author/src/components/RichTextEditor/Toolbar.js
```diff
@@ -123,20 +123,19 @@ class ToolBar extends React.Component {
...
-                {formattingButtons.map(this.renderButton)}
-                <Separator />
-                <PipingMenu
-                  disabled={isPipingDisabled}
-                  onItemChosen={onPiping}
-                />
-              </ButtonGroup>
-            </FadeTransition>
-          )}
+          <FadeTransition>
```
tgandrews: Is this transition needed?

PR #278, opened 2019-03-22
Fix piping not decorating currency answers.

samiwel:
There was a defect where the £ symbol was not being added to piped currency answers.
I tracked down the cause being that the type was not being added to the piping data attributes.

The type is required at present and used by publisher in order to determine how to correctly render the piped output in runner.
A change was made to shapeTree to ensure that the type property is present on answers when selecting values to pipe.

### How to review 
- Checkout master and ensure that when you pipe in a currency answer the £ symbol is not displayed, only the value.
- Checkout this branch and ensure that when you pipe in a currency answer then the answer is rendered correctly when previewed in survey runner.
- Observe that `data-type={{answer.type}}` is present in the HTML that gets persisted in question titles, descriptions etc. when piping in answer values.
- Check that other answer types e.g. Date, Date Range etc. exhibit similar behaviour.

### DIFF
--- a/eq-author/src/components/ContentPicker/shapeTree.js
+++ b/eq-author/src/components/ContentPicker/shapeTree.js
@@ -4,7 +4,7 @@ const shapeTree = answers =>
   reduce(
     answers,
     (result, value) => {
-      const answer = pick(value, ["id", "displayName"]);
+      const answer = pick(value, ["id", "displayName", "type"]);
       const page = pick(value.page, ["id", "displayName"]);
       const section = pick(value.page.section, ["id", "displayName"]);
 
--- a/eq-author/src/components/ContentPicker/shapeTree.test.js
+++ b/eq-author/src/components/ContentPicker/shapeTree.test.js
@@ -4,6 +4,7 @@ const answerTree = [
   {
     id: "6",
     displayName: "Date 1",
+    type: "Date",
     page: {
       id: "1",
       displayName: "Page (1.1)",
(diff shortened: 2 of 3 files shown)


eq-author/cypress/integration/e2e/e2e_piping_spec.js
```diff
@@ -0,0 +1,69 @@
+import { question, questionnaire, section } from "../../builders";
+import { get } from "lodash";
+describe("End to end piping", () => {
```
tgandrews: Could we add some piping to the existing e2e and update the snapshot?

tgandrews: Also we have a piping_spec - could it go there?

samiwel: Not quite as simple as updating and comparing the snapshot since the test pipes into the title and since the answerId being piped will be different between test runs, the deep equals comparison will fail unless we also ignore titles (which defeats the object of the test). It could live in the existing piping spec I guess, the only reason I separated it out was to distinguish between tests that only hit the API and tests that assert against publisher output.

tgandrews: Thats a shame. I guess we would need a custom matcher to ignore the id in the piping but care about the rest of it.

Ok, for the moment can we just move it to piping as I'm afraid it would get lost here. I don't think I'm too bothered about whether we should make a great distinction based on the assertion. 

I kinda viewed the e2e spec as one test that would grow to encompass a large amount of author features to build a complex survey. But that hasn't really happened.

PR #279, opened 2019-03-22
Improvements to modal

tgandrews:
Adds type and short title to the questionnaire modal and few small UI tweaks as per prototype.

- The type is used to determine the theme in publisher.
- The short title works like alias field for other entities and so questionnaire has a displayName property that checks short title first and then the title.

At the moment the type field means the theme field in author does nothing but I have left this as I presume we will use it to override the theme later.

### How to review 
1. Ensure you can create a questionnaire with a type
2. Ensure that short title works to override title and used in author wherever questionnaire name is displayed.


tgandrews:
[Improvements to modal (8)](https://trello.com/c/3vCHiVdg/726-improvements-to-modal-8)

tgandrews:
[Improvements to modal (8)](https://trello.com/c/3vCHiVdg/726-improvements-to-modal-8)

PR #281, opened 2019-03-25
Bug - Moving answers causes UI error

### LINKED ISSUE #274: Navigating whilst moving answers causes react errors
1. Create a page with a few answers
2. Move an answer up or down
3. Navigate to another page away during the animation 
4. Errors appear in the console

### Technical info
This occurs because the timeout is not being cleared on the unmounting the component so the state is being set on an unmounted component.

tgandrews:
This fixes the issue described in #274 

It also adds so cypress will error if a console.error is produced during the test.

Closes #274 

### How to review 
1. Prove that you see a console error when the `builder_spec` when moving answers
1. Prove that the test fails when you have the first commit
1. Prove that the test passes when it is fixed in the second commit

### DIFF
--- a/eq-author/cypress/support/errors.js
+++ b/eq-author/cypress/support/errors.js
@@ -1,0 +1,7 @@
+/* eslint-disable import/unambiguous */
+Cypress.on(`window:before:load`, win => {
+  cy.stub(win.console, `error`, msg => {
+    cy.now(`task`, `error`, msg);
+    throw new Error(msg);
+  });
+});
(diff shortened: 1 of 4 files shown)


eq-author/cypress/support/errors.js
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable import/unambiguous */
+Cypress.on(`window:before:load`, win => {
```
adammahmood: Nice 👍

PR #284, opened 2019-03-25
Add test ensuring that duplication works

tgandrews:
This aims to ensure we have permanently resolved the issue of answers appearing in the wrong order.

### How to review (Sorry its a bit awkward)
1. Check out `bug-answer-duplication-recreation` which is against postgres. See this test fail.
1. Check out this branch and see it passes

### DIFF
--- a/eq-author/cypress/integration/authenticated/duplicate_spec.js
+++ b/eq-author/cypress/integration/authenticated/duplicate_spec.js
@@ -1,4 +1,6 @@
 /* eslint-disable camelcase */
+import omitDeep from "omit-deep-lodash";
+
 import {
   addQuestionnaire,
   testId,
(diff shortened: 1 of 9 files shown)


tgandrews:
[Start and End period questions presenting the wrong way around](https://trello.com/c/fgzrjVld/1000-start-and-end-period-questions-presenting-the-wrong-way-around)

eq-author/cypress/integration/authenticated/duplicate_spec.js
```diff
@@ -75,6 +99,32 @@ describe("Duplicate", () => {
...
+          return cy.getPublisherOutput(id).then(dupe => {
+            return cy
+              .getPublisherOutput(originalQuestionnaireId)
+              .then(original => {
+                return [original.body, dupe.body];
+              });
+          });
+        })
+        .then(stuff => {
+          const [originalOutput, dupeOutput] = stuff;
```
adammahmood: Can we rename `stuff` to something more meaningful?

PR #285, opened 2019-03-26
Bug - Errors thrown in routing

tgandrews:
Fix issues when creating

When running the `routing_spec` locally it would frequently fail as cypress now throws errors when it sees a `console.error`.

Issues:
1. API was incorrectly creating a right side for a new binary expression when it should be null. The right side it was creating wasn't valid for all scenarios either - it was fine for Radio but not basic answers.
2. Input was switching from uncontrolled to controlled causing React to log an error which caused Cypress to blow up. This was caused by the right side value defaulting to undefined not found. Have change it to null which resolves this.

### How to review 
1. Ensure that routing_spec fails for you locally ~~(seems to work in Travis 😫)~~ because they are only thrown in dev mode
2. On this branch it should no longer fail

### DIFF
--- a/eq-author-api/package.json
+++ b/eq-author-api/package.json
@@ -52,6 +52,7 @@
       "!config/**/*",
       "!coverage/**/*",
       "!tests/**/*",
+      "!**/tests/**/*",
       "!scripts/*",
       "!middleware/import.js",
       "!app.js",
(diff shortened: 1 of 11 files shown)


tgandrews:
[Navigation sidebar flash](https://trello.com/c/J8n5v5uS/1008-navigation-sidebar-flash)

PR #286, opened 2019-03-26
Moved page to own type as prep work for calculated summary page

SamGodwin2:
Removed the moveQuestionPage, deleteQuestionPage and queryQuestionPage queries and mutations and instead made them more generic page operations in order to be able to share them between page types.
Also cleaned up what is required in a questionPage and what can be placed on the page interface.

### How to review 
Changes make sense and all functionality and tests remain unchanged.


eq-author-api/schema/typeDefs.js
```diff
@@ -451,10 +454,10 @@ type Mutation {
   createPage(input: CreatePageInput!): Page
```
tgandrews: I think we can get rid of this?

SamGodwin2: Yup missed this one.

Issue #276, opened 2019-03-22
Visual bug ghosting on properties panel


samiwel:
I've noticed a weird visual bug after moving answers, deleting an answer and then deleting a page.

### Expected behaviour
User does not see the ghost remnants of the answers

### Actual behaviour
Ghost image of deleted answers are visible to the user.
A refresh is required to get rid of the ghost image.

### Steps to reproduce the behaviour
Create a questionnaire.
Add three answers (tested with three currency answers)
~Move one of the answers down/up~
Delete the middle answer
Delete the page.

### Technical information
Tested on chrome in staging environment.

#### Browser
Chrome  72.0.3626.121

#### Operating System
Mac OS X Mojave

### Screenshot
![image](https://user-images.githubusercontent.com/4097796/54820108-f07be400-4c95-11e9-8406-78d3dd1c0d1b.png)

tgandrews:
You do not need to move an answer to recreate. You just need to delete the middle answer.

But it looks like a Chrome bug

samiwel:
Yeah I thought as much, we've seen similar weird ghosting behaviour in the past. IIRC adding a `postion: relative` fixed it in other places.

pricea1:
Fixed by https://github.com/ONSdigital/eq-author-app/pull/287

PR #287, opened 2019-03-27
Add accordion to properties panel

adammahmood:
- Adds a new accordion component
- Moves question properties and answer properties into accordions

### How to review 
- Ensure tests pass
- Ensure the properties panel interactions and design matches https://github.com/ONSdigital/eq-author-app/tree/prototype-multiple-answer-validation


adammahmood:
[Implement accordion for properties panel (2)](https://trello.com/c/RgLzhpyZ/992-implement-accordion-for-properties-panel-2)

eq-author/cypress/integration/authenticated/properties_spec.js
```diff
@@ -63,49 +63,35 @@ describe("Answer Properties", () => {
...
+      answerTypeTitles.forEach(({ title, type }) => {
         addAnswerType(type);
-        cy.get(`[data-test="properties-title-${index}"]`).contains(title);
+        cy.get(`[data-test="accordion-${kebabCase(title)}-button"]`).contains(
+          title
+        );
       });
 
       removeAnswer({ multiple: true });
+      removeAnswer({ multiple: true }); //Remove any answers not in viewport during first removeAnswer
```
tgandrews: This must have been causing those flaky properties specs. Nice find! :+1:

eq-author/src/App/questionPage/Design/PropertiesPanel/index.js
```diff
@@ -76,32 +75,30 @@ class PropertiesPanel extends React.Component {
...
-                    </PropertiesPanelTitle>
-                    <Properties
-                      id={getIdForObject(answer)}
-                      answer={{ ...answer, index }}
-                      onSubmit={this.handleSubmit}
-                    />
-                    <AnswerValidation answer={answer} />
-                  </AnswerPropertiesContainer>
+                    <AnswerPropertiesContainer>
+                      <Properties
```
SamGodwin2: Same issue as above.

eq-author/src/App/questionPage/Design/PropertiesPanel/index.js
```diff
@@ -76,32 +75,30 @@ class PropertiesPanel extends React.Component {
         <PropertiesPaneBody>
           <ScrollPane>
             {!isEmpty(page) && (
-              <div>
-                <QuestionProperties page={page} />
-              </div>
+              <Accordion title="Optional fields">
+                <Padding>
+                  <QuestionProperties page={page} />
```
SamGodwin2: Looks like you can still tab through the contents of this accordion panel even while it's hidden.

eq-author/cypress/integration/authenticated/properties_spec.js
```diff
@@ -33,9 +33,9 @@ describe("Answer Properties", () => {
     it("Should show answer type as uppercase text", () => {
       [TEXTFIELD, NUMBER, CURRENCY, TEXTAREA].forEach(answerType => {
         addAnswerType(answerType);
-        cy.get(`[data-test="properties-title-0"]`)
-          .should("have.length", 1)
-          .contains(toUpper(answerType));
+        cy.get(
+          `[data-test="accordion-${answerType.toLowerCase()}-button"]`
```
samiwel: I think we have a `testId` helper that could be used here.

eq-author/cypress/integration/authenticated/properties_spec.js
samiwel: Why does removeAnswer not find the answers outside of the viewport? Does the selector not match them?

adammahmood:
Also fixes https://github.com/ONSdigital/eq-author-app/issues/276

adammahmood:
@samiwel Cypress `click()` only interacts with elements which are in the viewport. We could pass `{force: true}` to be certain all elements are clicked.

eq-author/cypress/integration/authenticated/properties_spec.js
tgandrews: I double checked the [docs](https://docs.cypress.io/api/commands/click.html#Click-all-buttons-found-on-the-page) and the way I read it was that it would call click on all of them in a loop. And click scrolls them into view. So this should already work.

tgandrews:
@adammahmood @samiwel 
I double checked the [docs](https://docs.cypress.io/api/commands/click.html#Click-all-buttons-found-on-the-page) and the way I read it was that it would call click on all of them in a loop. And click scrolls them into view. So this should already work.

tgandrews:
Looks like the properties spec is still failing. So I've created #294

PR #288, opened 2019-03-27
Bug - flashing display name

### LINKED ISSUE #275: Navigation sidebar "flash" when toggling optional fields
### Expected behaviour

No visual side-effects from toggling optional fields

### Actual behaviour

When toggling optional fields, the title of the current page will "flash" in the navigation sidebar

1. On a question page
1. Toggle an optional field
1. Title of that field should "flash" in the navigation sidebar

tgandrews:
Fixes #275 

This was caused by the optimistic update setting the displayName to an empty string. To resolve this we need access to the unfiltered entity which was not provided by `withEntityEditor`. So `withEntityEditor` has been changed to no longer filter the entity and to allow the mutators (e.g. `withUpdatePage`) to filter down the entity as they choose.

### How to review 
1. On master, update a field in page. See that the display name in the side bar flashes
2. See that this doesn't happen in this branch.

### DIFF
--- a/eq-author/src/App/MetadataModal/MetadataTable/Row.js
+++ b/eq-author/src/App/MetadataModal/MetadataTable/Row.js
@@ -1,6 +1,5 @@
 import React, { Component } from "react";
 import CustomPropTypes from "custom-prop-types";
-import metadataFragment from "graphql/fragments/metadata.graphql";
 import PropTypes from "prop-types";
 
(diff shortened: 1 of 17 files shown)


tgandrews:
[Navigation sidebar flash](https://trello.com/c/J8n5v5uS/1008-navigation-sidebar-flash)

PR #289, opened 2019-03-28
Add default metadata for business surveys

adammahmood:
- Adds default metadata to business surveys

### How to review 
- Ensure tests pass
- Ensure default metadata specified in https://trello.com/c/9KPOTWcN/928-default-metadata-2 is set when creating a business survey
- Ensure there isn't any metadata set when creating a social survey

### DIFF
--- a/eq-author-api/utils/defaultMetadata.js
+++ b/eq-author-api/utils/defaultMetadata.js
@@ -1,3 +1,7 @@
+const uuid = require("uuid");
+const { includes } = require("lodash");
+const { filter, flow, map } = require("lodash/fp");
+
 const defaultTypeValueNames = {
   Date: "dateValue",
   Text: "textValue",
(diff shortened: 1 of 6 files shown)


adammahmood:
[Default Metadata (2)](https://trello.com/c/9KPOTWcN/928-default-metadata-2)

eq-author-api/utils/defaultMetadata.js
```diff
@@ -93,8 +95,54 @@ const defaultValues = [
   },
 ];
 
+const defaultBusinessSurveyMetadata = [
```
tgandrews: Could this be a subset of the above list? Instead of maintaining two default values etc. IIRC during the planning the values (the only difference) were just examples and did not have to be followed.

adammahmood: Good idea 👍

eq-author-api/utils/defaultMetadata.js
```diff
@@ -23,13 +27,13 @@ const defaultValues = [
     key: "ru_name",
     alias: "Ru Name",
     type: "Text",
-    value: "ESSENTIAL ENTERPRISE LTD.",
+    value: "APPLE COMPUTER LIMITED",
```
tgandrews: Are we sure we want these to be the default values? If there is no confirmation to change them can we leave them as before.

adammahmood: So leave the `defaultValues` as they were but change the `defaultBusinessValues` to use existing `defaultValues`?

PR #292, opened 2019-03-29
Bug - Not publishing question description

tgandrews:
Read whether description is enabled otherwise it will never be published

I'm not sure where to put the test for this. Maybe in the e2e?

### How to review 
1. On master create a questionnaire containing a question with a description.
2. See that the description is not in the published output even when enabled in the app.
3. Check out this branch and see that it is now there.


tgandrews:
[Migrating Stocks surveys (1 day)](https://trello.com/c/M7Kym3ZS/1006-migrating-stocks-surveys-1-day)

adammahmood:
LGTM 👍 I think the e2e would be the best place to test this.

PR #296, opened 2019-03-29
Add google tag manager and analytics

rmccar:
This is to add the code needed for google analytics with google tag manager to monitor the author app

### How to review 
- Tests pass
- Google Analytics account updates when using the app

### DIFF
--- a/eq-author/public/index.html
+++ b/eq-author/public/index.html
@@ -1,20 +1,41 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
-    <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta http-equiv="cache-control" content="max-age=0" />
     <meta http-equiv="cache-control" content="no-cache" />
     <meta http-equiv="expires" content="0" />
     <meta http-equiv="expires" content="Tue, 01 Jan 1980 1:00:00 GMT" />
     <meta http-equiv="pragma" content="no-cache" />
-    <link rel="apple-touch-icon" sizes="180x180" href="%PUBLIC_URL%/apple-touch-icon.png">
-    <link rel="icon" type="image/png" sizes="32x32" href="%PUBLIC_URL%/favicon-32x32.png">
-    <link rel="icon" type="image/png" sizes="16x16" href="%PUBLIC_URL%/favicon-16x16.png">
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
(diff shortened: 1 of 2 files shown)


eq-author/public/index.html
```diff
@@ -1,20 +1,55 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
-    <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <!-- Google Tag Manager -->
```
tgandrews: Can we move this with the rest of the tracking code. They are still within the head.

eq-author/public/index.html
```diff
@@ -27,54 +62,100 @@
...
-        })(window, document, window['_fs_namespace'], 'script', 'user');
+        })(window, document, window["_fs_namespace"], "script", "user");
       }
     </script>
   </head>
   <body>
+    <!-- Google Tag Manager (noscript) -->
+    <noscript
+      ><iframe
+        src="https://www.googletagmanager.com/ns.html?id=GTM-5STKLX7"
```
tgandrews: This `id` looks specific to the app. So I wonder if it should come from an environment variable. If so it would have to follow the pattern of the other variables e.g. `REACT_APP_HOT_JAR_ID`

eq-author/public/index.html
```diff
@@ -1,20 +1,55 @@
...
+      (function(w, d, s, l, i) {
+        w[l] = w[l] || [];
+        w[l].push({ "gtm.start": new Date().getTime(), event: "gtm.js" });
+        var f = d.getElementsByTagName(s)[0],
+          j = d.createElement(s),
+          dl = l != "dataLayer" ? "&l=" + l : "";
+        j.async = true;
+        j.src = "https://www.googletagmanager.com/gtm.js?id=" + i + dl;
+        f.parentNode.insertBefore(j, f);
+      })(window, document, "script", "dataLayer", "GTM-5STKLX7");
```
tgandrews: Yeah, this id needs to be an environment variable

eq-author/public/index.html
rmccar: Yeah thats fine, the docs just said to put it as far up in the head as possible

eq-author/public/index.html
rmccar: It is specific, its the id of the google tag manager container

samiwel: Now that this has appeared in the source code, we should probably look to rotate it with a new Id if possible.

rmccar: Ive recreated the container now so it has a new id. Just need to set up the environment variable in concourse and then I can give it a test again

rmccar:
[Turn on tag manager/GA for Author (1)](https://trello.com/c/iBjAeUAh/1013-turn-on-tag-manager-ga-for-author-1)

PR #299, opened 2019-04-02
Added error boundary to handle 404 errors

pricea1:
Display Page Not Found page when navigating to non existant questionnaire in URL. URL is not to be changed.

### How to review 
1. Navigate to non existant questionnaire e.g. /#/q/non-existant
2. Page not found will display. nb: in Dev, the triggered JS error will display after a few seconds. The error is dismissable and only displayed in dev - this is a by design in React. see https://github.com/facebook/create-react-app/issues/3627
3. Ensure unit test passes
4. Ensure new Error_spec integration test passes.

### DIFF
--- a/eq-author/cypress/integration/authenticated/errors_spec.js
+++ b/eq-author/cypress/integration/authenticated/errors_spec.js
@@ -1,0 +1,23 @@
+import { testId } from "../../utils";
+describe("errors", () => {
+  beforeEach(() => {
+    cy.visit("/");
+    cy.login();
+  });
+
+  it("displays page not found page if navigating to non-existant questionnaire", done => {
+    cy.on("uncaught:exception", () => {
+      // return false to prevent the error from
+      // failing this test
+      return false;
+    });
+
+    cy.visit("/#/q/not-existant-id");
+
+    cy.get(testId("not-found-page-title"))
+      .should("have.length", 1)
+      .then(() => {
+        done();
+      });
+  });
+});
--- a/eq-author/src/App/ErrorBoundary/index.js
+++ b/eq-author/src/App/ErrorBoundary/index.js
@@ -1,0 +1,25 @@
+import React from "react";
+import PropTypes from "prop-types";
+import NotFound from "App/NotFoundPage";
+import { ERR_PAGE_NOT_FOUND } from "constants/error-codes";
+
+export default class ErrorBoundary extends React.Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired
+  };
+
+  state = { error: false };
+
+  componentDidCatch(error) {
+    this.setState({ error });
+  }
+
+  render() {
+    if (this.state.error.message === ERR_PAGE_NOT_FOUND) {
+      // You can render any custom fallback UI
+      return <NotFound />;
+    }
+
+    return this.props.children;
+  }
+}
--- a/eq-author/src/App/ErrorBoundary/index.test.js
+++ b/eq-author/src/App/ErrorBoundary/index.test.js
@@ -1,0 +1,30 @@
+import React from "react";
+import { shallow } from "enzyme";
+
+import ErrorBoundary from "./";
+import { ERR_PAGE_NOT_FOUND } from "constants/error-codes";
+import NotFound from "App/NotFoundPage";
+
+const MockComponent = () => null;
+
+describe("ErrorBoundary", () => {
+  let wrapper;
+
+  beforeEach(() => {
+    wrapper = shallow(
+      <ErrorBoundary>
+        <MockComponent />
+      </ErrorBoundary>
+    );
+  });
+
+  it("should render the child component by default", () => {
+    expect(wrapper.find(MockComponent)).toHaveLength(1);
+  });
+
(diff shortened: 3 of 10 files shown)


pricea1:
[Error on questionnaire doesn't exist (2)](https://trello.com/c/jBtoLGw7/999-error-on-questionnaire-doesnt-exist-2)

eq-author/src/App/ErrorBoundary/index.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default class ErrorBoundary extends React.Component {
+  static propTypes = {
+    children: PropTypes.object.isRequired, // eslint-disable-line
+  };
+
+  state = { error: false };
+
+  componentDidCatch(error, errorInfo) {
+    // Catch errors in any components below and re-render with error message
```
tgandrews: I don't think these comments are needed

eq-author/src/App/ErrorBoundary/index.test.js
```diff
@@ -0,0 +1,30 @@
...
+  beforeEach(() => {
+    wrapper = shallow(
+      <ErrorBoundary>
+        <MockComponent />
+      </ErrorBoundary>
+    );
+  });
+
+  it("should render the child component by default", () => {
+    expect(wrapper.instance().render()).toMatchObject(<MockComponent />);
```
tgandrews: Could this be something like this?
`expect(wrapper.find(MockComponent)).toHaveLength(1)`

eq-author/src/App/QuestionnaireDesignPage/index.js
```diff
@@ -215,8 +217,15 @@ export default withMutations(props => (
       },
     }}
   >
-    {innerProps => (
-      <UnwrappedQuestionnaireDesignPage {...innerProps} {...props} />
-    )}
+    {innerProps => {
+      if (!innerProps.data || !innerProps.data.questionnaire) {
```
tgandrews: There are some code paths that check for `loading` in `render` that I don't think will ever run now. As we will never have data whilst we are loading.

I presume we still want to show loading messages

eq-author/src/App/ErrorBoundary/index.js
pricea1: now removed

eq-author/src/App/ErrorBoundary/index.test.js
pricea1: yes. code updated

eq-author/src/App/QuestionnaireDesignPage/index.js
pricea1: refactored so <UnwrappedQuestionnaireDesignPage /> always returned unless component not loading when no questionnaire data.

eq-author/cypress/integration/authenticated/errors_spec.js
```diff
@@ -0,0 +1,23 @@
...
+  });
+
+  it("displays page not found page if navigating to non-existant questionnaire", done => {
+    cy.on("uncaught:exception", () => {
+      // return false to prevent the error from
+      // failing this test
+      return false;
+    });
+
+    cy.visit("http://localhost:3000/#/q/not-existant-id");
```
tgandrews: Ah this domain driven by an environment variable which is defined here https://github.com/ONSdigital/eq-author-app/blob/e85f6f9d2ef9b9e86e59bcebee4722aabeffd7f2/eq-author/.env.test#L1

However you would probably want to read it with https://docs.cypress.io/guides/references/configuration.html#Cypress-config

eq-author/src/App/ErrorBoundary/index.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import PropTypes from "prop-types";
+import NotFound from "App/NotFoundPage";
+import { ERR_PAGE_NOT_FOUND } from "constants/error-codes";
+
+export default class ErrorBoundary extends React.Component {
+  static propTypes = {
+    children: PropTypes.object.isRequired, // eslint-disable-line
```
SamGodwin2: Can we use `Proptypes.node` here, save having to disable the line?

eq-author/cypress/integration/authenticated/errors_spec.js
pricea1: Now changed to relative URL.

eq-author/src/App/ErrorBoundary/index.js
pricea1: Done

eq-author/src/App/ErrorBoundary/index.js
```diff
@@ -0,0 +1,25 @@
...
+
+  render() {
+    if (this.state.error.message === ERR_PAGE_NOT_FOUND) {
+      // You can render any custom fallback UI
+      return <NotFound />;
+    }
+
+    return this.props.children;
+  }
+}
```
samiwel: As discussed, just an issue here about resetting the state.

pricea1: Now resetting state on url pathname change and sign out

PR #302, opened 2019-04-05
Reduce what we return in a scan

tgandrews:
The scan was very slow as it was returning all the data in author. So
now instead, we store just the data needed for the listing page in the
list table which will still scan.

Then when we are reading we read all the data from the versions table
and ensure we are reading the latest version by the sort/range key on
the field updatedAt.

This takes a scan of ~160 questionnaires from >2 seconds to 170ms.

### How to review 
1. Ensure all tests pass
2. Ensure that it works locally
3. Ensure that the list table is faster after importing a large number of questionnaires.

### DIFF
--- a/eq-author-api/db/models/DynamoDB.js
+++ b/eq-author-api/db/models/DynamoDB.js
@@ -1,4 +1,5 @@
 const dynamoose = require("dynamoose");
+const { pick } = require("lodash/fp");
 
 let throughput = "ON_DEMAND";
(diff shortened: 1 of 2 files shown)


tgandrews:
Coverage gone down slightly. We need to discuss how we can test this - especially the error and merging scenarios. Currently our black box method does not allow for that easily.

tgandrews:
@samiwel please confirm you are happy to merge with coverage drop as discussed.

PR #305, opened 2019-04-09
Fix for tag manager

rmccar:
Tag manger wasn't working correctly when the user isnt signed into google. There is a bug in the code that needed fixing so that the document was passed in when creating the body element for google tag manager

### How to review 
- Tests pass
- Log into author in a incognito tab and see in the network tab in the dev tools that the request to GTM returns a 200 status


rmccar:
[Turn on tag manager/GA for Author (1)](https://trello.com/c/iBjAeUAh/1013-turn-on-tag-manager-ga-for-author-1)

PR #308, opened 2019-04-11
Added migration to bring new metadata values into new resolver format

SamGodwin2:
After moving to the new db structure it was noticed that a number of surveys would no longer launch as the metadata was not in the correct form for the new structure.
This pr changes the resolvers to be in line with that new structure and also adds a migration to bring new metadata values into line.

### How to review 
Tests pass and all functionality remains the same after migration.

Also if you import the old pre-prod questionnaires they should now publish correctly.


eq-author-api/migrations/updateMetadataValue.test.js
```diff
@@ -0,0 +1,73 @@
...
+        id: "297",
+        key: "language",
+        languageValue: "cy",
+        regionValue: null,
+        textValue: null,
+        type: "Language",
+        value: "en",
+      },
+      {
+        alias: "language",
```
samiwel:
```suggestion
        alias: "region",
```

eq-author-api/migrations/updateMetadataValue.test.js
```diff
@@ -0,0 +1,73 @@
...
+        dateValue: null,
+        id: "297",
+        key: "language",
```
samiwel:
```suggestion
        key: "region",
```

eq-author-api/migrations/updateMetadataValue.test.js
```diff
@@ -0,0 +1,73 @@
...
+        languageValue: "cy",
+        regionValue: null,
+        textValue: null,
+        type: "Region",
```
samiwel: Are you testing a specific scenario in this last assertion that I'm missing? It feels like this in last entry is invalid configuration. In theory you should never be able to specify a `Region` metadata with a non-null `languageValue`?

eq-author-api/schema/resolvers/base.js
```diff
@@ -960,14 +959,11 @@ const Resolvers = {
   },
 
   Metadata: {
-    textValue: ({ type, value }) => (type === "Text" ? value : null),
-    languageValue: ({ type, value }) => (type === "Language" ? value : null),
-    regionValue: ({ type, value }) => (type === "Region" ? value : null),
-    dateValue: ({ type, value }) => {
-      if (type !== "Date" || !value) {
+    dateValue: ({ type, dateValue }) => {
```
samiwel: Nice tidy up 👍

eq-author-api/schema/tests/metadata.test.js
```diff
@@ -191,27 +191,35 @@ describe("metadata", () => {
...
             alias,
             type,
-            dateValue: type === DATE ? value : null,
-            regionValue: type === REGION ? value : null,
-            languageValue: type === LANGUAGE ? value : null,
-            textValue: type === TEXT ? value : null,
+            dateValue: type === DATE ? dateValue : null,
+            regionValue: type === REGION ? regionValue : null,
+            languageValue: type === LANGUAGE ? languageValue : null,
+            textValue: type === TEXT ? textValue : null,
```
samiwel: You can DRY this up slightly.

```js
const expected = {
  id: metadata.id,
  key,
  alias,
  type,
  dateValue: type === DATE ? dateValue : null,
  regionValue: type === REGION ? regionValue : null,
  languageValue: type === LANGUAGE ? languageValue : null,
  textValue: type === TEXT ? textValue : null,
};

updatedMetadata = await updateMetadata(questionnaire, expected);

expect(updatedMetadata).toMatchObject(expected);
```

eq-author-api/migrations/updateMetadataValue.js
```diff
@@ -0,0 +1,18 @@
+//This is an auto-generated file.  Do NOT modify the method signature.
+
+module.exports = function updateMetadataValue(questionnaire) {
+  const typeMatrix = {
+    Text: "textValue",
+    Date: "dateValue",
+    Region: "regionValue",
+    Language: "languageValue",
+  };
```
samiwel: This is already defined in `defaultTypeValueNames` in `defaultMetadata.js`

eq-author-api/migrations/updateMetadataValue.js
```diff
@@ -0,0 +1,18 @@
...
+  questionnaire.metadata.forEach(metadata => {
```
tgandrews: This probably be `map`. forEach works but is more awkward.

eq-author-api/migrations/updateMetadataValue.test.js
```diff
@@ -0,0 +1,73 @@
...
+    ];
+    const updatedQuestionnaire = updateMetadataValue({ metadata });
+    updatedQuestionnaire.metadata.forEach(metadata => {
+      expect(metadata).not.toHaveProperty("value");
+    });
+    expect(updatedQuestionnaire.metadata[0].textValue).toEqual("new Value");
+    expect(updatedQuestionnaire.metadata[1].dateValue).toEqual("2019-04-09");
+    expect(updatedQuestionnaire.metadata[2].regionValue).toEqual("GB_ENG");
+    expect(updatedQuestionnaire.metadata[3].languageValue).toEqual("en");
+    expect(updatedQuestionnaire.metadata[4].languageValue).toEqual(
```
tgandrews: Shouldn't this be null given that the type is different. The metadata is completely wrong really as regionValue is also null

SamGodwin2: Should it? Think I might be missing something here, this final expect tests that if there is no value field in the db the metadata is unchanged?

eq-author-api/migrations/updateMetadataValue.js
SamGodwin2: Is map really better than `.forEach` when we're not returning anything and just wanna execute a cb on all elements?

eq-author-api/migrations/updateMetadataValue.test.js
samiwel: @SamGodwin2 I think it might be worth making that explicit by having a separate test. I was confused by that last expect too.

e.g. `it("should not change existing values if no value field")`

SamGodwin2: Ah yeah I see it now the metadata is completely wrong, should be a `Language` type not `Region` I swap that and move it into a seperate test for clarity.

eq-author-api/migrations/updateMetadataValue.js
tgandrews: You are modifying each of the metadata - so I would say map is what you are doing. The only difference is you filter some out to not make changes to.

eq-author-api/migrations/updateMetadataValue.test.js
SamGodwin2: This metadata was intended to be language rather than region, I just forgot to change the type, I've changed it now though to bring it in line with the alias and key.

PR #312, opened 2019-04-12
Add function to get datatype if it is missing from entity

heofs:
No comma is showing up when piping a currency answer into a question for some of the older questionnaires. This is because the \<span\> entity is missing the "data-type" property.

Added a function that will check if the data-type is missing from \<span\>. If data-type is missing, the function will get the data-type from the questionnaireJson and insert it into the variable.

Added a test for when a \<span\> is missing its data-type property and the data-type can be accessed through the questionnaireJson. Code should return a span with the correct data-type inserted.

### How to review 
1. Create a new questionnaire with a currency answer piped into another question. 
2. Open the dynamodb-admin and remove the "data-piped" property from the \<span\> entity. 
3. View the questionnaire and see that the currency answer is correctly piped in as a "Currency" data type.


tgandrews:
Could this be done as a migration instead? So when the old questionnaires are loaded it will be resolved. Then this complexity would only live there.

SamGodwin2:
Yeah think this would be better off living as a migration, might also be worth us validating these spans as well somewhere, as silent failures like this can be quite hard to spot.

samiwel:
Yes, it could be done as a migration and that was my original thought also. But after thinking about it some more, I think that migration would be trickier to write IMO. Since you would essentially need to parse strings (everywhere where we currently allow piping), dealing with the edge cases, e.g. piping multiple values in the same text input etc.

One thing the migration approach has in its favour is that it maintains data integrity at the API. The publisher fix masks the underlying problem.

tgandrews:
The migration would only have to deal with only the piping locations we deal with now rather than add complexity especially or have to deal with later in another migration if/when we change how piping works.

My preference would be to eat the complexity now and resolve this inconsistency now.

I can't imagine its going to be too bad - just need a list of fields on entities and cheerio? Also we can unit test it and run it against the old questionnaire so it seems relatively risk free.

samiwel:
Okay, @heofs are you happy to take a look at writing a migration function for this instead? See https://github.com/ONSdigital/eq-author-app/blob/master/eq-author-api/README.md#migrations

PR #313, opened 2019-04-12
Added date range secondaryLabel into piping

pricea1:
Updated DATE RANGE answer id on pubisher

Secondary field of date range was not appearing in list of piped fields

### How to review 
1. Create a question with a date range answer
2. Create a second question and pipe previous date range answers into title
3. Test output on survey runner

### DIFF
--- a/eq-author/src/components/RichTextEditor/PipingMenu.js
+++ b/eq-author/src/components/RichTextEditor/PipingMenu.js
@@ -141,6 +141,36 @@ const calculateEntityName = ({
   }
 };
 
+const postProcessPipingContent = entity => {
+  if (!entity) {
+    return false;
+  }
+
+  const processedAnswers = [];
+
+  entity.availablePipingAnswers.forEach(answer => {
+    if (answer.type === "DateRange") {
+      processedAnswers.push({
+        ...answer,
+        id: `${answer.id}from`,
+      });
+
+      processedAnswers.push({
+        ...answer,
+        id: `${answer.id}to`,
+        displayName: answer.secondaryLabel || answer.secondaryLabelDefault,
+      });
+    } else {
+      processedAnswers.push(answer);
+    }
+  });
+
+  return {
+    ...entity,
+    availablePipingAnswers: processedAnswers,
+  };
+};
+
 export const UnwrappedPipingMenu = props => {
   if (!props.canFocus) {
     return <MenuButton {...buttonProps} disabled />;
@@ -156,7 +186,8 @@ export const UnwrappedPipingMenu = props => {
     >
       {({ data = {}, ...innerProps }) => {
         const entityName = calculateEntityName(props.match.params);
-        const entity = data[entityName] || {};
+        const entity = postProcessPipingContent(data[entityName]) || {};
+
         return (
           <Menu
             answerData={shapeTree(entity.availablePipingAnswers)}
--- a/eq-author/src/components/RichTextEditor/index.js
+++ b/eq-author/src/components/RichTextEditor/index.js
@@ -21,7 +21,16 @@ import createFormatStripper from "./utils/createFormatStripper";
 
 import cheerio from "cheerio";
 
-import { flow, uniq, map, keyBy, mapValues, isNull, trim } from "lodash/fp";
+import {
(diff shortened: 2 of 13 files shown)


eq-author/src/components/RichTextEditor/PipingMenu.js
```diff
@@ -126,6 +126,44 @@ const calculateEntityName = ({ pageId, confirmationId }) => {
...
+
+    } else {
+      processedAnswers.push(answer);
+    }
+   
+  });
+
+  return {
+    ...entity,
+    availablePipingAnswers: processedAnswers
```
SamGodwin2: Just wondering about having this logic live in the front end as it feels a little bit like it should be a concern of the api to return the availablePipingAnswers correctly rather than forcing the f/e to apply additional logic to the query result?

tgandrews: I was thinking the same thing. The only weird thing is that `DateRange` is an answer type and not a composite answer as it was before.

If this is only returning answers then it would seem odd if it would expand just DateRange out. 

So I see we have two options if we do want to push it the API:
1. Bring back composite answer and have this return the 2 Date answers from it.
2. Accept the complexity of having DateRange be special

SamGodwin2: Ideally I think it would be nice to keep that composite answer as it seems to be closer to the actual model of a DateRange answer. However for now maybe it's worth just doing option 2, use the logic that's already inside this Pr and just push it into the API?

tgandrews: Yeah, discussed with @samiwel and @pricea1 as we only have one instance of this complex answer we will leave it as is for the moment.

pricea1: After further investigation, adding to eq-author-api proved to create more issues. Changing the ID at this level lost the answers connection to the page and section, which used the answer ID as the connection. After discussion with @samiwel and @tgandrews it was decided to use the first approach and keep in frontend and tackle the issue as a bigger task at a later date.

pricea1:
[Piping Date Range into a Question](https://trello.com/c/9IzC7BOh/1020-piping-date-range-into-a-question)

eq-author/src/components/RichTextEditor/index.js
```diff
@@ -268,8 +270,27 @@ class RichTextEditor extends React.Component {
     if (answerPipes.length === 0) {
       return;
     }
+    
+    const processAnswerType = answers => {
+      return answers.map( answer => {
+        if (get("entity.data.type")(answer) === "DateRange") {
```
SamGodwin2:
```suggestion
        if (get("entity.data.type", answer) === "DateRange") {
```

eq-author/src/components/RichTextEditor/index.js
```diff
@@ -268,8 +270,27 @@ class RichTextEditor extends React.Component {
...
+          return {
+            ...answer,
+            entity: {
+              data: {
+                id: get("entity.data.id")(answer).replace(/(to|from)$/, "")
```
SamGodwin2:
```suggestion
                id: get("entity.data.id", answer).replace(/(to|from)$/, "")
```

eq-author/src/components/RichTextEditor/index.js
```diff
@@ -296,8 +317,34 @@ class RichTextEditor extends React.Component {
...
+
+      answers.forEach( answer => {
+        
+        if (answer.type === "DateRange") {
+          processedEntries.push({
+            ...answer,
+            id: `${answer.id}from`
+          });
+
+          processedEntries.push({
```
SamGodwin2: I think push can take n arguments so I believe you can just call it once here with both new objects.

pricea1:
I've fixed the bug of listing the an empty To label or removing the label of the To field after it has been piped. I've had to introduce a new field on the Answer data type in order to have access to the default value. Simply setting the secondaryLabel to the default value if empty meant that 'Untitled Answer' text appeared in Label field. Not sure if there's another approach which doesn't introduce a new field.

PR #319, opened 2019-04-17
Fixed piping formatting issue

heofs:
This is an alternative to <https://github.com/ONSdigital/eq-author-app/pull/312>.

Added a function for retrieving the data type from the answer when publisher processes the questionnaire rather than having the data type stored in \<span\> in the database. This should fix the problem with wrong formatting of answers in publisher. And, there is no longer a need to have the data type stored in both the answer and the \<span\> element. 

This is also done to be consistent as it is the same way that metadata data-type is piped.
```js
retrieve: ({ id }, ctx) => getMetadata(ctx, id.toString()),
```

From
```js
retrieve: element => element,
```
To
```js
retrieve: ({ id }, ctx) => getAnswer(ctx, id.toString()),
```

Tests have been corrected to no longer expect a data-type attribute in the \<span\> element. The tests in `convertPipes.test.js` will test to see if the correct formatting is still being applied.

#### Migration
Added a new migration to remove all data-type fields from questionnaires.

### How to review 
1. Create a new questionnaire and pipe in a value which require formatting, for example currency or number. You can also add a description and a guidance with piped values.
2. Publish survey and see that the correct formatting is applied to the piped value.
3. Check the dynamodb-admin and confirm that the new questionnaire does **not** have a data-type in the \<Span\> element.

### DIFF
--- a/eq-author-api/migrations/dropDatatypeFieldFromPipedValues.js
+++ b/eq-author-api/migrations/dropDatatypeFieldFromPipedValues.js
@@ -1,0 +1,41 @@
+const { flatMap, get, set } = require("lodash");
+
+module.exports = function dropDatatypeFieldFromPipedValues(questionnaire) {
+  const regex = /\s?data-type="[^"]*"/g;
+
+  const sections = flatMap(questionnaire.sections, section => section);
+  const pages = flatMap(questionnaire.sections, section =>
+    flatMap(section.pages, page => page)
+  );
+
+  const PIPEABLE_PAGE_FIELDS = [
+    "title",
+    "totalTitle",
+    "description",
+    "guidance",
+    "additionalInfoContent",
+    "confirmation.title",
+  ];
+  const PIPEABLE_SECTION_FIELDS = ["introductionTitle", "introductionContent"];
+
+  const removePipe = (ctx, prop) => {
+    const value = get(ctx, prop);
+    if (typeof value === "string") {
+      set(ctx, prop, value.replace(regex, ""));
+    }
+  };
+
+  pages.forEach(page => {
+    PIPEABLE_PAGE_FIELDS.forEach(prop => {
+      removePipe(page, prop);
+    });
+  });
+
+  sections.forEach(section => {
+    PIPEABLE_SECTION_FIELDS.forEach(prop => {
+      removePipe(section, prop);
+    });
+  });
+
+  return questionnaire;
+};
--- a/eq-author-api/migrations/dropDatatypeFieldFromPipedValues.test.js
+++ b/eq-author-api/migrations/dropDatatypeFieldFromPipedValues.test.js
@@ -1,0 +1,101 @@
+const dropDatatypeFieldFromPipedValues = require("./dropDatatypeFieldFromPipedValues.js");
+
+describe("dropDatatypeFieldFromPipedValues", () => {
+  let questionnaire;
+  beforeEach(() => {
+    questionnaire = {
+      sections: [
+        {
+          randomNum: 100,
+          introductionTitle:
+            '<p>Intro content <span data-piped="answers" data-id="a06d223f-667a-45d4-85ce-22a9d8ff07d8" data-type="Currency">[Revenue 2018]</span></p>',
+          introductionContent:
+            '<p>Section 2 intro <span data-piped="answers"data-id="a06d223f-667a-45d4-85ce-22a9d8ff07d8"data-type="Currency">[Revenue 2018]</span></p>',
+          pages: [
+            {
+              title:
+                '<p>How much of <span data-piped="answers"data-id="b4a52321-b404-473f-a459-c50b57cf1588"data-type="Currency">[Revenue 2018]</span> was sales?</p>',
+            },
+          ],
+        },
+        {
+          pages: [
+            {
+              randomBool: true,
+              title:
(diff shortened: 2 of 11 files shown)


eq-author-api/migrations/dropDatatypeFieldFromPipedValues.js
```diff
@@ -0,0 +1,38 @@
+//This is an auto-generated file.  Do NOT modify the method signature.
+const cheerio = require("cheerio");
+const { flatMap } = require("lodash");
+
+module.exports = function dropDatatypeFieldFromPipedValues(questionnaire) {
+  const pages = flatMap(questionnaire.sections, section =>
+    flatMap(section.pages, page => page)
+  );
+
+  pages.map(page => {
```
samiwel: It is also possible to pipe values into section introduction pages. We would need to also deal with updating these values.

eq-author-api/migrations/dropDatatypeFieldFromPipedValues.js
```diff
@@ -0,0 +1,38 @@
...
+    ["title", "description", "guidance"].map(prop => {
```
samiwel: I would extract this into a `const`. e.g. `PIPEABLE_PAGE_FIELDS`

eq-author-api/migrations/dropDatatypeFieldFromPipedValues.js
```diff
@@ -0,0 +1,38 @@
...
+      let $elem;
+      if (prop === "title" && page.title) {
+        $elem = cheerio.load(page.title);
+      } else if (prop === "description" && page.description) {
+        $elem = cheerio.load(page.description);
+      } else if (prop === "guidance" && page.guidance) {
+        $elem = cheerio.load(page.guidance);
```
samiwel: You can get rid of this `if` statement with something like `$elem = cherio.load(page[prop]);`

heofs: The script breaks if it tries to load an element that is empty.
Could I use something like this?
```js
      let $elem;
      if (page[prop]) {
        $elem = cheerio.load(page[prop]);
      } else {
        return;
      }
```

eq-author-api/migrations/dropDatatypeFieldFromPipedValues.js
heofs: Sorry, I was not aware of that. It should be fixed now.

eq-author-api/migrations/dropDatatypeFieldFromPipedValues.test.js
```diff
@@ -0,0 +1,217 @@
...
+                  previousAnswer: null,
+                },
+              },
+            },
+          ],
+        },
+      ],
+    },
+  ],
+};
```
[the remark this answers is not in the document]
heofs: This should have been removed in <https://github.com/ONSdigital/eq-author-app/pull/319/commits/5db2c5e9588726607f73867669a2e9c97d6d18f9>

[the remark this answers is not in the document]
heofs: I have fixed these two now.

eq-author-api/migrations/dropDatatypeFieldFromPipedValues.test.js
```diff
@@ -0,0 +1,91 @@
...
+
+describe("dropDatatypeFieldFromPipedValues", () => {
+  let questionnaire;
+  beforeEach(() => {
+    questionnaire = {
+      sections: [
+        {
+          randomNum: 100,
+          introductionTitle:
+            '<p>Intro content <span data-piped="answers"data-id="a06d223f-667a-45d4-85ce-22a9d8ff07d8"data-type="Currency">[Revenue 2018]</span></p>',
```
tgandrews: There should be spaces between the attributes as that would match the data in the server.

eq-publisher/src/utils/convertPipes.test.js
```diff
@@ -41,64 +51,64 @@ describe("convertPipes", () => {
...
+        const html = createPipe({ id: "4" });
         expect(convertPipes(createContext())(html)).toEqual(
-          "{{ answers['answer123'] | format_date }}"
+          "{{ answers['answer4'] | format_date }}"
         );
       });
 
       it("should format Currency answers with `format_currency`", () => {
-        const html = createPipe({ id: "123", type: "Currency" });
+        const html = createPipe({ id: "2", type: "Currency" });
```
tgandrews: Can this `type` be removed?

eq-publisher/src/utils/convertPipes.test.js
```diff
@@ -41,64 +51,64 @@ describe("convertPipes", () => {
...
         expect(convertPipes(createContext())(html)).toEqual(
-          "{{ format_currency(answers['answer123'], 'GBP') }}"
+          "{{ format_currency(answers['answer2'], 'GBP') }}"
         );
       });
 
       it("should format Number answers with `format_number`", () => {
-        const html = createPipe({ id: "123", type: "Number" });
+        const html = createPipe({ id: "5", type: "Number" });
```
tgandrews: Can this `type` also be moved?

eq-author-api/migrations/dropDatatypeFieldFromPipedValues.js
```diff
@@ -0,0 +1,40 @@
...
+  const pages = flatMap(questionnaire.sections, section =>
+    flatMap(section.pages, page => page)
+  );
+
+  const PIPEABLE_PAGE_FIELDS = [
+    "title",
+    "description",
+    "guidance",
+    "additionalInfoContent",
+    "confirmation.title",
```
SamGodwin2: The `totalTitle` field on a calculated page can also have answers piped into it.

samiwel:
[Comma Missing from Piping](https://trello.com/c/MBmnhf5S/1022-comma-missing-from-piping)

PR #320, opened 2019-04-17
Add linting to ensure we have run prettier

tgandrews:
We have had a couple of PRs get merged which haven't had prettier run. This is to ensure that PRs are rejected if prettier isn't run.

### How to review 
1. This PR should fail in CI before the files have been changed.


tgandrews:
@pricea1 I missed that. It's failing now with the correct error.

Issue #318, opened 2019-04-17
Lots of delete confirm elements on questionnaire list page


tgandrews:
### Steps to reproduce the behaviour
1. Visit the questionnaire list page
2. Expect to see at most 1 delete confirm dialog in the DOM
3. Actually see a large number of them

### Screenshot
![Screenshot 2019-04-17 at 09 00 57](https://user-images.githubusercontent.com/180903/56271022-66396b00-60ef-11e9-962e-bbc666dc6d95.png)

pricea1:
https://trello.com/c/3VH8kS1K/1060-eq-author-app-318-lots-of-delete-confirm-elements-on-questionnaire-list-page-issue

pricea1:
https://github.com/ONSdigital/eq-author-app/pull/322

PR #322, opened 2019-04-23
Fixed multple DOM elements for dashboard delete modal and added title…

pricea1:
Multiple DOM elements on page for delete confirm modal. Also, questionnaire title was missing from delete confirm modal.

### How to review 
1. In Master, go to dashboard with multiple questionnaires listed.
2. View DOM and observe multiple empty delete confirm DOM elements attached to bottom of body
3. Click to delete a questionnaire, observer no title being displayed in modal
4. Repeat in branch and see no DOM elements attached (modal now only attached when clicking delete).
5. Click delete button and observe modal with questionnaire title


pricea1:
[eq-author-app: #318 - Lots of delete confirm elements on questionnaire list page (Issue)](https://trello.com/c/3VH8kS1K/1060-eq-author-app-318-lots-of-delete-confirm-elements-on-questionnaire-list-page-issue)

pricea1:
@tgandrews approach to make the change in Modal component would have been straightforward and involved only rendering when modal is open. However, react-modal docs advise not to conditionally render react-modal as it loses the out-transition.

I've moved the deleteConfirm modal from each <Row> component of the questionnaire list to the parent <QuestionnairesTable> component, which was also called the mutation. The Row passes the relevant questionnaire details up to the parent to display the modal and delete. This means that there is a single DeleteConfirm modal and the modal and mutation are dealt with within the same component

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/index.js
```diff
@@ -55,6 +58,21 @@ export class UnconnectedQuestionnairesTable extends React.PureComponent {
 
   state = {
     focusedId: null,
+    showDeleteQuestionnaireDialog: false,
+    deleteQuestionnaire: {},
```
tgandrews: Could we default this to `null` or `undefined` so that its obvious its not there. `{}` is truthy

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/index.test.js
```diff
@@ -157,6 +158,49 @@ describe("QuestionnairesTable", () => {
     ).toMatchObject({ autoFocus: true });
   });
 
+  it("should show delete question confirmation when modal confirm clicked", () => {
```
tgandrews: This test is fully covered in the test below. Maybe we could remove this one?

pricea1: Agreed. I thought the same but I just transposed existing tests from Row to QuestionnaireTable so wasn't sure if there was a reason for it.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/index.js
pricea1: Changed to null with other corresponding changes to cope with null value

samiwel:
@pricea1 Looks good. Can you rebase the commit before merging?

PR #323, opened 2019-04-25
Bug - Stocks - Cannot publish calculated summary

tgandrews:
Calculate summary is the first page without answers. A lot of the resolvers assumed that all pages had answers and exploded when there was a calculated summary added to the questionnaire.

To resolve this we now have a series of methods in `schema/utils` to access different parts of the questionnaire. This PR aims to switch the resolvers to use them.

### How to review 
1. Create a questionnaire with routing on radio options
2. See that it works
3. Add a calculated summary. See viewing routing and publishing blows up on master. This branch should resolve it.


eq-author-api/schema/resolvers/routing2/routingRule2/index.js
```diff
@@ -11,6 +7,11 @@ const {
   pick,
   first,
 } = require("lodash/fp");
+
+const isMutuallyExclusive = require("../../../../utils/isMutuallyExclusive");
+
+const conditions = require("../../../../constants/routingConditions");
+
```
pricea1: Would these be better as absolute paths?

tgandrews: As discussed we have no nice way of doing this in the api at the moment

tgandrews:
[Cannot publish after creating calculated summary](https://trello.com/c/BbCgzsxv/1063-cannot-publish-after-creating-calculated-summary)

PR #326, opened 2019-04-26
Disabled Add Question Page & Add Calculated Summary options on Intro page

pricea1:
Add Question Page and Add Calculated Summary options are not valid on Introduction page and produce error. They should be disabled on introduction page

### How to review 
On Master
1. Create Business survey with Introduction page
2. On Introduction page, click either Add Question Page or Add Calculated Summary - error occurs

On branch
1. Create Business survey with intro page
2. Add Question Page and Add Calculated Summary page are disabled when on Intro page
3. Add Question Page and Add Calculated Summary page are enabled and work when on any other page


pricea1:
[Error when adding calculated summary on Introduction Page](https://trello.com/c/XIotugrK/1079-error-when-adding-calculated-summary-on-introduction-page)

eq-author/src/App/QuestionnaireDesignPage/index.js
```diff
@@ -135,6 +135,22 @@ export class UnwrappedQuestionnaireDesignPage extends Component {
...
+        params: { entityName },
+      },
+      loading,
+    } = this.props;
+
+    return !loading && entityName !== "introduction";
+  }
+
+  canAddCalculatedSummaryPage() {
+    // Criteria for adding summary page same as question page at the moment
```
tgandrews: I don't think the comment is necessary. Maybe we could come up with a better method name? So its obvious they are the same logic

pricea1: I've amalgamated the 2 methods into 1

PR #327, opened 2019-04-29
Removed ability to add a confirmation question to cal sum page

SamGodwin2:
Just noticed that it was possible to add a confirmation question while on a calculated summary page. This causes the questionnaire to reach an invalid state as calculated summary page do not support the required fields for a confirmation page.

### How to review 
No longer able to add a confirmation page while on a calculated summary page and the new test confirms this.


SamGodwin2:
[Bug - Able to add confirmation question to calculated summary page.](https://trello.com/c/PU0HXcGW/1081-bug-able-to-add-confirmation-question-to-calculated-summary-page)

eq-author/src/App/QuestionnaireDesignPage/index.test.js
```diff
@@ -236,14 +236,27 @@ describe("QuestionnaireDesignPage", () => {
...
 
+    it("should disable adding question confirmation when not on a question page", () => {
+      questionnaire.sections[0].pages[0] = {
+        id: "1",
+        title: "",
+        pageType: "NotQuestionPage",
+        position: 0,
+      };
+      wrapper.setProps({ data: { questionnaire } });
+      expect(wrapper.find(NavigationSidebar).props()).toMatchObject({
```
pricea1: Can this be changed to test for specific props. [PR 326](https://github.com/ONSdigital/eq-author-app/pull/326/files#) adds other properties which would break this test. See file change [https://github.com/ONSdigital/eq-author-app/pull/326/files#diff-49e7391792e178cd6643a4c5073cc283](https://github.com/ONSdigital/eq-author-app/pull/326/files#diff-49e7391792e178cd6643a4c5073cc283) for reference

PR #329, opened 2019-05-02
Multiple Answer Validation - Condition Location

tgandrews:
This updates the condition location based on feedback from @hamishtaplin 

It also fixes an issue where the updated condition was not being shown in the dropdown after being closed and re-opened.

### How to review 
1. Ensure the condition is updated so it now appears with the title on the button


tgandrews:
[Multiple answer validation - totals (6)](https://trello.com/c/pkIcjKHs/408-multiple-answer-validation-totals-6)

PR #331, opened 2019-05-03
Bug - Flakey cypress test caused by nested modals

tgandrews:
Nested modals causing focus to not go back to the original modals button.

So this fixes an issue which was causing a flakey cypress test.

With a lot of help from @pricea1 

### Recreation steps
1. Open a validation with the ability to pick previous answers
1. Select previous answer from the pill tabs
1. Switch away from previous answer on the pill tabs (notice the focus is on the validation button)
1. Close the validation modal

Expected: The validation button gets focus
Actual: The window has focus

### Explanation
react-modal (the library we use under the hood for all our modals) has a nice feature that when you close the modal the focus goes back to the button that opened it. 

However, it runs this logic even when you unmount the modal and before it is ever shown. This in itself would not be bad but it uses a static array to store the buttons to return focus to. This means that unmounting the an existing modal (even if it is never shown) the focus returns to the button that last opened a react-modal. react-modal incorrectly equates unmounting and closing which is incorrect.

In the example of the validations - we have the validation modal and a possible modal (previous answer or metadata picker) and as these render react-modals when you switch the pill tabs the modals are unmounted and the focus management kicks in.

### The fix
This changes the way we render react-modal so that we never render a react-modal until it is opened so that it is focus management does not kick in until it closes. This however had a slight knock on effect of losing the animation out timeout so we actually keep it in the DOM long enough for the animation out to run.

### How to review 
1. Ensure you can recreate the issue on master
1. Ensure the issue is resolved in this PR

### DIFF
--- a/eq-author/src/components/modals/Modal/Modal.test.js
+++ b/eq-author/src/components/modals/Modal/Modal.test.js
@@ -1,15 +1,11 @@
 import React from "react";
 import { shallow, mount } from "enzyme";
+
 import Modal, { CloseButton } from "./";
 
-const createWrapper = (props = {}, render = shallow, children = undefined) => {
-  return render(<Modal {...props}>{children}</Modal>);
-};
-
 describe("components/Modal", () => {
+  jest.useFakeTimers();
   let props;
-  let wrapper;
-  let children;
 
   beforeEach(() => {
     props = {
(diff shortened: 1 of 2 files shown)


tgandrews:
[Questionnaire Import/Export](https://trello.com/c/I1NIkZso/1092-questionnaire-import-export)

PR #332, opened 2019-05-03
Redirect users if they dont use Chrome

heofs:
Author breaks if user opens application in Internet Explorer.

Added a script in `index.html` to redirect users to `unsupported-browser.html` if they don't use Chrome.
Also added a script in `unsupported-browser.html` to redirect users back to Author if they are using Chrome.

### How to review
1. Open Author in Internet Explorer 11 (vm can be found here https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/)
2. You should be redirected to `http://localhost:3000/unsupported-browser.html`.
3. Copy the whole url in Internet Explorer and open it in Google Chrome.
4. You should be redirected back to the author application.


eq-author/public/index.html
```diff
@@ -8,6 +8,16 @@
     <meta http-equiv="expires" content="0" />
     <meta http-equiv="expires" content="Tue, 01 Jan 1980 1:00:00 GMT" />
     <meta http-equiv="pragma" content="no-cache" />
+    <script>
+      var isChrome =
+        !!window.chrome &&
+        (!!window.chrome.webstore || !!window.chrome.runtime);
+      var isElectron = !!process.versions.electron;
```
tgandrews: this will blow up as `process` isn't a global in most browsers.
```suggestion
      var isElectron = process && process.versions && process.versions.electron;
```

eq-author/public/index.html
```diff
@@ -8,6 +8,16 @@
     <meta http-equiv="expires" content="0" />
     <meta http-equiv="expires" content="Tue, 01 Jan 1980 1:00:00 GMT" />
     <meta http-equiv="pragma" content="no-cache" />
+    <script>
+      var isChrome =
+        !!window.chrome &&
```
tgandrews: Isn't the check for `window.chrome` enough? https://stackoverflow.com/questions/4565112/javascript-how-to-find-out-if-the-user-browser-is-chrome

eq-author/public/index.html
heofs: I have added a proper function for it now.

eq-author/public/index.html
heofs: I think it is because some browsers might have the window.chrome variable.

tgandrews: Where did you get the check from? The linked stackoverflow seems to suggest `window.chrome` is enough to detect Chromium

eq-author/public/index.html
```diff
@@ -8,6 +8,21 @@
...
+        (!!window.chrome.webstore || !!window.chrome.runtime);
+      function isElectron() {
+        return (
+          typeof process !== "undefined" &&
```
tgandrews: A falsy check here is sufficient. Are we finding places where that is not sufficient?

eq-author/public/index.html
heofs: I got it from: https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser

The stackoverflow you linked suggest that 

> new Opera 18+ also outputs to true for window.chrome

But I guess it would not be a problem if the page was displayed for an Opera user. So I can remove it if you think it would be better?

eq-author/public/index.html
heofs: Only check if `process` is falsy? Which means it is being run in a browser?

eq-author/public/index.html
tgandrews: I guess at the moment we are saying its Chrome only so we should be as restrictive as possible.

eq-author/public/index.html
tgandrews: I mean just do
`process && process.versions && process.versions.electron`

No need for explicitly checking undefined.

heofs: The function breaks in IE11 when doing it that way.

tgandrews: how? Surely process is undefined so it will shortcut it?

tgandrews: I just checked. Its read as a global. You could `window.process && ...`

heofs: When I run this in IE11:
```js
if(process && process.versions && process.versions.electron) {
  console.log("true")}
else {
  console.log("false")
}
```
It displays an error with `process' is undefined`

While this one logs `false`
```js
if(typeof process !== "undefined" && process.versions && process.versions.electron !== undefined) {
  console.log("true")
} 
else {
  console.log("false")
}
```

tgandrews: Yeah I missed it would throw because the `process` variable wouldn't be defined. But as discussed this isn't needed now

tgandrews:
[Chrome restriction](https://trello.com/c/4oA1DbEX/961-chrome-restriction)

tgandrews:
@heofs please rebase and squash this

PR #334, opened 2019-05-08
Map our user type to a table.

SamGodwin2:
PR to add the concept of a user into our app.

This PR is dependent on https://github.com/ONSdigital/eq-author-terraform/pull/10 

Once this PR goes in the two new scripts `import-users-to-table.js` and `bulkMigrateDataStore.js` will both have to be run targeting all environments one by one.

### How to review the Import scripts

1. Checkout master.
2. Create a number of questionnaires with various users new and old.
3. Switch to this branch.
4. Change [this line](https://github.com/ONSdigital/eq-author-app/pull/334/files#diff-7719754e0e817167ed1622f31f045a22R7) to `let userTableName = "dev-author-users";` This is needed because the new env vars won't be picked up as the container will not be restarted.
4. Follow the instructions here: https://firebase.google.com/docs/cli/auth to export your existing users to a json file.
5. Add the default author team user to your firebase dump.
6. Run the `DYNAMO_QUESTIONNAIRE_TABLE_NAME=dev-author-questionnaires DYNAMO_QUESTIONNAIRE_VERSION_TABLE_NAME=dev-author-questionnaire-versions
DYNAMO_USER_TABLE_NAME=dev-author-users
AWS_ACCESS_KEY_ID=dummy AWS_SECRET_ACCESS_KEY=dummy DYNAMO_ENDPOINT_OVERRIDE=http://localhost:8050 AWS_REGION=eu-west-1 DATA_DIR=<path to firebase exports> node scripts/import-users-to-table.js` script with the firebase export.
7. Run the `DYNAMO_QUESTIONNAIRE_TABLE_NAME=dev-author-questionnaires DYNAMO_QUESTIONNAIRE_VERSION_TABLE_NAME=dev-author-questionnaire-versions
DYNAMO_USER_TABLE_NAME=dev-author-users AWS_ACCESS_KEY_ID=dummy AWS_SECRET_ACCESS_KEY=dummy DYNAMO_ENDPOINT_OVERRIDE=http://localhost:8050 AWS_REGION=eu-west-1 node bulkMigrateDataStore.js`script.
8. Restart the front-end to get the proxy in the webpack to work.
9. You should find all questionnaires have been updated to a user account, the user table has been correctly populated and any unclaimed users will be defaulted to 'Author Team'.

### How to review the rest.

After migrating to the new structure all questionnaires should be created and linked to a user account and in all other cases the app should work as expected.
Tests pass.


SamGodwin2:
[Map our user type to a table. (6)](https://trello.com/c/b5tDb8j3/976-map-our-user-type-to-a-table-6)

eq-author-api/migrations/updateCreatedByToUseUsers.js
```diff
@@ -0,0 +1,25 @@
+//This is an auto-generated file.  Do NOT modify the method signature.
+
+const { getUserByName, getUserByEmail } = require("../utils/datastore");
```
tgandrews: As discussed these functions could be moved here as they aren't used in the rest of the app.

eq-author-api/middleware/usertUser.test.js
```diff
@@ -0,0 +1,32 @@
...
+  it("should create a user is one doesn't exist", async () => {
+    await upsertUser(req, res);
+    const user = await getUserBySub(req.auth.sub);
+    expect(user).toMatchObject({
+      sub: "new_user",
+      id: expect.any(String),
+    });
+    expect(res.json).toHaveBeenCalledWith({ status: "OK" });
+  });
+  it("should update the user if any changes", async () => {
```
tgandrews: As discussed this only tests update because of the previous test. So this is sharing state between tests which we shouldn't be doing.

eq-author-api/scripts/test.sh
```diff
@@ -33,6 +34,18 @@ if [ ! -f "data/QuestionnaireList.json" ]; then
     echo "[]" > "data/QuestionnaireList.json";
 fi
 
+if [ ! -f "data/UserList.json" ]; then
```
tgandrews: I think this can be removed as we are not supporting the file system data store any more.

eq-author-api/scripts/test.sh
```diff
@@ -33,6 +34,18 @@ if [ ! -f "data/QuestionnaireList.json" ]; then
     echo "[]" > "data/QuestionnaireList.json";
 fi
 
+if [ ! -f "data/UserList.json" ]; then 
+    echo "creating UserList.json";
+    echo "[]" > "data/UserList.json";
+fi
+
+ echo "creating test user..."
```
tgandrews: As discussed this is more shared state between tests

eq-author-api/keys.test.yml
```diff
@@ -0,0 +1,61 @@
+keys:
```
tgandrews: As discussed these keys need to get into the production secrets

eq-author-api/utils/verifyPublisherRequest.js
```diff
@@ -0,0 +1,13 @@
+const yaml = require("js-yaml");
+const fs = require("fs");
+const jwt = require("jsonwebtoken");
+
+const keysFile = process.env.KEYS_FILE || "./keys.yml";
+
+module.exports = authHeader => {
+  const keysYaml = yaml.safeLoad(fs.readFileSync(keysFile, "utf8"));
+  const keysJson = JSON.parse(JSON.stringify(keysYaml));
+  const publisherVerificationKey = keysJson.keys.publisherVerificationKey.value;
```
tgandrews: All this work to this point in this function could be moved out of this function and made static on the file. We probably shouldn't be running all the parsing for each request.

eq-author/cypress/support/commands.js
```diff
@@ -10,8 +10,25 @@ Cypress.Commands.add("login", options => {
...
 
-  const accessToken = createAccessToken(tokenPayload);
-  window.localStorage.setItem("accessToken", accessToken);
+  cy.window()
+    .its("__auth__")
+    .then(auth => {
+      return new Promise(res => {
+        // when firebase has triggered default logout
+        auth.onAuthStateChanged(() => {
+          setImmediate(() => {
```
tgandrews: https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate
I found out this is a non-standard API and requires a so we should replace it with `setTimeout(fn, 0)`

eq-author-api/db/models/DynamoDB.js
```diff
@@ -106,6 +111,33 @@ const questionnaireVersionsSchema = new dynamoose.Schema(
...
+      hashKey: true,
+      required: true,
+    },
+    email: {
+      type: String,
+    },
+    name: {
+      type: String,
+    },
+    sub: {
```
tgandrews: As discussed I think this needs renaming

eq-author-api/schema/resolvers/base.js
```diff
@@ -501,6 +489,10 @@ const Resolvers = {
       questionnaire.shortTitle || questionnaire.title,
   },
 
+  User: {
+    displayName: user => user.name || user.email,
```
tgandrews: Is this still necessary?

SamGodwin2: Yeah, this doesn't fix the issue with not sending the name on login, it just stops us saving the email against the questionnaire so if they login a second time it fixes itself.

eq-author-api/tests/utils/executeQuery.js
```diff
@@ -1,13 +1,12 @@
 const schema = require("../../schema");
 const { graphql } = require("graphql");
-const auth = require("./mockAuthPayload");
+
 const graphqlTools = require("graphql-tools");
-const { originalItemSymbol } = require("dynamoose/dist/Model");
+// const { originalItemSymbol } = require("dynamoose/dist/Model");
```
tgandrews: This can be removed

eq-author-api/tests/utils/executeQuery.js
```diff
@@ -19,18 +18,6 @@ Resulted in:
 ${response.errors.map(e => e.message).join("\n----\n")}
     `);
   }
-  // Required because dynamoose transaction does not update the questionnaireModel's
```
tgandrews: Woo hoo! Nice that this now goes!

eq-publisher/src/middleware/createAuthToken.js
```diff
@@ -1,15 +1,24 @@
...
+const yaml = require("js-yaml");
+const fs = require("fs");
+
+const SIGNING_ALGORITHM = "RS256";
+const keysFile = process.env.KEYS_FILE || "./keys.yml";
 
 module.exports = (req, res, next) => {
+  const keysYaml = yaml.safeLoad(fs.readFileSync(keysFile, "utf8"));
+  const keysJson = JSON.parse(JSON.stringify(keysYaml));
+  const signingKey = keysJson.keys.publisherAuthSigningKey.value;
```
tgandrews: We can move all this work of file parsing etc outside of the request.

eq-author-api/middleware/identification/index.js
```diff
@@ -0,0 +1,60 @@
...
+  const jwtToken = jwt.decode(accessToken);
+  if (isNil(jwtToken)) {
+    logger.error("Could not decode JWT token.");
+    res.send(401);
+    return;
+  }
+
+  if (jwtToken.serviceName) {
+    await verifyServiceRequest(accessToken, jwtToken.serviceName);
+    req.user = {
```
tgandrews: Should this be `req.identity`?

eq-author-api/middleware/identification/index.js
```diff
@@ -0,0 +1,60 @@
...
+    return;
+  }
+
+  let user = await getUserByExternalId(jwtToken.sub);
+  if (!user) {
+    req.user = {
+      name: jwtToken.name,
+      externalId: jwtToken.sub,
+      email: jwtToken.email,
+      userRegistered: false,
```
tgandrews: `isRegistered` I think is a better name.

eq-author-api/middleware/identification/index.js
```diff
@@ -0,0 +1,60 @@
...
+    logger.error("Could not decode JWT token.");
+    res.send(401);
+    return;
+  }
+
+  if (jwtToken.serviceName) {
+    await verifyServiceRequest(accessToken, jwtToken.serviceName);
+    req.user = {
+      name: jwtToken.serviceName,
+      serviceVerified: true,
```
tgandrews: Could use the same boolean as `userRegistered` if we come up with a better name

eq-author-api/middleware/identification/rejectUnidentifiedUsers.test.js
```diff
@@ -0,0 +1,21 @@
+const rejectUnidentifiedUsers = require("./rejectUnidentifiedUsers");
+
+describe("rejectUnidentifiedUsers", () => {
```
tgandrews: I think this should be identities as well

eq-author-api/tests/utils/contextBuilder/index.js
```diff
@@ -26,78 +25,95 @@ const { createCollapsible } = require("./collapsible");
 
 const buildRouting = require("./buildRouting");
 
+const defaultUser = require("../mockUserPayload");
+
 //@todo - Split into smaller functions to avoid deeply nested chaining
```
tgandrews: Can we do this or remove the comment

eq-author-api/middleware/identification/index.js
SamGodwin2: I had them both as the same name at first, think it was `isValidIdentity` switched it to two seperate flags though as it felt better.

eq-author-api/tests/utils/contextBuilder/index.js
SamGodwin2: Probably worth doing at some point, it is getting a little complex to figure out what's nested where. Something to look at in the near future?

eq-author-api/middleware/identification/index.js
tgandrews: It would be nice if the rest of the app didn't care where it came from. So it feels a little odd to leak the multiple types of identities.

eq-author-api/tests/utils/contextBuilder/index.js
tgandrews: Agreed, but the comment seems pointless. We should open a tech debt issue for it.

eq-author-api/middleware/identification/index.js
SamGodwin2: The app might care about where it comes from in future, while at the moment services and users are equal my preference would be to keep them separate and just be a little more verbose in the rejection middleware.

tgandrews: I'd rather it be "pure" until we get the use case for it not being so later on

eq-author-api/middleware/identification/index.js
tgandrews: As discussed we will hold off the identity change until we have a discussion about what an identity should be.

eq-author-api/middleware/identification/upsertUser.test.js
```diff
@@ -0,0 +1,36 @@
...
+  let req, res, next;
+  beforeEach(() => {
+    req = { params: { questionnaireId: 1 }, user: defaultUser() };
+    res = {
+      json: jest.fn(),
+    };
+    next = jest.fn();
+  });
+
+  it("should create a user is one doesn't exist", async () => {
```
heofs: Spelling: "if one.."

eq-author-api/middleware/identification/upsertUser.test.js
```diff
@@ -0,0 +1,36 @@
...
+    await upsertUser(req, res, next);
+    const user = await getUserByExternalId(req.user.externalId);
+    expect(user).toMatchObject(req.user);
+    expect(res.json).toHaveBeenCalledWith({ status: "OK" });
+  });
+
+  it("should update the user if any changes", async () => {
```
heofs: Changes to what?

eq-author-api/schema/tests/answerGroupValidation.test.js
```diff
@@ -81,23 +80,23 @@ describe("Answer group validation", () => {
...
         questionPageId: pageId,
         type: CURRENCY,
       });
 
-      const updatedPage = await queryPage(questionnaire, pageId);
+      const updatedPage = await queryPage(ctx, pageId);
       expect(updatedPage.totalValidation).toBeNull();
     });
 
     it("should not add it even when a second is added", async () => {
```
heofs: could this line be more explanatory?

SamGodwin2: In what way? This line change is just due to requiring the whole ctx object to run queries and mutations?

heofs: I meant the test description.

PR #338, opened 2019-05-13
Bug - Duplicated mutually exclusive options

tgandrews:
We store all options in an array on a multipleChoiceAnswer so we need
to filter them out when we return them as mutually exclusive are
returned from another field.

### How to review 
1. Add a mutually exclusive option (an Or option) to a multiple choice answer and give it a label
2. Refresh the page 
3. On master the master it is duplicated above and below the or.

### DIFF
--- a/eq-author-api/schema/tests/multipleChoiceAnswer.test.js
+++ b/eq-author-api/schema/tests/multipleChoiceAnswer.test.js
@@ -1,4 +1,4 @@
-const { last, get, flow, find } = require("lodash/fp");
+const { get, flow, find } = require("lodash/fp");
 
 const {
   buildQuestionnaire,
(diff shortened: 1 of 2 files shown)


eq-author-api/schema/tests/multipleChoiceAnswer.test.js
```diff
@@ -145,7 +145,10 @@ describe("multiple choice answer", () => {
       });
 
       it("should resolve options", () => {
-        expect(last(queriedAnswer.options).id).toEqual(last(answer.options).id);
+        expect(queriedAnswer.options).toHaveLength(2);
+        expect(
+          last(queriedAnswer.options.filter(o => !o.mutuallyExclusive)).id
+        ).toEqual(last(answer.options.filter(o => !o.mutuallyExclusive)).id);
```
samiwel: Change looks good, but can the test be re-worked to not perform the filtering in the test itself? I think it would be clearer.

PR #339, opened 2019-05-13
1076 table redesign

heofs:
Change the dashboard layout following trello card <https://trello.com/c/2CH7r9kp/1076-table-redesign-4>.  

Used https://github.com/ONSdigital/eq-author-app/tree/prototype-sharing as a guide.

Layout changes
* Added questionnaire short title to table. This should appear above the title in the row.
* Changed "Created by" column to "Owner".
* Made entire row clickable.
* Enabled hover styles on rows.
* Move questionnaire title to header
* Renamed "Create" to "Create questionnaire" and moved button to the right. 

### How to review 
* Check that the dashboard layout has changed according to the above changes.
* See that the animation when deleting questionnaires is working smoothly, even when adding several questionnaires.
* See that focus is working correctly after deleting a questionnaire.
* Check that the entire row is clickable.
* Use keyboard to navigate through layout.


eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.js
```diff
@@ -83,49 +105,87 @@ class Row extends React.Component {
...
-                data-test="btn-delete-questionnaire"
-                onClick={this.handleOpenDeleteQuestionnaireDialog}
-              />
-            </ButtonGroup>
+            <div onFocus={this.handleButtonFocus}>
+              <ButtonGroup>
+                <DuplicateButton
+                  data-test="btn-duplicate-questionnaire"
+                  onClick={this.handleDuplicateQuestionnaire}
+                  style={{ marginRight: "0.5em" }}
```
hamishtaplin: Create a styled component for this (eg. `DuplicateQuestionnaireButton`) and add the margin to that.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
```diff
@@ -60,51 +60,95 @@ describe("Row", () => {
...
+    expect(wrapper.state("linkHasFocus")).toBeTruthy();
+    tableRow.simulate("blur");
+    expect(wrapper.state("linkHasFocus")).toBeFalsy();
+  });
 
-    wrapper.find(DuplicateButton).simulate("click");
+  it("should handle button focus state change correctly", () => {
+    const wrapper = shallow(<Row questionnaire={questionnaire} {...props} />);
+    const tableRow = wrapper.find(TR);
+    const actionButton = wrapper.find(TD).find("div");
```
hamishtaplin: This is a fragile selector—the test will break if we change the `div` to a `span` or even add another `div` to the `TD`. Use a `data-test` attribute where possible.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
```diff
@@ -60,51 +60,95 @@ describe("Row", () => {
...
+    wrapper.instance().focusLink = jest.fn();
+
+    wrapper.setProps({
+      autoFocus: true,
+    });
+
+    expect(wrapper.instance().focusLink).toHaveBeenCalled();
+  });
+
+  it("should apply alignment styles to table cell", () => {
```
hamishtaplin: Not sure what this test is for? It doesn't test any feature of this component but rather the DOM.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/index.js
```diff
@@ -13,23 +13,31 @@ import Row from "App/QuestionnairesPage/QuestionnairesTable/Row";
...
   border-collapse: collapse;
   table-layout: fixed;
   text-align: left;
 `;
 
 const TH = styled.th`
-  padding: 1.5em 1em;
-  color: #8e8e8e;
+  padding: 1em;
+  color: #666;
```
hamishtaplin: Use a colour constant for this: `${colors.darkGrey}`;

eq-author/src/components/Header/index.test.js
```diff
@@ -141,7 +142,13 @@ describe("withCurrentUser", () => {
     };
 
     expect(
-      shallow(<WrappedComponent {...defaultProps} match={match} />)
+      shallow(
+        <WrappedComponent
+          {...defaultProps}
+          match={match}
+          title={"test-title"}
```
hamishtaplin: This can be just a string: `title="test-title"`. Surprised the linter doesn't pick up this.

eq-author/src/components/Header/index.test.js
```diff
@@ -150,7 +157,13 @@ describe("withCurrentUser", () => {
       path: Routes.QUESTIONNAIRE,
     };
     expect(
-      shallow(<WrappedComponent {...defaultProps} match={match} />)
+      shallow(
+        <WrappedComponent
+          {...defaultProps}
+          match={match}
+          title={"test-title"}
```
hamishtaplin: Same as above

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
heofs: Removed this test, and changed the TD to have no alignment set as the default is left.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/index.js
heofs: Done.

eq-author/src/components/Header/index.test.js
heofs: Done.

eq-author/src/components/Header/index.test.js
heofs: Done.

eq-author/src/App/QuestionnaireDesignPage/index.js
```diff
@@ -193,8 +193,9 @@ export class UnwrappedQuestionnaireDesignPage extends Component {
       throw new Error(ERR_PAGE_NOT_FOUND);
     }
 
+    const title = this.getTitle();
     return (
-      <BaseLayout questionnaire={questionnaire}>
+      <BaseLayout questionnaire={questionnaire} title={loading ? "" : title}>
```
tgandrews: I think this can become `title={this.getTitle("")}` as it is already handling the loading state there

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.js
```diff
@@ -14,65 +15,81 @@ import QuestionnaireLink from "App/QuestionnairesPage/QuestionnaireLink";
...
   display: flex;
-  justify-content: space-evenly;
 `;
 
-const TR = styled.tr`
-  border-bottom: 1px solid #e2e2e2;
+export const TR = styled.tr`
   border-top: 1px solid #e2e2e2;
+  border-bottom: 1px solid #e2e2e2;
+  background-color: ${props =>
```
tgandrews: Can we do this with an `nth-of-type` selector instead of passing it down as a prop. As its purely used for display CSS seems a more appropriate location for it.
https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-of-type

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.js
```diff
@@ -14,65 +15,81 @@ import QuestionnaireLink from "App/QuestionnairesPage/QuestionnaireLink";
...
+    props.odd ? "rgba(0,0,0,0)" : "rgba(0,0,0,0.02)"};
   opacity: 1;
   color: ${props => (props.disabled ? `${colors.textLight}` : "inherit")};
```
tgandrews: You could wrap the whole property definition then you wouldn't need the inherit.
`${props => props.disabled && `color: ${colors.textLight};`}`

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.js
```diff
@@ -14,65 +15,81 @@ import QuestionnaireLink from "App/QuestionnairesPage/QuestionnaireLink";
...
+    linkHasFocus: false,
   };
 
+  rowRef = React.createRef();
+
   handleDuplicateQuestionnaire = () => {
     this.props.onDuplicateQuestionnaire(this.props.questionnaire);
   };
 
-  shouldComponentUpdate(nextProps) {
```
tgandrews: :+1: If we have fixed the perf enough so we don't need this.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.js
```diff
@@ -83,49 +100,86 @@ class Row extends React.Component {
...
-                data-test="btn-duplicate-questionnaire"
-                onClick={this.handleDuplicateQuestionnaire}
-              />
-              <IconButtonDelete
-                hideText
-                data-test="btn-delete-questionnaire"
-                onClick={this.handleOpenDeleteQuestionnaireDialog}
-              />
-            </ButtonGroup>
+            <div onFocus={this.handleButtonFocus} data-test="action-btn">
```
tgandrews: Maybe we could call this `action-btn-group` as its not just one button

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
```diff
@@ -24,29 +38,16 @@ describe("Row", () => {
...
 
   it("should render", () => {
-    const wrapper = shallow(
-      <Row
-        questionnaire={questionnaire}
-        onDuplicateQuestionnaire={handleDuplicateQuestionnaire}
-        onDeleteQuestionnaire={handleDeleteQuestionnaire}
-      />
-    );
+    const wrapper = shallow(<Row questionnaire={questionnaire} {...props} />);
```
tgandrews: Could we move questionnaire into props, so it just becomes `<Row {...props}>`?

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
```diff
@@ -60,51 +61,109 @@ describe("Row", () => {
...
-        questionnaire={questionnaire}
-        onDeleteQuestionnaire={handleDeleteQuestionnaire}
-        onDuplicateQuestionnaire={handleDuplicateQuestionnaire}
-      />
-    );
+  it("should handle row focus state change correctly", () => {
+    const wrapper = shallow(<Row questionnaire={questionnaire} {...props} />);
+    const tableRow = wrapper.find(TR);
+
+    expect(wrapper.state("linkHasFocus")).toBeFalsy();
```
tgandrews: Can we write this test without relying on state that maybe checking for a style or something as it feels really internal and doesn't actually prove it does anything.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
```diff
@@ -60,51 +61,109 @@ describe("Row", () => {
...
+    wrapper = shallow(<Row questionnaire={questionnaire} {...props} />);
+    expect(wrapper.find(ShortTitle)).toHaveLength(0);
+
+    questionnaire.shortTitle = originalShortTitle;
+  });
+
+  it("should call focusLink in componentDidMount when autoFocus is true", () => {
+    const focusLink = jest.fn();
+    const originalFunc = Row.prototype.focusLink;
+    Row.prototype.focusLink = focusLink;
```
tgandrews: This feels pretty dirty as it knows about the private functions of the Row which is not great and will make it harder to refactor later. We should ideally test the outcomes and not care about internal implementation

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
```diff
@@ -60,51 +61,109 @@ describe("Row", () => {
...
+    shallow(<Row questionnaire={questionnaire} autoFocus {...props} />);
+    expect(focusLink).toHaveBeenCalled();
+    Row.prototype.focusLink = originalFunc;
+  });
+
+  it("should style table row correctly", () => {
+    const tableRow = shallow(<TR />);
```
tgandrews: You could pass directly instead of doing this in two steps.

tgandrews: But we can completely get rid of the need for this test if do this in pure css.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/index.js
```diff
@@ -107,30 +134,18 @@ export class UnconnectedQuestionnairesTable extends React.PureComponent {
...
-      <>
+      <Panel>
         <Table>
           <thead>
             <tr>
-              <TH colWidth="50%">Questionnaire name</TH>
-              <TH colWidth="15%">Date</TH>
-              <TH colWidth="22%">Created by</TH>
-              <TH colWidth="14%" />
+              <TH colWidth="45%">Title</TH>
```
tgandrews: These percentages don't add up to 100

eq-author/src/components/MainCanvas/index.js
```diff
@@ -1,10 +1,14 @@
...
 const MainCanvas = styled.div`
-  max-width: 55em;
+  width: ${props => props.width};
   margin: 0 auto 1em;
   padding: 0 1em;
   position: relative;
 `;
 
+MainCanvas.defaultProps = {
+  "min-width": "55em",
```
tgandrews: Default prop is `min-width`? But the read prop is `width`?

eq-author/src/App/QuestionnairesPage/index.js
```diff
@@ -75,30 +78,31 @@ export class UnconnectedQuestionnairesPage extends React.PureComponent {
...
-                onSubmit={onCreateQuestionnaire}
-                confirmText="Create"
-              />
-            </StyledButtonGroup>
-            <StyledCenteredPanel>
-              <Query query={QUESTIONNAIRES_QUERY}>{this.renderResults}</Query>
-            </StyledCenteredPanel>
-          </MainCanvas>
+          <ScrollPane permanentScrollBar>
+            <MainCanvas width="70em">
```
hamishtaplin: Remove this for now as the table doesn't have extra columns yet so doesn't need to be this wide.

eq-author/src/components/MainCanvas/index.js
hamishtaplin: This is likely copied from the prototype, probably just needs removing.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/index.js
hamishtaplin: Worth writing a test for this?

tgandrews: I'm not sure how much it matters as it only affects layout like CSS which we don't test directly. Its just odd.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.js
heofs: I just realised that the prop disabled is never passed to this component, so the line can be removed.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
heofs: How would I test that the focusLink is being called then? I am struggling with having enough test coverage to make codecov pass.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
heofs: Removed this test as I am using pure css instead.

eq-author/src/App/QuestionnairesPage/index.js
heofs: Done. Set to default of 55em.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
tgandrews: Can we stub the ref instead?

eq-author/src/App/QuestionnaireDesignPage/index.js
heofs: Problem here is, while the page is loading, the title does not exist, giving me a proptype error when opening a questionnaire, because it is not received a title.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
heofs: stub the ref?

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.js
heofs: Done.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.js
heofs: Done.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
heofs: Done.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/index.js
heofs: Changed it so it adds up to 100, however, im not sure what they should be, just picked something that looked good.

eq-author/src/components/MainCanvas/index.js
heofs: Yes, thats correct. I changed it to width.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.js
heofs: Done.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
heofs: Changed this to use data-test instead of looking for the div.

eq-author/src/App/QuestionnaireDesignPage/index.js
heofs: Changed to get the questionnaire displayName from questionnaire json directly.

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.test.js
heofs: Done.

tgandrews:
[Table redesign (4)](https://trello.com/c/2CH7r9kp/1076-table-redesign-4)

tgandrews:
[Table redesign (4)](https://trello.com/c/2CH7r9kp/1076-table-redesign-4)

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.js
```diff
@@ -10,122 +11,194 @@ import Truncated from "components/Truncated";
...
+          onFocus={this.handleFocus}
+          onBlur={this.handleBlur}
+          linkHasFocus={this.state.linkHasFocus}
+          onClick={this.handleClick}
+        >
           <TD>
-            <TruncatedQuestionnaireLink
+            <QuestionnaireLink
               data-test="anchor-questionnaire-title"
               questionnaire={questionnaire}
```
hamishtaplin: We don't need to pass this prop in now as `QuestionnaireLink` is just a styled component, and is being passed onto the DOM by that library. 

Screenshot 2019-05-17 10 02 32

eq-author/src/App/QuestionnairesPage/QuestionnairesTable/Row.js
```diff
@@ -10,122 +11,194 @@ import Truncated from "components/Truncated";
...
-              title={questionnaire.displayName}
+              title={questionnaire.title}
```
hamishtaplin: Would be good to pass both shortcode and title into this so it reads:

"SHORT CODE - Questionnaire Title"

PR #342, opened 2019-05-15
Add tests around saveQuestionnaire

tgandrews:
This adds tests around datastore dynamo saveQuestionnaire. This is an important part of the system and has some complex logic.

As part of adding tests a bug was discovered in the way we are merging arrays. This is caused by our patches relying on array index which is not deterministic when applied to a modified object. This does not have an easy solution so we need a proper ticket for investigation and resolution.

Unfortunately tools following either of [json patch](https://tools.ietf.org/html/rfc6902) or [json merge patch](https://tools.ietf.org/html/rfc7396) are insufficient as they either completely replace the array or merge using array indexes which doesn't work for us.

### How to review 
1. Ensure the tests make sense
2. Ensure the tests pass

### DIFF
--- a/eq-author-api/utils/datastoreDynamo.test.js
+++ b/eq-author-api/utils/datastoreDynamo.test.js
@@ -1,0 +1,245 @@
+const { pick, omit } = require("lodash");
+
+const originalKey = Symbol("originalItem");
+
+describe("Datastore dynamo", () => {
+  let dynamoose,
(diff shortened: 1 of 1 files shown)


tgandrews:
@pricea1 I'm not sure what to do about it - I'd rather leave it in until we solve it but can't enable it

tgandrews:
There is still an issue #345

PR #344, opened 2019-05-16
Fix flakey validation spec

tgandrews:
This attempts to resolve a flakey validation spec test where an answer was being left over after attempting to clear up between each test.

This change now waits for each answer to be deleted before trying to delete the next one

### How to review 
1. Ensure that validation_spec passes locally for you
2. Ensure after this change that travis passes consistently for this PR

### Run Count without failure 
4

### DIFF
--- a/eq-author/cypress/utils/index.js
+++ b/eq-author/cypress/utils/index.js
@@ -208,7 +208,17 @@ export const findInputByLabel = text =>
     .then($label => $label.prop("control"));
 
 export const removeAnswers = () => {
-  cy.get(testId("btn-delete-answer")).click({ multiple: true });
+  let count = 0;
+  cy.get(testId("btn-delete-answer")).each(() => {
+    count += 1;
+  });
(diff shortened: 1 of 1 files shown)


eq-author/cypress/integration/authenticated/validation_spec.js
```diff
@@ -738,7 +738,17 @@ describe("Validation", () => {
 
       afterEach(() => {
         closeValidationModal(sidebarButton);
-        removeAnswers();
+        cy.get(testId("btn-delete-answer")).should("have.length", 2);
```
SamGodwin2: Could we put this in the `removeAnswers` function as otherwise isn't there potential for this bug to appear in all of the other `afterEach`'s

tgandrews: @pricea1 can you confirm that the new change resolves @SamGodwin2's issue and dismiss his review if true

pricea1: Good for me

Issue #341, opened 2019-05-15
'See how these fields are used` wrapping issue


hamishtaplin:
### Expected behaviour

UI should not have any weird wrapping issues

### Actual behaviour

'See how these fields are used` text wraps in a weird way

### Steps to reproduce the behaviour

1. Change your browser width to be < 1000px
2. Navigate to any question page
3. Observe wrapping of button in properties panel

#### Browser

- Chrome
- browser width < 1000px

### Screenshot

<img width="1388" alt="Screenshot 2019-05-15 09 59 35" src="https://user-images.githubusercontent.com/930398/57762945-5b76f380-76f8-11e9-9d39-e99d6531b70a.png">

PR #346, opened 2019-05-16
Fixed wrapping icon text issue

SamGodwin2:
Fixes https://github.com/ONSdigital/eq-author-app/issues/341

### How to review 
Check to make sure the bug can be reproduced on master, then check it's fixed on this branch.


SamGodwin2:
['See how these fields are used` wrapping issue](https://trello.com/c/lGXzuaNe/1121-see-how-these-fields-are-used-wrapping-issue)

PR #350, opened 2019-05-20
Add opentracing instrumentation to GraphQL API

samiwel:
Added the ability to instrument and trace the GraphQL requests through the API which is useful for identifying and troubleshooting slow performing queries and mutations so we can make more informed decisions about where to make performance improvements.

Added docker-compose and application configuration necessary to instrument the Graphql API using Apollo OpenTracing library.

I have gone with a simple configuration which sees prometheus metrics collected and displayed using the Jaeger tracer library. This is only intended as a configuration to aid local development and should not be considered suitable for production use.

Added separate command for running the API with tracing enabled and added some documentation to the Readme.

### How to review 
- Ensure you have run `docker-compose down` and killed any existing API containers before switching between the two docker-compose configuration files. I have found not doing this can lead to issues starting and stopping the services.
- Run `docker-compose up` inside the API directory.
- Run the front-end as you normally would.
- Perform some actions in the UI e.g. create survey, add question etc.
- Navigate to the Jaeger UI which should be running on http://localhost:16686
- Select the eq_author_api service and view the traces.
- Should see the requests and be able to order them by the longest time and drill into them etc.

We can add additional tags to the traces over time to add additional contextual information to make the traces a bit more useful.

### DIFF
--- a/eq-author-api/README.md
+++ b/eq-author-api/README.md
@@ -84,3 +84,23 @@ It is possible to retrieve the full questionnaire data by making a `GET` request
 
 When the environment variable `ENABLE_IMPORT` is set to `true` then it exposes `/import` which will take the
 `POST` body and save it to the database. It performs no validation as it does this.
+
+## Instrumentation and tracing
+
+Instrumenting the GraphQL API is useful for troubleshooting errors and to help identify the root cause of slow running queries and mutations.
+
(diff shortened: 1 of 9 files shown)


tgandrews:
Can we merge this @samiwel ?

PR #353, opened 2019-05-22
Attempt to fix flakey calculated summary spec

tgandrews:
This was failing when going to preview and the total title was not set.
I managed to recreate this locally once and it was due to the answer
shortcut click response containing a blank title. It must have been out
of sync.

So this now waits for the total title to have finished saving by
checking the lack of saving indicator.

### Note
This spec seems to show an error whilst running as it randomly (not sure of the mechanism) loses the token mid way through a test. It is re-using the questionnaire as well and then faking login each time so I wonder if it is going against the grain with the way it is doing this. Seeding the questionnaire but using a separate one for each test seems most appropriate.

### How to review 
1. Ensure this passes for you locally.
2. Ensure this passes successfully in CI at least 10 times. (Count: 0)


tgandrews:
This doesn't work needs more investigation

PR #356, opened 2019-05-29
Adding modified column

rmccar:
This PR is to add a modified column to the questionnaire list table.

### How to review 
- Create a questionnaire and check that the modified column is present and todays date is in it

### DIFF
--- a/eq-author/src/App/QuestionnairesPage/QuestionnairesView/QuestionnairesTable/Row/__snapshots__/index.test.js.snap
+++ b/eq-author/src/App/QuestionnairesPage/QuestionnairesView/QuestionnairesTable/Row/__snapshots__/index.test.js.snap
@@ -46,6 +46,11 @@ exports[`Row should render 1`] = `
         />
       </Row__TD>
       <Row__TD>
+        <FormattedDate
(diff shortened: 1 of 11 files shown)


rmccar:
[Modified date column (1.5)](https://trello.com/c/n5tFsHz0/1073-modified-date-column-15)

eq-author/src/App/QuestionnairesPage/QuestionnairesView/QuestionnairesTable/Row/__snapshots__/index.test.js.snap
```diff
@@ -45,6 +45,9 @@ exports[`Row should render 1`] = `
           date="2017/01/02"
         />
       </Row__TD>
+      <Row__TD>
+        <FormattedDate />
```
tgandrews: Should this contain the updated date?

rmccar: Yeah you’re right good spot I've just pushed up a fix

PR #359, opened 2019-06-05
Add a script to grant firebase users admin roles.

samiwel:
There are instances where it would be useful to restrict certain features of Author to only "admins" within the team. An example of this might be restricting access import/export functionality in different environments, being able to manage user accounts, A/B testing new features before they are released to all users etc.

Whilst the permissions model is intended to address the issue of fine grained access to questionnaires, being able to give certain users admin privileges feels like an admin process that should happen outside of the application itself.

This PR introduces an admin script that can be run by a firebase service account to grant (and remove) admin status for any firebase user within the firebase project.

### How to review 
- Create a test firebase project.
- Enable authentication.
- Create a user within the firebase project.
- Create a service account for the firebase project and download the JSON key file.
- Run the script using the project name, uid and service account key.
- If successful the user should have the custom claim { admin: true } in their JWT token the next time they log in.
- Run the script again with the `-r` flag to remove the admin privilege.

### DIFF
--- a/eq-author-api/scripts/make-admin.js
+++ b/eq-author-api/scripts/make-admin.js
@@ -1,0 +1,72 @@
+#!/usr/bin/env node
+/* eslint-disable no-console */
+const admin = require("firebase-admin");
+const program = require("commander");
+const fs = require("fs").promises;
+
+program
+  .description("A utility that applies the admin role to a firebase user.")
+  .usage("-p <projectName> -u <uid> -k <serviceAccountKey> [-r]")
+  .option(
+    "-k, --service-account-key <serviceAccountKey>",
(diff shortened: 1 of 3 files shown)


eq-author-api/scripts/make-admin.js
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env node
+/* eslint-disable no-console */
+const admin = require("firebase-admin");
+const program = require("commander");
+const fs = require("fs").promises;
+
+program
+  .description("An utility that applies the admin role to a firebase user.")
```
tgandrews:
```suggestion
  .description("A utility that applies the admin role to a firebase user.")
```

PR #367, opened 2019-06-10
Spike - Fix Sentry configuration and get it to post to Slack

rmccar:
This PR is to fix the link to the Sentry app and implement a notification on error, this will be a Slack notification at the moment.

### How to review 
- Get the sentry project id from the secrets file or the sentry account
- Hardcode the id in this line `Sentry.init({ dsn: window.config.REACT_APP_SENTRY_DSN });`
- Run the Author app and create an error by changing the url to point to an invalid questionnaire id
- See the error in my the Sentry app or the Sentry test Slack channel. See me and ill invite you to it

This PR is linked to https://github.com/ONSdigital/eq-author-terraform/pull/16 and https://github.com/ONSdigital/eq-deploy/pull/73

### DIFF
--- a/eq-author/package.json
+++ b/eq-author/package.json
@@ -80,6 +80,7 @@
     "workbox-webpack-plugin": "latest"
   },
   "dependencies": {
+    "@sentry/browser": "latest",
     "apollo-cache-inmemory": "latest",
     "apollo-client": "latest",
     "apollo-link": "latest",
--- a/eq-author/src/apollo/createApolloErrorLink.js
+++ b/eq-author/src/apollo/createApolloErrorLink.js
@@ -1,13 +1,28 @@
+import {
+  setSentryUser,
+  setSentryTag,
+  sendSentryError,
+} from "../apollo/sentryUtils";
 import { onError } from "apollo-link-error";
 import { apiDownError } from "redux/saving/actions";
 import { signOutUser } from "../redux/auth/actions";
 
-export const errorHandler = (getStore, { networkError }) => {
-  if (networkError && networkError.bodyText === "User does not exist") {
-    getStore().dispatch(signOutUser());
-  } else if (networkError) {
+export const errorHandler = (getStore, error) => {
+  const { networkError, graphQLErrors } = error;
+  setSentryUser(window.localStorage.getItem("accessToken"));
+  if (networkError) {
     getStore().dispatch(apiDownError());
+    setSentryTag("networkError");
+    sendSentryError(networkError);
+    if (networkError.bodyText === "User does not exist") {
+      getStore().dispatch(signOutUser());
+    }
+    return;
   }
+  graphQLErrors.forEach(error => {
+    setSentryTag("graphQLError");
+    sendSentryError(error);
+  });
 };
 
 export default getStore => onError(errors => errorHandler(getStore, errors));
--- a/eq-author/src/apollo/createApolloErrorLink.test.js
+++ b/eq-author/src/apollo/createApolloErrorLink.test.js
@@ -1,27 +1,84 @@
+import uuid from "uuid";
+import jwt from "jsonwebtoken";
 import { errorHandler } from "./createApolloErrorLink";
+import { sendSentryError, setSentryTag, setSentryUser } from "./sentryUtils";
 
-const dispatch = jest.fn();
-
-const createMockStore = () => ({
-  dispatch,
-});
+jest.mock("./sentryUtils");
 
 describe("createErrorLink", () => {
+  let dispatch;
+  let networkError;
+  let graphQLErrors;
+
+  const createAccessToken = (token, signingKey = uuid.v4()) => {
+    return jwt.sign(token, signingKey);
+  };
+
+  const createMockStore = () => ({
+    dispatch,
+  });
+
+  beforeAll(() => {
+    dispatch = jest.fn();
+  });
+
+  beforeEach(() => {
+    /* eslint-disable-next-line camelcase */
+    const tokenData = { user_id: "Test User", email: "test.user@email.com" };
+    const token = createAccessToken(tokenData);
+    localStorage.setItem("accessToken", token);
+
+    networkError = {
+      fail: "Uh oh",
+    };
+    graphQLErrors = [
+      {
+        fail: "Uh oh",
+      },
+    ];
+  });
+
   it("should dispatch a logout event when the server responds with a 401 error", () => {
     errorHandler(createMockStore, {
-      networkError: {
-        bodyText: "User does not exist",
-      },
+      networkError: { bodyText: "User does not exist" },
     });
     expect(dispatch).toHaveBeenCalledWith(expect.any(Function));
   });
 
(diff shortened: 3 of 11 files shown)


rmccar:
[slack updates on error messages (1-2 days)](https://trello.com/c/pGkHnpq8/1129-slack-updates-on-error-messages-1-2-days)

eq-author/package.json
```diff
@@ -98,6 +98,7 @@
     "workbox-webpack-plugin": "latest"
   },
   "dependencies": {
+    "@sentry/browser": "^5.4.0",
```
tgandrews: This should be `latest`

eq-author/src/App/index.js
```diff
@@ -20,6 +20,10 @@ import QuestionnaireDesignPage from "./QuestionnaireDesignPage";
 import NotFoundPage from "./NotFoundPage";
 import ErrorBoundary from "./ErrorBoundary";
 
+import * as Sentry from "@sentry/browser";
```
tgandrews: You can move this config to `src/index` which is where the rest of the set up is. You can also import `src/config` which allows for the environment variable to be overridden in dev.

eq-author-api/scripts/ci.sh
```diff
@@ -2,8 +2,8 @@
 set -e
 
 yarn install --frozen-lockfile
-yarn lint --max-warnings=0
-yarn test --coverage
+#yarn lint --max-warnings=0
+#yarn test --coverage
```
samiwel: Don't forget to uncomment these!

eq-author/scripts/ci.sh
```diff
@@ -2,8 +2,8 @@
 set -e
 
 yarn install --frozen-lockfile
-yarn lint
-yarn test --coverage
+#yarn lint
+#yarn test --coverage
```
samiwel: And these...

eq-author/src/index.js
```diff
@@ -15,8 +17,14 @@ import getHeaders from "middleware/headers";
 
 import App from "App";
 
-if (config.REACT_APP_SENTRY_DSN !== "") {
-  Raven.config(config.REACT_APP_SENTRY_DSN).install();
+console.log(config.REACT_APP_SENTRY_ENV);
+console.log(config.REACT_APP_SENTRY_DSN);
```
samiwel: Once working we can get rid of these these `console.log` statements.

eq-author/src/index.js
```diff
@@ -15,8 +17,14 @@ import getHeaders from "middleware/headers";
...
+if (config.REACT_APP_SENTRY_DSN) {
+  Sentry.init({
+    dsn: config.REACT_APP_SENTRY_DSN,
+    environment: config.REACT_APP_SENTRY_ENV,
+    debug: true,
```
samiwel: Set debug to false or remove altogether before merging.

eq-publisher/scripts/ci.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 yarn install --frozen-lockfile
-yarn lint
-yarn test --coverage
+#yarn lint
+#arn test --coverage
```
samiwel: Uncomment and add the missing `y` before merging

eq-author/src/apollo/createApolloErrorLink.js
```diff
@@ -1,13 +1,28 @@
...
+  if (networkError) {
     getStore().dispatch(apiDownError());
+    setSentryTag("networkError");
+    sendSentryError(networkError);
+    if (networkError.bodyText === "User does not exist") {
+      getStore().dispatch(signOutUser());
+    }
+    return;
   }
+  graphQLErrors.forEach(error => {
```
tgandrews: I presume its an empty array when there is no problem?

eq-author/src/apollo/createApolloErrorLink.test.js
```diff
@@ -1,27 +1,83 @@
...
+  beforeAll(() => {
+    /* eslint-disable-next-line camelcase */
+    const tokenData = { user_id: "Test User", email: "test.user@email.com" };
+    const token = createAccessToken(tokenData);
+    localStorage.setItem("accessToken", token);
+
+    dispatch = jest.fn();
+  });
+
+  let networkError = {
```
tgandrews: Can you define these in beforeEach so that you are guaranteed they are not shared between tests.

eq-author/src/apollo/createApolloErrorLink.test.js
```diff
@@ -1,27 +1,83 @@
...
+      networkError,
+    });
+    expect(sendSentryError).toHaveBeenCalledWith(networkError);
+  });
+
+  it("should send graphql errors to sentry", () => {
+    errorHandler(createMockStore, {
+      graphQLErrors,
+    });
+    expect(sendSentryError).toHaveBeenCalledWith(graphQLError);
```
tgandrews: That object was never given to the errorHandler can you do something like this instead.
```suggestion
    expect(sendSentryError).toHaveBeenCalledWith(graphQLErrors[0]);
```

eq-author/src/apollo/sentryUtils.test.js
```diff
@@ -0,0 +1,39 @@
...
+const createAccessToken = (token, signingKey = uuid.v4()) => {
+  return jwt.sign(token, signingKey);
+};
+
+describe("sentry utils functions", () => {
+  let networkError = {
+    fail: "Uh oh",
+  };
+
+  it("should call sentry with email and username", () => {
```
tgandrews: Can we add test when the token is missing or invalid

eq-author/src/apollo/createApolloErrorLink.js
rmccar: Yeah its just that if there is more than one graphql error it will come back as an array so had to loop through them

rmccar: If there are no errors the errorhandler wont be called

tgandrews: Good point! 🤦‍♂

eq-author/src/apollo/sentryUtils.test.js
rmccar: I think this will be covered by the test checking if the user exists in the system but but Ive added a test to make sure it does the same thing as that when the token is missing

eq-author/src/apollo/sentryUtils.js
```diff
@@ -0,0 +1,17 @@
+import * as Sentry from "@sentry/browser";
+import jwt from "jsonwebtoken";
+
+export const setSentryUser = token => {
+  const decodedToken = jwt.decode(token);
```
tgandrews: If this gets an invalid token it returns `null`. This will cause the email and user_id lookups to blow up. We then lose the graphql error.

rmccar: Ive added code and tests to stop this happening now

PR #369, opened 2019-06-11
Added firebase jwt token Verification

pricea1:
Verification of JWT token against firebase to ensure user is valid

### How to review
**IMPORTANT: Do NOT commit created file .firebase-admin-service-account.json **

1. Generate a private key file for firebase service account and save to eq-author-api/.firebase-admin-service-account.json - see [https://firebase.google.com/docs/admin/setup/](https://firebase.google.com/docs/admin/setup/)
   
   To generate a private key file for your service account:

   1. In the Firebase console, open Settings > Service Accounts.

   2. Click Generate New Private Key, then confirm by clicking Generate Key.

   3. Save the JSON file containing the key to .eq-author-api/.firebase-admin-service-account.json

2. login and use site as normal - ensure login and questionnaire editing work

3. Use dev tools to modify access token (Application > Local Storage > accessToken) with validtoken but invalid user and observe 401 error being returned from requests. Editing of questionnaire is not possible. example jwt token is: `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJodHRwczovL3NlY3VyZXRva2VuLmdvb2dsZS5jb20vb25zLWVxLWF1dGhvci1hcHAiLCJhdWQiOiJvbnMtZXEtYXV0aG9yLWFwcCIsImF1dGhfdGltZSI6MTU2MDI2MTIzOSwidXNlcl9pZCI6ImtaQXZWWHZsRkpnSkx5R0RPajg2cExFQ3FVbDIiLCJzdWIiOiJrWkF2Vlh2bEZKZ0pMeUdET2o4NnBMRUNxVWwyIiwiaWF0IjoxNTYwMjYxMjM5LCJleHAiOjE1NjAyNjQ4MzksImVtYWlsIjoiYW5kcmV3MkBsZW1vbmNvZy5jb20iLCJlbWFpbF92ZXJpZmllZCI6dHJ1ZSwiZmlyZWJhc2UiOnsiaWRlbnRpdGllcyI6eyJlbWFpbCI6WyJhbmRyZXcyQGxlbW9uY29nLmNvbSJdfSwic2lnbl9pbl9wcm92aWRlciI6InBhc3N3b3JkIn19.mP1he9JpqdOMaL1fbO0KPMSqKY6dq7vtdyBaVZvcYxA`

4. Delete accessToken from localStorage and ensure application doesn't work

### Dependencies
- https://github.com/ONSdigital/eq-author-terraform/pull/18
- https://github.com/ONSdigital/eq-deploy/pull/78


pricea1:
[Verify firebase Id tokens (4)](https://trello.com/c/O8B7ce50/1145-verify-firebase-id-tokens-4)

eq-author-api/utils/jwtVerifier.js
```diff
@@ -0,0 +1,32 @@
+const admin = require("firebase-admin");
```
tgandrews: Can we move this to next to the middleware as it is not a generic utility

eq-author-api/middleware/auth.js
```diff
@@ -0,0 +1,30 @@
...
+  const jwtToken = await jwtVerifier(accessToken);
+  if (jwtToken.error) {
+    logger.error("JWT token error", jwtToken.error);
+    res.send(401);
+    return;
+  }
+  req.auth = jwtToken;
+
+  next();
+  return;
```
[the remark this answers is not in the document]
samiwel: It is used in `server.js` in call to `createAuthMiddleware`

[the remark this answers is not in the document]
SamGodwin2: That should get handled by the `identificationMiddleware` now. I did a bit of refactoring around this in the user table stuff. Think this file might just have been brought back in a rebase somewhere?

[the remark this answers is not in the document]
pricea1: yes, I think the rebase must have brought it back in. PR originally changed auth.js but had to be refactored when rebased against Map User to Table feature. I'll remove auth.js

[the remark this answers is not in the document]
pricea1: Now removed.

eq-author-api/utils/jwtVerifier.js
pricea1: Moved

eq-author-api/middleware/identification/index.test.js
```diff
@@ -10,10 +10,25 @@ const keysJson = JSON.parse(JSON.stringify(keysYaml));
...
+        : mockJwt.decode(accessToken);
+    },
+  };
+});
+
+describe("auth middleware", async () => {
   let logger;
 
-  beforeEach(() => {
+  beforeEach(async () => {
```
heofs: Does beforeEeach need to be async? Also on line 52.

eq-author-api/middleware/identification/index.test.js
```diff
@@ -82,10 +97,19 @@ describe("auth middleware", () => {
...
+        req.header.mockImplementation(() => `Bearer invalid.access.token`);
 
-        middleware(req, res);
+        await middleware(req, res);
+
+        expect(logger.error).toHaveBeenCalled();
+        expect(res.send).toHaveBeenCalledWith(401);
+      });
+
+      it("should send a 401 response if token is valid jwt but invalid token", async () => {
```
heofs: Could this end with "but invalid signature"?

pricea1: Changed to be more explanatory.

eq-author-api/middleware/identification/index.test.js
pricea1: nope. I must have had an await call in there at some point. I've removed the async on both lines.

eq-author-api/middleware/identification/verifyJwtToken.js
```diff
@@ -0,0 +1,32 @@
+const admin = require("firebase-admin");
+
+/*
+Must define firebase service account credentials for each environment:
```
tgandrews: Can we move this to README - is this now a part of dev setup?

eq-author-api/middleware/identification/verifyJwtToken.js
```diff
@@ -0,0 +1,32 @@
...
+1. Download firebase service account json file and save in eq-author-api root as .firebase-admin-service-account.json
+    https://firebase.google.com/docs/admin/setup
+
+2. Set env variable. Create /eq-author-api/.env file containing:
+    GOOGLE_APPLICATION_CREDENTIALS=/app/.firebase-admin-service-account.json
```
tgandrews: As discussed this can be hardcoded into the docker-compose no need for an env file

tgandrews: Can it be added `.gitignore`

.gitignore
```diff
@@ -72,3 +72,4 @@ typings/
 fragmentTypes.json
 
 data/
+eq-author-api/.firebase-admin-service-account.json
```
samiwel: I wouldn't bother with this to be honest, someone could name their key something different and still check it in.

eq-author-api/docker-compose.yml
```diff
@@ -31,6 +31,7 @@ services:
       - JAEGER_SAMPLER_MANAGER_HOST_PORT=http://jaeger:5778/sampling
       - JAEGER_SAMPLER_TYPE=probabilistic
       - JAEGER_SAMPLER_PARAM=1
+      - GOOGLE_APPLICATION_CREDENTIALS=/app/.firebase-admin-service-account.json
```
samiwel: I would also not bother supplying this value. It should be a developer setup step to generate and configure this key. The app will fail to start without it, and since we're not checking keys into the source it doesn't make sense to provide the path. Or am I missing something?

.gitignore
tgandrews: I requested this as if every dev needs it and it shouldn’t be checked in then every dev would need to set this up and the necessary docker-compose change. So it’s simpler to tell them to just set it up with that name and then it just works. Are you expecting a different workflow?

samiwel: Yeah I think it's a good idea, but adding the name seems quite specific, If we are going to add  a `.gitignore` rule I think a better approach would be to instruct developers to place their keys in a certain directory e.g. `/secrets` and to `.gitignore` the secrets directory instead.

tgandrews: We could do but I can't imagine having many (if any) more. Seems fine to me but this is not a hill I'm willing to die on 😆

samiwel: Same here. Let's just go with this for now 👍

eq-author-api/docker-compose.yml
jonnyshaw89: was this question answered?

eq-author-api/middleware/identification/verifyJwtToken.js
```diff
@@ -0,0 +1,32 @@
...
+*/
+
+admin.initializeApp();
+
+const verifyJwtToken = function(accessToken) {
+  // Verify JWT token comes from Firebase
+  return admin
+    .auth()
+    .verifyIdToken(accessToken)
+    .then(function(decodedToken) {
```
jonnyshaw89:
```suggestion
    .then(function(verifiedToken) {
```

tgandrews:
@samiwel @pricea1 shall we close and delete this branch?

pricea1:
PR introduced too much complexity to the deploy pipeline. Another approach is to be considered.

pricea1:
> @samiwel @pricea1 shall we close and delete this branch?

I've closed the PR but kept the branch for now in case its needed for reference.

PR #414, opened 2019-06-18
Added filter for unowned questionnaires

SamGodwin2:
Added a toggle button to the questionnaire list page to filter out unowned questionnaires.

### How to review 
Tests pass and the questionnaires are correctly filtered.

### DIFF
--- a/eq-author/src/App/QuestionnairesPage/QuestionnairesView/Header/AccessFilter.js
+++ b/eq-author/src/App/QuestionnairesPage/QuestionnairesView/Header/AccessFilter.js
@@ -1,0 +1,100 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled, { css } from "styled-components";
+import { colors } from "constants/theme";
+import VisuallyHidden from "components/VisuallyHidden";
+
+const Fieldset = styled.div`
+  margin: 0;
+  padding: 0;
+  display: flex;
+  align-items: center;
+  color: ${colors.textLight};
+`;
+
+const Buttons = styled.div`
+  display: flex;
+`;
+
+const FilterButton = styled.label`
+  background: white;
+  padding: 0.5rem 1rem;
+  font-size: 0.9em;
+  font-weight: bold;
+  cursor: pointer;
+  border: 1px solid ${colors.bordersLight};
+  flex: 0 0 50%;
+  text-align: center;
+  position: relative;
+  white-space: nowrap;
+
+  &:first-of-type {
+    border-radius: 4px 0 0 4px;
+    margin-right: -1px;
+  }
+
+  &:last-of-type {
+    border-radius: 0 4px 4px 0;
+  }
+
+  &:hover {
+    background: ${colors.lightMediumGrey};
+  }
+
+  &:focus-within {
+    border-color: ${colors.blue};
+    outline-color: ${colors.blue};
+    box-shadow: 0 0 0 3px ${colors.tertiary};
+  }
+
+  ${props =>
+    props.checked &&
+    css`
+      color: white;
+      background: ${colors.primary};
+      border-color: #2d5e7a;
+      z-index: 1;
+
+      &:hover {
+        background: ${colors.secondary};
+      }
+    `}
+`;
+
+const AccessFilter = ({ hideUnowned, handleToggleHideUnowned }) => {
+  return (
+    <Fieldset role="group">
+      <Buttons>
+        <FilterButton checked={!hideUnowned}>
+          All
+          <VisuallyHidden>
+            <input
+              type="radio"
+              name="filter"
+              onChange={() => handleToggleHideUnowned(false)}
+              checked={!hideUnowned}
+            />
+          </VisuallyHidden>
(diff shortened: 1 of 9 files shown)


eq-author/src/App/QuestionnairesPage/QuestionnairesView/Header/AccessFilter.js
```diff
@@ -0,0 +1,94 @@
...
+      border-color: #2d5e7a;
+      z-index: 1;
+
+      &:hover {
+        background: ${colors.secondary};
+      }
+    `}
+`;
+
+const AccessFilter = ({ hideUnowned, handleToggleHideUnowned }) => {
```
tgandrews: I was confused by these names and would prefer.
```suggestion
const AccessFilter = ({ filter, onFilterChanged }) => {
```
Although this would imply the filter is a constant instead of a boolean as I'm not sure it makes sense as a boolean

eq-author/src/App/QuestionnairesPage/QuestionnairesView/index.test.js
```diff
@@ -811,5 +822,61 @@ describe("QuestionnairesView", () => {
...
+        expect(queryByText("Questionnaire 2 Title")).toBeFalsy();
+        expect(queryByText("Questionnaire 3 Title")).toBeFalsy();
+
+        fireEvent.click(showAllButton);
+
+        expect(queryByText("Questionnaire 1 Title")).toBeTruthy();
+        expect(queryByText("Questionnaire 2 Title")).toBeTruthy();
+        expect(queryByText("Questionnaire 3 Title")).toBeTruthy();
+
+        fireEvent.click(hideUnownedButton);
```
tgandrews: Could this be `showOwnedButton`? I find this double negative (hide and un) a bit confusing

eq-author/src/App/QuestionnairesPage/QuestionnairesView/Header/AccessFilter.js
heofs: I would say Sam's naming was more clear. But is it really a toggle though? Is it not just 2 buttons?

eq-author/src/App/QuestionnairesPage/QuestionnairesView/index.test.js
```diff
@@ -811,5 +810,114 @@ describe("QuestionnairesView", () => {
...
+            questionnaires={props.questionnaires}
+          />
+        );
+
+        const showAllButton = getByLabelText("All");
+        const hideUnownedButton = getByLabelText("Owner");
+
+        expect(queryByText("Questionnaire 1 Title")).toBeTruthy();
+        expect(queryByText("Questionnaire 2 Title")).toBeFalsy();
+        expect(queryByText("Questionnaire 3 Title")).toBeFalsy();
```
heofs: Are these 3 lines necessary as you are already asserting that the button works on the last 3 lines? If you want to assert that the initial state is sorting by owner, maybe move it to another test, so it is more clear what you are testing?

eq-author/src/App/QuestionnairesPage/QuestionnairesView/reducer.js
```diff
@@ -66,6 +74,7 @@ export const ACTIONS = {
   SORT_COLUMN: "SORT_COLUMN",
   REVERSE_SORT: "REVERSE_SORT",
   SEARCH: "SEARCH",
+  TOGGLE_HIDE_UNKNOWN: "TOGGLE_HIDE_UNKNOWN",
```
heofs: This should be UNOWNED, same with line 121.

eq-author/src/App/QuestionnairesPage/QuestionnairesView/index.test.js
SamGodwin2: These three lines are just to show what we expect the starting state to be, and the 2nd group asseerts we can toggle too the new mode, and the 3rd group asserts we can toggle back.

eq-author/src/App/QuestionnairesPage/QuestionnairesView/Header/AccessFilter.js
SamGodwin2: Problem with showAll is it doesn't explain what the inverse is. I.e what does `!showAll` hide?

heofs: Ah thats true. Just stick with what you think will be best then.

eq-author/src/App/QuestionnairesPage/QuestionnairesView/index.test.js
heofs: I understand that, but will it make more sense to move the first group to its own test where we are testing that the initial state should be Owned?

samiwel:
[Filter questionnaire list to display questionnaires owned by the current user (4)](https://trello.com/c/m4jRyRST/1149-filter-questionnaire-list-to-display-questionnaires-owned-by-the-current-user-4)

PR #427, opened 2019-06-25
Verify firebase tokens on the API

pricea1:
**DO NOT MERGE UNTIL** [https://trello.com/c/Zly4QhLC/1173-refresh-firebase-token-6](https://trello.com/c/Zly4QhLC/1173-refresh-firebase-token-6) has been done and ENV variables set in all environments

JWT token is now verified against Firebase public keys.
Firebase rotates its public keys every few hours so needs to refetch if current keys are invalid.

### How to review 
1. Create eq-author-env variable to be your firebaseId. e.g. create eq-author-api/.env with `FIREBASE_PROJECT_ID="YOUR_FIREBASE_ID"`

2. login and use site as normal - ensure login and questionnaire editing work

3. Change REACT_APP_FIREBASE_PROJECT_ID & REACT_APP_FIREBASE_API_KEY env variables in eq-author to point to different firebase project and restart frontend. Open Incognito window and try to login and use site and see the "We're unable to save your progress right now." message. This step created a signed jwt token but for a different project.

4. Delete accessToken from localStorage and ensure application doesn't work


pricea1:
[Verify firebase Id tokens (2)](https://trello.com/c/O8B7ce50/1145-verify-firebase-id-tokens-2)

eq-author-api/middleware/identification/index.test.js
```diff
@@ -1,5 +1,5 @@
 const identificationMiddleware = require("./");
-const jwt = require("jsonwebtoken");
+const mockJwt = require("jsonwebtoken");
```
tgandrews: Where do we mock this?

eq-author-api/middleware/identification/verifyJwtToken.js
```diff
@@ -0,0 +1,49 @@
...
+} = require("../../constants/googleServiceKeys");
+
+let googlePublicKeys;
+
+const fetchUpdatedKeys = async () => {
+  try {
+    const response = await fetch(VERIFY_PUBLIC_KEYS_URL);
+
+    googlePublicKeys = await response.json();
+  } catch (error) {
```
tgandrews: Can we do something with this error instead of suppress it?

eq-author-api/middleware/identification/verifyJwtToken.test.js
```diff
@@ -0,0 +1,76 @@
...
+  VERIFY_VALIDITY_OPTIONS,
+} = require("../../constants/googleServiceKeys");
+
+const mockKid = "MY_SECRET_KEY";
+const mockUserId = "MOCK_USER_ID";
+
+const mockCertPrivate = fs.readFileSync(
+  __dirname + "/__mock__/rsa-private.pem"
+);
+const mockCertPublic = fs.readFileSync(__dirname + "/__mock__/rsa-public.pem");
```
tgandrews: Can this just be `./__mock__/rsa-public.pem`?

pricea1: Apparently not:
`ENOENT: no such file or directory, open './__mock__/rsa-private.pem'`

tgandrews: Good point. I always forget this.

eq-author-api/constants/googleServiceKeys.js
```diff
@@ -0,0 +1,24 @@
+const logger = require("pino")();
+
+module.exports.VERIFY_PUBLIC_KEYS_URL =
+  "https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com";
+
+const JwtAlgorithms = ["RS256"];
+const JwtTokenIssUrl = "https://securetoken.google.com";
+
+if (!process.env.REACT_APP_FIREBASE_PROJECT_ID) {
```
samiwel: Can we rename to drop the `REACT_APP`? I think we only use environment variables prefixed with this on the front-end.

eq-author-api/middleware/identification/verifyJwtToken.js
samiwel: If we can't fetch the keys I don't think there's any way to reasonably recover, so we could just throw the error here.

eq-author-api/middleware/identification/verifyJwtToken.js
```diff
@@ -0,0 +1,49 @@
...
+    //    console.log({ error });
+  }
+};
+fetchUpdatedKeys();
+
+const verifyToken = (accessToken, header, userId) => {
+  const currentPublicKey = googlePublicKeys && googlePublicKeys[header.kid];
```
samiwel: I think you can return `false` early from this function if the `kid` is not in the `Object.keys(googlePublicKeys)`. No need to verify if you detect the key has been rotated.

eq-author-api/middleware/identification/verifyJwtToken.js
```diff
@@ -0,0 +1,49 @@
...
+
+  // Token verification rules as per https://firebase.google.com/docs/auth/admin/verify-id-tokens
+  // https://github.com/auth0/node-jsonwebtoken
+  try {
+    const validDecodedToken = jwt.verify(accessToken, currentPublicKey, {
+      ...VERIFY_VALIDITY_OPTIONS,
+      subject: userId,
+    });
+    return validDecodedToken;
+  } catch (error) {
```
samiwel: Might be worth logging the error?

eq-author-api/middleware/identification/verifyJwtToken.js
pricea1: I wasn't sure how to respond to this error. It will occur if the Google public service keys aren't available, which is unlikely. How about the following, which should keep the site running if keys are unavailable:

1. Log error
2. Set the googlePublickKeys variable to null. 
3. When it comes to verifying, if googlePublicKeys is null then just do a decode.
4. Next request will attempt fetch of googlePublicKeys

eq-author-api/middleware/identification/verifyJwtToken.js
pricea1: Logging added

eq-author-api/middleware/identification/verifyJwtToken.js
pricea1: Done

eq-author-api/constants/googleServiceKeys.js
pricea1: Done

eq-author-api/middleware/identification/index.test.js
pricea1: Not mocked but component within jest.mock can only reference modules with names starting with mock. I've changed it so jwt is imported within jest.mock too so that it's clearer whats happening.

eq-author-api/middleware/identification/verifyJwtToken.js
samiwel: I agree we should log the error, but I don't think we should allow unverified tokens to call our API.

tgandrews: Could we just crash?

samiwel: Probably, I think if Google is down we have bigger problems anyway.

eq-author-api/package.json
```diff
@@ -41,6 +41,7 @@
     "jsonwebtoken": "latest",
     "jsrsasign": "latest",
     "lodash": "latest",
+    "node-fetch": "^2.6.0",
```
tgandrews: Can this be latest?

eq-author-api/constants/googleServiceKeys.js
```diff
@@ -0,0 +1,24 @@
+const logger = require("pino")();
+
+module.exports.VERIFY_PUBLIC_KEYS_URL =
+  "https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com";
+
+const JwtAlgorithms = ["RS256"];
+const JwtTokenIssUrl = "https://securetoken.google.com";
+
+if (!process.env.FIREBASE_PROJECT_ID) {
```
tgandrews: Should we just process.exit?

pricea1: I've added a process.exit(1)

eq-author-api/package.json
pricea1: changed

PR #437, opened 2019-06-27
Fixed publishing of playback page

SamGodwin2:
Playback pages were broken 😱 
Switching to uuids caused a bit of regex we were using to get the ids of the page to fail so this pr fixes that. 

### How to review 
You can publish playback pages again, also the edited test passes.


SamGodwin2:
[Broken playback page](https://trello.com/c/x08fVCYZ/1190-broken-playback-page)

PR #448, opened 2019-07-01
Fix user search when user has no name

tgandrews:
Due to firebase issue it is possible for users to not have a name. This
was causing the user search to crash.

This is resolved by ensuring the name is an empty string when null.

### How to review 
1. Create a new user 
1. On master attempt to search for that new user and see the user search crash
1. Switch to this branch and see that you can now search

### DIFF
--- a/eq-author/src/components/Header/SharingModal/UserSearch.js
+++ b/eq-author/src/components/Header/SharingModal/UserSearch.js
@@ -110,7 +110,7 @@ const UserSearch = ({ users, onUserSelect }) => {
                       }
                       const value = inputValue.toLowerCase();
                       return (
-                        user.name.toLowerCase().includes(value) ||
+                        (user.name || "").toLowerCase().includes(value) ||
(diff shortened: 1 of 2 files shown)


eq-author/src/components/Header/SharingModal/UserSearch.js
```diff
@@ -110,7 +110,7 @@ const UserSearch = ({ users, onUserSelect }) => {
                       }
                       const value = inputValue.toLowerCase();
                       return (
-                        user.name.toLowerCase().includes(value) ||
+                        (user.name || "").toLowerCase().includes(value) ||
                         user.email.toLowerCase().includes(value)
```
SamGodwin2: Could we do `(user.name || user.email).toLowerCase().includes(value);`?

tgandrews: We want to search both

PR #440, opened 2019-06-28
1065 validator multiple choice answer 6

pricea1:
### What is the context of this PR?
Added validation errors to Checkbox & Radio button answers

### How to review 
1. Create at least two checkboxes.
2. Ensure validation error displayed if a label is blank
3. Ensure validation error displayed if two or more labels are equal
4. Repeat for Radio buttons

PR #452, opened 2019-07-02
Fix refresh issue on duplicated option label validation

pricea1:
…labels in option is made.

Previous [PR](https://github.com/ONSdigital/eq-author-app/pull/440) neglected to update page to re-validate after option delete 

### How to review 
1. Create options with duplicate label fields
2. Duplicate label message displayed on both options
3. Delete one option
4. Duplicate label message removed from other field
5. Do the same with OR option

### DIFF
--- a/eq-author-api/schema/typeDefs.js
+++ b/eq-author-api/schema/typeDefs.js
@@ -575,7 +575,7 @@ type Mutation {
   createOption(input: CreateOptionInput!): Option
   createMutuallyExclusiveOption(input: CreateMutuallyExclusiveOptionInput!): Option
   updateOption(input: UpdateOptionInput!): Option
-  deleteOption(input: DeleteOptionInput!): Option
+  deleteOption(input: DeleteOptionInput!): MultipleChoiceAnswer
   toggleValidationRule(input: ToggleValidationRuleInput!): ValidationRule!
   updateValidationRule(input: UpdateValidationRuleInput!): ValidationRule!
   createMetadata(input: CreateMetadataInput!): Metadata!
--- a/eq-author/src/App/page/Design/answers/withDeleteOption.js
+++ b/eq-author/src/App/page/Design/answers/withDeleteOption.js
@@ -1,36 +1,14 @@
 import { graphql } from "react-apollo";
-import { remove, get } from "lodash";
 import deleteOptionMutation from "graphql/deleteOption.graphql";
-import fragment from "graphql/answerFragment.graphql";
-
-export const createUpdater = (optionId, answerId) => proxy => {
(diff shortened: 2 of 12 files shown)


pricea1:
[Validator - Multiple choice answer (6)](https://trello.com/c/8GN0SKac/1065-validator-multiple-choice-answer-6)

eq-author/src/App/page/Design/answers/withDeleteOption.js
```diff
@@ -31,6 +31,7 @@ export const mapMutateToProps = ({ mutate }) => ({
     return mutate({
       variables: { input: option },
       update,
+      refetchQueries: ["GetPage"],
```
tgandrews: Couldn't `deleteOption` just refetch the validations for the other options instead of refetching the whole page?

tgandrews: We've done this in some other places where deleting the entity returns the parent which means the above manual cache management could be removed as well

pricea1: Refactored so deleteOption returns parent answer and removed the refetchQuery option

tgandrews:
@pricea1 you should also be able to remove the manual cache management from `withDeleteOption`

eq-author-api/schema/typeDefs.js
```diff
@@ -575,7 +575,7 @@ type Mutation {
   createOption(input: CreateOptionInput!): Option
   createMutuallyExclusiveOption(input: CreateMutuallyExclusiveOptionInput!): Option
   updateOption(input: UpdateOptionInput!): Option
-  deleteOption(input: DeleteOptionInput!): Option
+  deleteOption(input: DeleteOptionInput!): MultipleChoiceAnswer
```
samiwel: Now that the API has changed, the resolver needs to be updated to return the answer instead of the option. https://github.com/ONSdigital/eq-author-app/blob/fc7e8ea7bcc8b18751228e905c2d9a147fa2b503/eq-author-api/schema/resolvers/base.js#L347

pricea1: Now done.

pricea1:
> I think we still need a test across deleting an option that would have picked up the first issue and then secondly withDeleteOption might be able to be simplified as we shouldn't have to do the manual cache management. Although there may be an issue with mutuallyExclusiveOptions

I've simplified withDeleteOption and picked up an issue with mutuallyExclusiveOption earlier - I should push the fix for to the PR soon. I'll look at adding tests around delete too.

pricea1:
mutuallyExclusiveOption and withDeleteOption tests now added.

PR #453, opened 2019-07-02
Enable graphql playground

tgandrews:
To work with subscriptions we need a decent client and either this requires using altair (a third party client) or we can enable graphql playground which comes with graphql server.

To enable this we need to allow GET requests through to graphql which required changing the auth middleware. 

This also required updating our CSP to allow it to import scripts and run them. Unfortunately it seems to use inline scripts and styles to run so we could either enable `unsafe-inline` or take the sha from the Chrome error. I chose the latter.

### How to review 
1. Hit `/graphql` on either http://localhost:3000/graphql or http://localhost:4000/graphql
1. Ensure that after adding an `authorization` header you can make calls and get autocomplete in the graphql playground interface.

### DIFF
--- a/eq-author-api/middleware/identification/index.js
+++ b/eq-author-api/middleware/identification/index.js
@@ -4,6 +4,11 @@ const jwt = require("jsonwebtoken");
 const { getUserByExternalId } = require("../../utils/datastore");
 
 module.exports = logger => async (req, res, next) => {
+  if (req.method === "GET") {
+    next();
+    return;
+  }
+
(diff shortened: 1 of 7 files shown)


samiwel:
I tried testing this but ran into some issues on this branch not sure if I am just missing some setup. Can we take a look in the morning?

tgandrews:
Did the playground load?

tgandrews:
No those restrictions are applied in the API so a user would not be able to
see or modify questionnaires they can’t see or modify in the UI

PR #455, opened 2019-07-03
Fixed bug on creation of a routing rule on a radio answer

SamGodwin2:
When creating a routing rule or expression on a radio answer the rule would start in an invalid state as right side wasn't being initialized correctly. Fixed in this PR and right side is created correctly on radio answers.

### How to review 
On master create a radio answer then create a rule and attempt to publish without changing anything. Should fail. Repeat on this branch and it should be successful.


SamGodwin2:
[Error in publisher when previewing a questionnaire](https://trello.com/c/QQKGxdXv/1194-error-in-publisher-when-previewing-a-questionnaire)

PR #456, opened 2019-07-03
Pluralised the acres and hectares

SamGodwin2:
Last PR I forgot to pluralize hectares and acres in publisher after doing it to the front end. This caused a mismatch and made hectare and acre answers fail to publish.

### How to review 
Hectares and acres both publish now.

I wonder if it's worth having a top level constants folder to help with this kind of issue. Think theres' a number of places where the same constant is being defined in two or more different projects?


tgandrews:
We really need to solve this shared constants problem

PR #457, opened 2019-07-03
Fix bugs after header redesign

tgandrews:
This fixes issues noticed by @SamGodwin2 after the header redesign.

1. Empty state on questionnaire view was not expanding to take available
space.

2. Fix issue with read only banner not being shown in read only mode

### How to review 
1. See that the issues on master as described above are no longer created in this branch


tgandrews:
@hamishtaplin are you happy with the sign in change? This fixes the issue of the sign in resizing when there are errors shown on sign in/up form

PR #465, opened 2019-07-05
Added a check on every request to see if the access token is valid

SamGodwin2:
Currently the app allows you to use an out-of-date jwt token in all the requests to the api. As we don't validate this currently that is not a problem however now we need to check the tokens are coming from firebase in this PR: https://github.com/ONSdigital/eq-author-app/pull/427

An invalid token could be rejected by the api if it is more than 7 hours old due to google rotating the signing keys.

This PR adds a check to ensure that the access token sent to the api is valid and if it is not it will use the refresh token to ask google for a new one.

### How to review 
Log in on master and check your access token, if its invalid great, if not wait an hour (not sure how else to do this?) then when it has expired check out this branch. You should find as soon as the app loads your old expired token has been replaced with a shiny new one. 

Also tests pass

### DIFF
--- a/eq-author/src/middleware/headers/authHeader.js
+++ b/eq-author/src/middleware/headers/authHeader.js
@@ -1,3 +1,19 @@
+import jwt from "jsonwebtoken";
+import config from "config";
+
+const url = `https://securetoken.googleapis.com/v1/token?key=${config.REACT_APP_FIREBASE_API_KEY}`;
+
+const fetchNewToken = refreshToken =>
+  window
+    .fetch(url, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded",
+      },
+      body: `grant_type=refresh_token&refresh_token=${refreshToken}`,
+    })
+    .then(response => response.json());
+
 const getQuestionnaireId = () => {
   const hash = window.location.hash;
   const matches = /#\/q\/([0-9a-zA-Z-_]+)\/?/.exec(hash);
@@ -7,12 +23,6 @@ const getQuestionnaireId = () => {
 };
 
 export default headers => {
-  if (!window.localStorage) {
-    return {
-      ...headers,
-    };
-  }
-
   const accessToken = localStorage.getItem("accessToken");
   const returnedHeaders = {
     ...headers,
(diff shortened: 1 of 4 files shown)


SamGodwin2:
[Refresh firebase token (6)](https://trello.com/c/Zly4QhLC/1173-refresh-firebase-token-6)

tgandrews:
Code change looks good I'll check the refresh over the weekend or is that too long?

SamGodwin2:
@tgandrews Yep can check after the weekend, any amount of time longer than an hour should trigger it and they don't seem to be refreshed any other time.

tgandrews:
I tried this morning it didn't attempt the refresh AFAICT.

eq-author/src/middleware/headers/authHeader.js
```diff
@@ -1,3 +1,19 @@
+import jwt from "jsonwebtoken";
+import config from "config";
+
+const url = `https://securetoken.googleapis.com/v1/token?key=${config.REACT_APP_FIREBASE_API_KEY}`;
```
tgandrews: Should this be an environment variable?

SamGodwin2: Thought about it but didn't seem worth adding it too all the pipeline stuff as it's not private and it's unlikely to be reused or re-written at any point. Could go either way though, thoughts?

samiwel: Like you said, it's public URL and I think there's similar URL in the API pointing at google's public keys so I'd say don't worry about it for now. But as soon as the URL needs to change it makes sense to extract it to avoid code changes.

eq-author/src/middleware/headers/authHeader.test.js
```diff
@@ -1,44 +1,58 @@
...
+        exp: Math.floor(Date.now() / 1000) + 300,
+      },
+      "shhhhh"
+    );
+    localStorage.setItem("accessToken", token);
+
+    expect(appendAuthHeader(otherHeaders)).resolves.toMatchObject({
       ContentType: "text/html",
-      authorization: "Bearer abc.def.ghi",
+      authorization: expect.any(String),
```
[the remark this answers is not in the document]
SamGodwin2: Good point, I'll make the change.

PR #466, opened 2019-07-05
1100 validator adding the other option 4

pricea1:
Added validation for mandatory label on "Other" option.

### How to review 
1. Create at checkbox/radio answer with at least 2 options
2. Empty labels in options should have validation error and disappear when filled in. Error does not appear until input is blurred.
3. Ensure duplicate labels for options generate validation error


eq-author-api/src/validation/index.js
```diff
@@ -27,14 +36,18 @@ module.exports = questionnaire => {
...
+        // Must be in array of object type so get object type
+        // e.g. /sections/0/pages/0/answers/0/options/0/label
+        objectType = dataPath[dataPath.length - 2];
+      }
 
       const contextPath = dataPath.slice(1).join(".");
       const contextObj = get(questionnaire, contextPath);
-
       validationErrors.push({
         id: contextObj.id,
```
tgandrews: I presume this is the id of the multiple choice answer?

pricea1: No, its the ID of the object the invalid field is in. If option label is invalid then it's ID of option. If additionalAnswer.label is invalid then it's id of additionalAnswer Textfield object.

pricea1:
[Validator - Adding the "Other" option (4)](https://trello.com/c/KfJJxC2z/1100-validator-adding-the-other-option-4)

PR #469, opened 2019-07-08
Validation numbers

tgandrews:
This adds validation numbers to the left hand nav. It does nothing with isNew and so currently shows all errors.

### How to review 
1. Create a questionnaire with validation errors
1. See the error count reflect in the left hand navigation
1. See the changes in validation count updated in real time

### Dependencies
- [x] - https://github.com/ONSdigital/eq-author-terraform/pull/22


eq-author/package.json
```diff
@@ -87,6 +87,7 @@
     "apollo-link-context": "latest",
     "apollo-link-error": "latest",
     "apollo-link-http": "latest",
+    "apollo-link-ws": "^1.0.18",
```
pricea1: Should this be "latest"?

eq-author-api/schema/resolvers/withWritePermission.js
```diff
@@ -18,6 +19,10 @@ const withWritePermission = mutation => async (root, args, ctx) => {
   const result = await mutation(root, args, ctx);
   await saveQuestionnaire(ctx.questionnaire);
   ctx.validationErrorInfo = validateQuestionnaire(ctx.questionnaire);
+  pubsub.publish("validationUpdated", {
```
SamGodwin2: Could we push this down into save questionnaire, seems pretty unrelated to WritePermission and I don't think we ever want to call save and not validate?

eq-author-api/schema/tests/page.test.js
```diff
@@ -336,4 +336,69 @@ describe("page", () => {
...
+                    ],
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+      });
+
+      questionnaire = ctx.questionnaire;
```
SamGodwin2: This line and 353 aren't used.

tgandrews: They delete the query at the end.

eq-author-api/schema/resolvers/withWritePermission.js
tgandrews: Now renamed to `createMutation` so this wrapper no longer exists

tgandrews:
[Validator - number notifications (8)](https://trello.com/c/8ynDssd8/1066-validator-number-notifications-8)

tgandrews:
@pricea1 please re-review as there have been quite a few changes after the public private and additional option validation

PR #478, opened 2019-07-10
Fix button colour on touch devices

### LINKED ISSUE #446: Buttons have grey background on touch devices
### Expected behaviour

- buttons should look the same on every device

### Actual behaviour

- on devices with touchscreens (i.e. some laptops), Chrome is applying a grey background to buttons that are transparent

1. use a laptop with touchscreen
2. see screenshot for reference

#### Browser

- Chrome

#### Operating System

- Windows

### Screenshot

grey-fields-chrome

tgandrews:
Fix the button background on touch devices. I've had a look around the app but these are the only ones I found.

Closes #446 

### How to review 
1. Switch chrome to responsive view via the mobile button in dev tools.
1. See that the buttons in the right hand panel are grey.
1. See in this branch they are now the correct colour.

### DIFF
--- a/eq-author/src/App/page/PropertiesPanel/QuestionProperties/index.js
+++ b/eq-author/src/App/page/PropertiesPanel/QuestionProperties/index.js
@@ -27,6 +27,7 @@ const FieldInfo = styled(IconText)`
 
 export const HelpButton = styled.button`
   --color-text: ${colors.primary};
+  background: ${colors.white};
   margin: 0.5em 0;
   position: relative;
   left: -0.5em;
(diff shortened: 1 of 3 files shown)


tgandrews:
@codecov is lying. This makes no change to publisher and has supposedly increased it to 100% coverage.

pricea1:
[eq-author-app: #446 - Buttons have grey background on touch devices (Issue)](https://trello.com/c/WzVTLPNm/1197-eq-author-app-446-buttons-have-grey-background-on-touch-devices-issue)

Issue #273, opened 2019-03-22
Optional fields should have transitions


hamishtaplin:
### Expected behaviour

Toggling an "optional field" on a question page should result in that field transitioning in or out.

### Actual behaviour

No transitions occur.

### Steps to reproduce the behaviour

1. Create a questionnaire
1. Navigate to question page
1. Toggle an "optional field"
1. Pay attention to that field appearing or disappearing

hamishtaplin:
![](https://github.trello.services/images/mini-trello-icon.png) [Move existing optional fields to the properties panel (9)](https://trello.com/c/m7c0m75w/734-move-existing-optional-fields-to-the-properties-panel-9)

PR #486, opened 2019-07-10
Added transitions to Optional fields add/remove

pricea1:
Optional fields were not transitioning in/out

### How to review 
Toggling an "optional field" on a question page should result in that field transitioning in or out.

1. Create a questionnaire
2. Navigate to question page
3. Toggle an "optional field"
4. Pay attention to that field appearing or disappearing - master branch won't have the 200ms transition (opacity & size)

closes https://github.com/ONSdigital/eq-author-app/issues/273

### DIFF
--- a/eq-author/src/App/page/Design/QuestionPageEditor/AnswersEditor/AnswerTransition.js
+++ b/eq-author/src/App/page/Design/QuestionPageEditor/AnswersEditor/AnswerTransition.js
@@ -1,3 +1,4 @@
+import React from "react";
 import { CSSTransition } from "react-transition-group";
 import PropTypes from "prop-types";
 import styled from "styled-components";
@@ -10,7 +11,7 @@ const handleExit = node => {
   node.style.height = `${height}px`;
 };
 
-const AnswerTransition = styled(CSSTransition).attrs({
+const AnswerTransitionWrapper = styled(CSSTransition).attrs({
   classNames: "answer",
   onExit: () => handleExit,
 })`
@@ -43,6 +44,14 @@ const AnswerTransition = styled(CSSTransition).attrs({
       transform ${halfTimeout}ms ease-in;
   }
 `;
+
+const AnswerTransition = ({ children, ...otherProps }) => {
+  return (
+    <AnswerTransitionWrapper {...otherProps}>
+      <div>{children}</div>
+    </AnswerTransitionWrapper>
+  );
+};
 
 AnswerTransition.propTypes = {
   timeout: PropTypes.number,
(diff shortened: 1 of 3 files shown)


pricea1:
[eq-author-app: #273 - Optional fields should have transitions (Issue)](https://trello.com/c/YifYBqDD/1209-eq-author-app-273-optional-fields-should-have-transitions-issue)

tgandrews:
@pricea1 how does the change here fix the issue?

pricea1:
> @pricea1 how does the change here fix the issue?

AnswerTransition, a styled(CSSTransition), passes a className prop to the child which specifies the transition states e.g. answer-enter . 

The <MultipleFieldEditor /> child is a fragment so doesn't have a root element to attach the className.
The <RichTextEditor /> takes the className prop and applies it to an inner <Input /> element rather than the root.

tgandrews:
Could we update the components instead to take the className prop?

hamishtaplin:
> Could we update the components instead to take the className prop?

I prefer this too as it's not clear to the un-initiated what the `div` is for.

pricea1:
> > Could we update the components instead to take the className prop?
> 
> I prefer this too as it's not clear to the un-initiated what the `div` is for.

Just updated my previous comment as tag didn't display properly.

I'll have a look. Not a problem for \<MultipleFieldEditor /> but  \<RichTextEditor /> intentionally passes the className prop down to its \<Input /> child. Not sure of the implications of this and whether className prop would get overwritten. I'll investigate. I can't see any instances where RichTextEditor is using the className prop.

Another approach would be for \<AnswerTransition /> to wrap everything in a \<div />. Would need testing for unintended issues.

hamishtaplin:
> > > Could we update the components instead to take the className prop?
> > 
> > 
> > I prefer this too as it's not clear to the un-initiated what the `div` is for.
> 
> Just updated my previous comment as tag didn't display properly.
> 
> I'll have a look. Not a problem for <MultipleFieldEditor /> but <RichTextEditor /> intentionally passes the className prop down to its <Input /> child. Not sure of the implications of this and whether className prop would get overwritten. I'll investigate. I can't see any instances where RichTextEditor is using the className prop.
> 
> Another approach would be for <AnswerTransition /> to wrap everything in a <div />. Would need testing for unintended issues.

Sounds like it'll just get complicated. 

Keep the `div` but maybe add a comment to it explaining why it's there?

pricea1:
Given that the extra div is a requirement of \<AnswerTransition />, I've refactored so a div is wrapped around the \<AnswerTransition/> children instead. It's a neater solution and means we don't have to think about it in future. Other AnswerTransitions still seem to work as intended.

PR #495, opened 2019-07-11
Dev experience - Automatically add migration to list

tgandrews:
Update create migration to automatically to add to the list of migrations.
This means it cannot be missed by devs when they create a migration.

### How to review 
1. Ensure that when you create a migration it is created and added to the list.

### DIFF
--- a/eq-author-api/scripts/createMigration.js
+++ b/eq-author-api/scripts/createMigration.js
@@ -1,51 +1,79 @@
+/* eslint-disable no-console */
+const { R_OK } = require("fs").constants;
 const fs = require("fs").promises;
 const chalk = require("chalk");
 const { camelCase } = require("lodash");
 
-if (process.argv.length < 3) {
-  /* eslint-disable-next-line no-console */
-  console.error(
-    chalk.red(
-      "Not enough arguments provided to createMigration. Maybe the migration name is missing? Please refer to README."
-    )
-  );
-  process.exit(1);
-}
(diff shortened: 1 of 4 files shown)


eq-author-api/scripts/createMigration.js
```diff
@@ -1,51 +1,65 @@
...
+    /];/,
+    `  require('./${name}'),
+];`
+  );
+
+  await fs.writeFile(MIGRATIONS_INDEX_PATH, updatedMigrationList);
+
+  console.info(
+    chalk.green(
+      `${name} migration successfully created. Remember to add this to 'migrations/index.js'`
```
pricea1: Should this message change now it's being automatically added?

eq-author-api/scripts/createMigration.js
```diff
@@ -1,51 +1,65 @@
...
-/* eslint-disable-next-line no-console */
-console.info(
-  chalk.green(
-    `${name} migration successfully created. Remember to add this to 'migrations/index.js'`
-  )
-);
+  const currentMigrationList = await fs.readFile(MIGRATIONS_INDEX_PATH, "utf8");
+  const updatedMigrationList = currentMigrationList.replace(
+    /];/,
+    `  require('./${name}'),
```
pricea1: If create-migration is run with an existing migration name then a duplicate require statement is created.

tgandrews: Yeah, we should probably reject that.

PR #499, opened 2019-07-12
Bug - Incorrectly calculating websocket protocol

tgandrews:
Fix issue where incorrectly calculating websocket protocol
When the API url contains a protocol we were always setting it to insecure websocket

### How to review 
1. Ensure the netlify preview linked to this PR loads


tgandrews:
@codecov is lying again 🤷‍♂

PR #500, opened 2019-07-12
Remove unused dependencies and configs

heofs:
Remove some unused packages from dependencies. If there is any of them you think we should be keeping, then I can add them back in.

I removed the sass loading from webpack as we dont use .scss or .sass files, but I could add it back in if you think it would be good to have in case we want to use it later on?

### How to review 
Check that we do not rely on any of the packages.


eq-author/package.json
```diff
@@ -93,17 +93,14 @@
     "draft-js-plugins-editor": "latest",
     "draft-js-raw-content-state": "latest",
     "draftjs-filters": "^1.0.0",
-    "enzyme-to-json": "latest",
```
tgandrews: This is used by jest and defined in `package.json` line 150

eq-author/package.json
```diff
@@ -38,6 +38,7 @@
     "dotenv-expand": "latest",
     "enzyme": "latest",
     "enzyme-adapter-react-16": "latest",
+    "enzyme-to-json": "^3.3.5",
```
tgandrews: Can this target `latest`

heofs: Changed it to latest

eq-author/package.json
```diff
@@ -93,19 +93,15 @@
     "draft-js-plugins-editor": "latest",
     "draft-js-raw-content-state": "latest",
     "draftjs-filters": "^1.0.0",
-    "enzyme-to-json": "latest",
     "firebase": "latest",
     "firebaseui": "latest",
     "graphql": "latest",
     "graphql-anywhere": "latest",
-    "graphql-iso-date": "latest",
```
samiwel: Don't we still use this? https://github.com/ONSdigital/eq-author-app/blob/master/eq-author-api/schema/resolvers/base.js#L1

heofs: This is for eq-author, eq-author-api `package.json` still has `graphql-iso-date` as dependency.

heofs: This is for eq-author, eq-author-api `package.json` still has `graphql-iso-date` as dependency.

PR #506, opened 2019-07-15
Tech Debt - Remove toasts from redux

tgandrews:
Remove toasts from redux and replace with React Context and provide a HoC to show a toast message.

Odd things. Components using the `TestProvider` now rely on having a toasts div which means it needs to be created in a couple of places where we are using enzyme and not react testing library which adds it automatically.

### How to review 
1. Ensure that the toasts still work
1. Ensure the change makes sense


tgandrews:
[Remove toasts and tabs from redux](https://trello.com/c/QeDduyHt/1215-remove-toasts-and-tabs-from-redux)

eq-author/src/App/index.js
```diff
@@ -21,34 +22,36 @@ import NotFoundPage from "./NotFoundPage";
...
-          path={RoutePaths.QUESTIONNAIRE}
-          exact={false}
-          component={QuestionnaireDesignPage}
-          isSignedIn={isSignedIn}
-        />
-        <Route path="*" component={NotFoundPage} exact />
-      </Switch>
-    </ErrorBoundary>
-  </Router>
+  <ErrorBoundary>
```
pricea1: This has stopped ErrorBoundary from catching a non-existent questionnaire. Not sure why. Does it need to be within Router?

eq-author/src/App/section/Design/index.test.js
```diff
@@ -136,6 +136,9 @@ describe("SectionRoute", () => {
 
   beforeEach(() => {
     childContextTypes = { router: PropTypes.object };
+    const toasts = document.createElement("div");
```
pricea1: Do we need a Toast div? There's no reference to toast in Design/index.js now

tgandrews: Yes, unfortunately using the TestProvider relies on it :(

eq-author/src/App/index.js
tgandrews: Good spot. It was caused by the `ErrorBoundary` rendering pages with Links on that need to be wrapped in `Router`

eq-author/src/components/Toasts/ToastContainer/index.js
```diff
@@ -38,15 +30,15 @@ const ToastWrapper = styled.div`
...
   return (
     <ToastWrapper>
       <ToastOuterContainer>
         <ToastInnerContainer>
           <ToastList>
-            {map(toasts, (toast, id) => (
-              <Toast key={id} id={id} timeout={5000} onClose={dismissToast}>
-                <StyledDiv>{toast.message}</StyledDiv>
+            {toasts.map(({ id, message }) => (
+              <Toast key={id} id={id} timeout={5000} onClose={onDismissToast}>
```
heofs: Im not sure on what the conventions are, but I find it a little confusing not passing a `handleFunction` to an `onClose`?

eq-author/src/components/Toasts/ToastContainer/index.js
```diff
@@ -38,15 +30,15 @@ const ToastWrapper = styled.div`
   justify-content: center;
 `;
 
-export const ToastArea = ({ toasts, dismissToast }) => {
+export const ToastArea = ({ toasts, onDismissToast }) => {
   return (
     <ToastWrapper>
       <ToastOuterContainer>
         <ToastInnerContainer>
```
heofs: Does `ToastWrapper`, `ToastOuterContainer` and `ToastInnerContainer` only wrap the list of toasts? If so, maybe call them `ToastsWrapper` or `ToastListWrapper` so it is more clear that they are not used for the toast itself?

tgandrews: Yeah, I'll update the names it is a bit confusing. I might drop the toast prefix if it still makes sense after.

eq-author/src/components/Toasts/ToastContainer/index.js
tgandrews: We don't need the handle as is it just passing the handler down. So it would seem weird to rename it in the component.

heofs: But could you change `<ToastContainer toasts={toasts} onDismissToast={handleDismissToast} />` to `<ToastContainer toasts={toasts} handleDismissToast={handleDismissToast} />`? and then use that?

tgandrews:
That doesn’t make sense as an API for ToastContainer.

eq-author/src/components/Toasts/ToastContainer/index.js
tgandrews: As discussed leaving as is

PR #508, opened 2019-07-15
Tech Debt - Remove tabs from redux

tgandrews:
This removes the tab state from redux and just makes it a prop on the component.

### How to review 
1. Ensure answer validation modal still works (the only place this is used)


tgandrews:
[Remove toasts and tabs from redux](https://trello.com/c/QeDduyHt/1215-remove-toasts-and-tabs-from-redux)

eq-author/src/components/modals/ModalWithNav/ModalWithNav.test.js
```diff
@@ -55,37 +55,57 @@ const NAV_ITEMS = [
...
-      gotoTab: jest.fn(),
+      startingTabId: "tab-1",
+      isOpen: true,
       navItems: NAV_ITEMS,
     };
   });
 
+  it("should not render anything when isOpen is false", () => {
+    const { queryByText } = render(<ModalWithNav {...props} isOpen={false} />);
+    expect(queryByText("Tellus Dollor")).toBeFalsy();
```
heofs: I think this will be `falsy` even if it is open (misspelled `Dolor`). And maybe add a check to see that it is `truthy` when it is open.

Issue #180, opened 2019-02-01
Deleting a routing rules destination doesn't then update the destination. 


SamGodwin2:
### Expected behaviour

1. Given 3 pages, 1,2 and 3.
2. Make page 1 route to 3.
3. Delete page 3.
4. Navigate back to page 1's routing.
5. Page 1 should now be routing to the next page.

### Actual behaviour

Given the steps above when you reach step 5 the app will crash due to the destination no longer being valid.

pricea1:
Fixed by
https://trello.com/c/2KGXE7Zd/1214-eq-author-app-180-deleting-a-routing-rules-destination-doesnt-then-update-the-destination-issue

PR #510, opened 2019-07-16
Removed routing rules referencing deleted page

pricea1:
Deleting a page that has been used as a destination in routing crashes the app when returning to edit the routing is appears on.

### How to review 
1. Create at least 3 pages.
2. Add 2 rules to page 1, one referencing page2, one referencing page3
3. Delete page2
4. View page1 routing: Only routing to page3 should exist
5. Delete page3
6. View page1 routing: No routing should be displayed

### DIFF
--- a/eq-author-api/src/businessLogic/onPageDeleted.js
+++ b/eq-author-api/src/businessLogic/onPageDeleted.js
@@ -1,0 +1,23 @@
+const { getPages } = require("../../schema/resolvers/utils");
+
+const onPageDeleted = (ctx, removedPageId) => {
+  const allPages = getPages(ctx);
+
+  allPages.forEach(page => {
+    if (!page.routing) {
+      return;
+    }
+
+    const validRules = page.routing.rules.filter(
+      rule => rule.destination && rule.destination.pageId !== removedPageId
+    );
+
+    if (validRules.length) {
+      page.routing.rules = validRules;
+    } else {
+      page.routing = null;
+    }
+  });
+};
+
+module.exports = onPageDeleted;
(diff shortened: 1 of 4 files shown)


pricea1:
[eq-author-app: #180 - Deleting a routing rules destination doesn't then update the destination.  (Issue)](https://trello.com/c/2KGXE7Zd/1214-eq-author-app-180-deleting-a-routing-rules-destination-doesnt-then-update-the-destination-issue)

eq-author-api/src/businessLogic/onPageDeleted.js
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    const validRules = page.routing.rules.filter(
+      rule => rule.destination && rule.destination.pageId !== removedPageId
+    );
+
+    if (validRules.length) {
+      page.routing.rules = validRules;
+    } else {
+      page.routing = null;
```
tgandrews: Can we not set it to an empty array of rules?

eq-author-api/tests/utils/contextBuilder/routing/updateRouteDestination.js
```diff
@@ -0,0 +1,35 @@
+const { updateRightSide } = require("./updateRightSide");
+const { updateRoutingRule } = require("./updateRoutingRule");
+
+const updateRouteDestination = async (
+  ctx,
+  page,
+  optionIndex,
+  destinationPageId
+) => {
+  //  console.log({ page: JSON.stringify(page.routing.rules) });
```
tgandrews: These commented out log messages can go

eq-author-api/src/businessLogic/onPageDeleted.js
pricea1: No, it needs to null page.routing otherwise we're just left with an else statement but no rules. Setting page.routing to null puts it back into the same state as it was before rules were added.

eq-author-api/tests/utils/contextBuilder/routing/updateRouteDestination.js
pricea1: Removed

PR #513, opened 2019-07-17
Disable view survey if validation errors

heofs:
<https://trello.com/c/S4FNVqG1/1096-disabling-view-survey-4>
Added a function that checks if there is any validation errors and disables the view survey button if any errors are found. 

### How to review 
1. Create a new questionnaire
1. Button should be immediately disabled because you haven't added any questions.
1. Add question text and blur out of the field, the button should be enabled again.
1. Try creating new answers and questions and see that the behaviour is consistent.

### DIFF
--- a/eq-author-api/schema/resolvers/base.js
+++ b/eq-author-api/schema/resolvers/base.js
@@ -514,6 +514,8 @@ const Resolvers = {
       }
       return "Read";
     },
+    totalErrorCount: (questionnaire, args, ctx) =>
+      ctx.validationErrorInfo.totalCount,
   },
 
   User: {
--- a/eq-author/src/components/EditorLayout/Header/index.test.js
+++ b/eq-author/src/components/EditorLayout/Header/index.test.js
@@ -16,6 +16,7 @@ describe("Header", () => {
     questionnaire = {
       id: "456",
       displayName: "Questionnaire of Awesomeness",
+      totalErrorCount: 0,
       createdBy: {
         id: "1",
         name: "Pinky Malinky",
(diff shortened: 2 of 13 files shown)


eq-author-api/schema/resolvers/base.js
```diff
@@ -514,6 +514,8 @@ const Resolvers = {
       }
       return "Read";
     },
+    totalErrorCount: (questionnaire, args, ctx) =>
+      ctx.validationErrorInfo ? ctx.validationErrorInfo.totalCount : 0,
```
tgandrews: Why is this check necessary?

heofs: The 3 tests under `duplicate a questionnaire` in `eq-author-api/schema/tests/duplication.test.js` was failing without it. Giving totalCount is undefined.

tgandrews: I guess this is an issue with duplication not validating the questionnaire

eq-author/src/components/EditorLayout/Header/index.test.js
```diff
@@ -92,14 +93,33 @@ describe("Header", () => {
...
+
+      const viewSurveyButton = getByTestId("btn-preview");
+      expect(viewSurveyButton.getAttribute("disabled")).toBe("");
+    });
+
+    it("should not disable the view survey button when questionnaire is valid", () => {
+      const { getByTestId } = renderWithContext(<Header {...props} />);
+
+      const viewSurveyButton = getByTestId("btn-preview");
+      expect(viewSurveyButton.getAttribute("disabled")).toBeFalsy();
```
tgandrews: Empty string is also falsy. So we can check for something explicitly like `null` or `undefined`

eq-author-api/schema/resolvers/base.js
```diff
@@ -514,6 +514,8 @@ const Resolvers = {
       }
       return "Read";
     },
+    totalErrorCount: (questionnaire, args, ctx) =>
```
tgandrews: Can we add a test on the server api for this.

eq-author/src/components/EditorLayout/Header/index.test.js
heofs: Aah thats right, I changed it to check for `null`.

eq-author-api/schema/resolvers/base.js
heofs: doesnt `totalErrorCount: expect.any(Number),` in `eq-author-api/schema/tests/questionnaire.test.js` test this properly?

tgandrews: Good point I missed that.

eq-author/src/components/EditorLayout/Header/index.test.js
SamGodwin2: Might it be worth bringing in something like https://github.com/testing-library/jest-dom to assert the button is disabled. This way of doing it doesn't feel very rtl-ly if that makes sense?

tgandrews: That sounds like a much better idea. Good find.

heofs: Now checks using this. And fixed the button so you no longer can tab into it.

tgandrews: Whats wrong with being able to tab to it? https://ux.stackexchange.com/questions/103239/should-disabled-elements-be-focusable-for-accessibility-purposes

heofs: The user was able to tab into it and click enter to still go to the link, even when it was disabled. Hamish suggests not being able to tab into it. And there is no point being able to tab into something you cant interact with.

eq-author/src/components/EditorLayout/Header/index.test.js
```diff
@@ -111,14 +111,14 @@ describe("Header", () => {
       const { getByTestId } = renderWithContext(<Header {...props} />);
 
       const viewSurveyButton = getByTestId("btn-preview");
-      expect(viewSurveyButton.getAttribute("disabled")).toBe("");
+      expect(viewSurveyButton).toHaveAttribute("disabled");
```
tgandrews: You can use `toBeDisabled`

PR #516, opened 2019-07-18
Bug - Questionnaire not loading when it contains calculated summary

tgandrews:
Fix issue where not loading when questionnaire contains a calculated summary

### How to review 
1. Create a questionnaire with a calculated summary
1. See it does not load 
1. See with this branch the questionnaire does load


tgandrews:
@heofs yes I think it would have done.

PR #521, opened 2019-07-19
Bump publisher dependency versions.

samiwel:
Bumped publisher dependency versions and fixed some lining errors that manifested as a result.

### How to review
All tests and checks pass.

### DIFF
--- a/eq-publisher/src/middleware/fetchData.js
+++ b/eq-publisher/src/middleware/fetchData.js
@@ -1,5 +1,9 @@
 const { get } = require("lodash");
 
+const setQuestionnaire = async (res, questionnaire) => {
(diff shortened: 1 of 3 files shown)


eq-publisher/src/middleware/fetchData.js
```diff
@@ -1,5 +1,9 @@
 const { get } = require("lodash");
 
+const setQuestionnaire = async (res, questionnaire) => {
```
SamGodwin2: Is this async needed?

samiwel: Apparently not, eslint satisfied without it 👍

PR #529, opened 2019-07-23
Add unit label validation

tgandrews:
Add validation for label on units

### How to review 
1. Ensure that unit answers must have a label and if they don't an error is shown. The error should also be shown in the left hand as an additional error on the page.

### DIFF
--- a/eq-author-api/constants/answerTypes.js
+++ b/eq-author-api/constants/answerTypes.js
@@ -9,6 +9,8 @@ const DATE_RANGE = "DateRange";
 const PERCENTAGE = "Percentage";
 const UNIT = "Unit";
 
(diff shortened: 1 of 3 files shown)


tgandrews:
[Show validation errors for units](https://trello.com/c/02aExn8N/1224-show-validation-errors-for-units)

eq-author-api/constants/answerTypes.js
```diff
@@ -9,6 +9,8 @@ const DATE_RANGE = "DateRange";
 const PERCENTAGE = "Percentage";
 const UNIT = "Unit";
 
+const BASIC_ANSWERS = [TEXTFIELD, TEXTAREA, CURRENCY, NUMBER, UNIT];
```
heofs: Is this whole file or only this line for testing? If BASIC_ANSWER is only used for testing, maybe give it a comment so it is obvious that the application doesn't depend on it?

PR #531, opened 2019-07-24
Fix flashing side bar when adding question page

tgandrews:
Depending on how you load your cache you can create a situation where
an error is thrown when you add a question page. This causes a
noticeable flash to the user.

This is because apollo does not seem to be able to merge structures on
an object unless that is an object with its own id and normalized in the
apollo cache.

This change now returns the validation error info with an id so it can
be normalized. This then means that the cache is merged correctly.

### How to review 
1. Add question pages on question pages and section pages
1. See that on master sometimes you will see the sidebar flash and an error in the console.
1. See that on this branch that this is resolved

### DIFF
--- a/eq-author-api/schema/resolvers/base.js
+++ b/eq-author-api/schema/resolvers/base.js
@@ -530,7 +530,7 @@ const Resolvers = {
       getPreviousAnswersForSection(ctx.questionnaire, id),
     availablePipingMetadata: (section, args, ctx) => ctx.questionnaire.metadata,
     validationErrorInfo: ({ id }, args, ctx) =>
-      ctx.validationErrorInfo.sections[id] || { errors: [], totalCount: 0 },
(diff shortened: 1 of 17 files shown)


heofs:
I still seem to get the error when trying to add a question page. ```Warning: Failed prop type: The prop `questionnaire` is marked as required in `MovePageModal`, but its value is `undefined`.```

tgandrews:
@heofs you must start from a clean page refresh (you can't rely on HMR) as this bug is all about inconsistent data in the apollo cache.

Additionally I force pushed a few small changes so please make sure you are up to date.

If the above two is true do you know of the recreation steps?

tgandrews:
@heofs 	good spot. I think it should be resolved now. Here are the recreation steps I followed that caused the same error.
1. Create a social survey
2. From that first page create a section

PR #532, opened 2019-07-24
Improve handling of error that was observed in production.

samiwel:
There was an issue observed when debugging a live survey where additional information had been toggled on for a question but the question did not have any answers. This should no longer be an issue once #530 has been merged, but there's still a potential to run into this error when hitting the Publisher endpoint directly.

This PR handles the error case to provide some context instead of letting an unhandled runtime error to occur.

### How to review 
All tests and checks pass.


samiwel:
I'm going to merge this, the Codecov change is bogus.

PR #537, opened 2019-07-25
Redirect to frontpage when already signed in

heofs:
This fixes the issue with an already signed in user being able to go to the `/sign-in` page. 

### How to review 
See that a logged in user will be redirected to the list of questionnaires if already signed in.


samiwel:
[Redirect user away from login page when already logged in](https://trello.com/c/xLOZK6qU/1234-redirect-user-away-from-login-page-when-already-logged-in)

PR #538, opened 2019-07-25
Give admins write access to all questionnaires.

samiwel:
Currently only owners and editors have write access. We have the ability to specify "Admin" users controlled outside of Author as an offline process.

This PR allows users who have been granted admin privileges write access to all questionnaires.

### How to review 
Spin up an environment.
Create two users.
With user 1 create a questionnaire.
Login as user 2.
You should not be able to change anything on the questionnaire.
Make user 2 an admin. Refresh the page.
You should now be able to make changes to the questionnaire.

### DIFF
--- a/eq-author-api/middleware/loadQuestionnaire.js
+++ b/eq-author-api/middleware/loadQuestionnaire.js
@@ -1,4 +1,5 @@
 const { getQuestionnaire } = require("../utils/datastore");
+const { get } = require("lodash");
 
 module.exports = async (req, res, next) => {
   const questionnaireId = req.header("questionnaireId");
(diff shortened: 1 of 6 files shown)


eq-author-api/middleware/loadQuestionnaire.js
```diff
@@ -9,7 +10,11 @@ module.exports = async (req, res, next) => {
 
   const questionnaire = await getQuestionnaire(questionnaireId);
 
-  if (questionnaire && questionnaire.isPublic === false) {
+  const notAdmin = !get(req, "user.admin", false);
```
SamGodwin2: Instead of using inverted logic could we do something like:
```
const isAdmin = get(req, "user.admin")

if(isAdmin){
	req.questionnaire = questionnaire;
	next();
	return; 
}
```
and just short circuit the whole thing?

samiwel:
[Allow eQ Author/Runner team to view private questionnaires even if they are not an editor](https://trello.com/c/ACJOuhvI/1223-allow-eq-author-runner-team-to-view-private-questionnaires-even-if-they-are-not-an-editor)

PR #545, opened 2019-07-29
Fix validation error info not returning an id field.

samiwel:
When resolving a question page validationErrorInfo field, an Id value is not being returned but it is required in the graphql schema.
This is causing adding an answer to a page to break.

### How to review 
all tests should pass, should be able to add answers to a page.


samiwel:
[Unable to add an answer](https://trello.com/c/2frp5iSs/1241-unable-to-add-an-answer)

PR #546, opened 2019-07-29
Fix issue when adding routing rule to a page that has no answers.

samiwel:
When creating a routing rule, there was an incorrect assumption being made in the business logic that the page will always have an answer. This fix checks if an answer exists and sets the routing rule right hand side value accordingly.

### How to review
Changes look good. Tests pass.
Should be able to add a routing rule to a page that has no answers.

### DIFF
--- a/eq-author-api/src/businessLogic/createRightSide.test.js
+++ b/eq-author-api/src/businessLogic/createRightSide.test.js
@@ -1,0 +1,21 @@
+const createRightSide = require("./createRightSide");
+const { RADIO, NON_RADIO_ANSWERS } = require("../../constants/answerTypes");
+
+describe("createRightSide", () => {
+  it("should return undefined when there is no answer on the page", () => {
+    expect(createRightSide(null)).toBeUndefined();
+  });
+
+  it("should return empty selected options array for radio answers", () => {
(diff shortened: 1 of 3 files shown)


samiwel:
[Unable to add routing rule to a page with no answers.](https://trello.com/c/c0xWATjd/1242-unable-to-add-routing-rule-to-a-page-with-no-answers)

eq-author-api/src/businessLogic/createRightSide.test.js
```diff
@@ -0,0 +1,21 @@
+const createRightSide = require("./createRightSide");
+const { RADIO, NON_RADIO_ANSWERS } = require("../../constants/answerTypes");
+
+describe("createRightSide", () => {
+  it("should return undefined when there is no answer on the page", () => {
+    expect(createRightSide(null)).toBeUndefined();
+  });
+
+  it("should return empty selected options array fro radio answers", () => {
```
heofs: should it be from not fro?

samiwel: Actually it should be `for`, good spot!

samiwel:
```suggestion
  it("should return empty selected options array for radio answers", () => {
```

PR #551, opened 2019-07-31
Fix some bugs in login code

SamGodwin2:
Recently there has been some instability around the sign-in screen since it got removed from the redux store. This Pr aims to resolve that and have it fail in more reasonable ways. 

### How to review 
Tests pass, and you cant break it whatever you do. (Try signing in with a down api, edit the access tokens, etc )

### DIFF
--- a/eq-author/.env.test
+++ b/eq-author/.env.test
@@ -1,2 +1,6 @@
 REACT_APP_API_URL="http://localhost:4000/graphql"
 REACT_APP_LAUNCH_URL="http://localhost:4000/launch"
+REACT_APP_FIREBASE_PROJECT_ID="eq-author-test"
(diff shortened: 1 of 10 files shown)


samiwel:
[Bug around the new login implementation.](https://trello.com/c/Mp8ncp6d/1244-bug-around-the-new-login-implementation)

samiwel:
Seeing this error on console when first visiting the sign in page. 
image

SamGodwin2:
> Seeing this error on console when first visiting the sign in page.
> 
> image
> 
> The same error is observed on an anonymous component when signing out.

Fixed, forgot that the isSigningIn can be null when we're waiting on everything.

PR #552, opened 2019-07-31
Fixed deleting of page used in ELSE route of another page

pricea1:
Routing page crashed when page or section referred to in ELSE route is deleted

### How to review 
1. Create 3 pages
2. Create a page route and specify ELSE destination another page
3. Delete the ELSE destination page
4. ELSE destination of page route should now be changed to next page

Repeat for using Section pages as rule and else destinations

### DIFF
--- a/eq-author-api/src/businessLogic/onPageDeleted.js
+++ b/eq-author-api/src/businessLogic/onPageDeleted.js
@@ -1,4 +1,5 @@
 const { getPages } = require("../../schema/resolvers/utils");
+const { NEXT_PAGE } = require("../../constants/logicalDestinations");
 
 const onPageDeleted = (ctx, removedPageId) => {
   const allPages = getPages(ctx);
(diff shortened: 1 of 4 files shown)


pricea1:
[Deleting page when referred to in routing as else destination](https://trello.com/c/dPqAz4Kg/1222-deleting-page-when-referred-to-in-routing-as-else-destination)

eq-author-api/src/businessLogic/onPageDeleted.js
```diff
@@ -8,6 +9,12 @@ const onPageDeleted = (ctx, removedPageId) => {
       return;
     }
 
+    const elseRoute = page.routing.else;
+    if (elseRoute.pageId === removedPageId) {
+      elseRoute.pageId = null;
```
samiwel: Probably worth just ensuring that the sectionId is nulled out here as well.

```js
page.routing.else = {
  sectionId: null,
  pageId: null,
  logical: NEXT_PAGE
}

pricea1: sectionId now also nulled

pricea1:
> Could we handle deleted sections in the else in the same way as they currently suffer from the same issue as these pages?

Deleted sections now handled as else and rule destination

PR #553, opened 2019-08-01
Prevent empty data from rendering move page modal.

samiwel:
There seems to be an issue where sometimes, possibly when a value is both simultaneously being mutated and queried from the cache, where a Query component will return an empty data object, even though the query has finished loading and there are no errors.

This was causing our application to crash because the modal assumes that it is being passed a valid data object, based on assumptions being made in the PropTypes.

This is a known issue and this fix uses refetchPartial which according to the docs this is fine however we may have to keep an eye on this issue as it moves along apollographql/react-apollo/issues/2114

### How to review 
1. Checkout master
1. Add question page
1. Observe crash and error
1. Checkout this branch
1. Add a question page
1. Should no longer crash.


samiwel:
[Error when adding page](https://trello.com/c/uEAag9Wv/1245-error-when-adding-page)

PR #555, opened 2019-08-02
Fix stay on page when refreshing.

samiwel:
There was a small bug where refreshing the page would take you back to the list page.
This PR aims to fix.

The issue seemed to be that the history location state was undefined when refreshing after returning from the signIn fetch call.

This PR aims to fix by using the current pathname where possible and falling back to the list page when it is not possible to do that.

### How to review 
- Checkout master.
- Create a questionnaire.
- Go to the questionnaire design page.
- Refresh the browser.
- It will take you back to the questionnaire list page.
- Repeat on this branch and observe that you stay on the same page.

### DIFF
--- a/eq-author/src/App/MeContext.js
+++ b/eq-author/src/App/MeContext.js
@@ -25,7 +25,13 @@ const signIn = (setSignInSuccess, history, user) => {
       if (!res.ok) {
         throw Error(`Server responded with a ${res.status} code.`);
       }
-      history.push(get(history, "location.state.returnURL", "/"));
+      history.push(
+        get(
+          history,
+          "location.state.returnURL",
+          get(history, "location.pathname", "/")
+        )
+      );
       setSignInSuccess(true);
     })
     .catch(e => {


heofs:
It works well, but is it possible to make it behave more like it does in prod at the moment? So on this branch, if you refresh the page, the header and body flashes with `Logging in...`, which seems a little strange when you are just navigating the app.

SamGodwin2:
@heofs I could be wrong but I'm not sure this is easily possible, the issue is the only way that we know on our end that someone is logged in after a refresh is by looking at the accessToken in local storage. If we lean on this however it opens up a bunch of issues where the access token is present but invalid. To be fair refreshing the page feels like a pretty abnormal thing to do in an app like this and I'd argue that maybe it should be logging you back in after a refresh.

heofs:
In firebase you can do this to check if the user is signed in. I believe it also verifies that the token is valid.

```
firebase.auth().onAuthStateChanged(function(user) {
  if (user) {
    // User is signed in.
  } else {
    // No user is signed in.
  }
});
```

samiwel:
[Fix stay on page when refreshing](https://trello.com/c/R4Sg1wDU/1247-fix-stay-on-page-when-refreshing)

PR #560, opened 2019-08-06
Fixed adding new section errors

pricea1:
Two errors identified when adding new a new section to a new business survey.

1. Create questionnaire & immediately add section results in console error

2. Select a newly created section, add a new page (page should be between section and existing page), immediately delete created page results in no page being highlighted in nav some of the time

### How to review 
Confirm above happens on master but not on this branch


pricea1:
[Errors when adding new section to new business survey](https://trello.com/c/cYj7QDsO/1251-errors-when-adding-new-section-to-new-business-survey)

PR #564, opened 2019-08-07
Simplify setup of GTM.

samiwel:
The way in which we were setting up GTM was more complex than it needed to be.

There is an existing library which integrates nicely with React applications which takes care of injecting the relevant GTM snippets.

We were also hard coding the preview Id which made it impossible to successfully turn on GTM in both pre-prod and production environments.

This change renames the existing ENV_ID environment variable to GTM_AUTH and introduces a separate GTM_PREVIEW environment variable to simplify the configuration across environments and bring the terminology inline with what GTM calls these variables.

### How to review 
Set the environment variables appropriately.
Observe that google tag manager is injected into the page source.

### Dependencies
- https://github.com/ONSdigital/eq-author-terraform/pull/26

### DIFF
--- a/eq-author/README.md
+++ b/eq-author/README.md
@@ -81,10 +81,13 @@ The following environment variables should be placed in a `.env` file in the eq-
 ### Third party services
 
 | Name                      | Description                                     | Required |
-| ------------------------- | ----------------------------------------------- | -------- |
+|---------------------------|-------------------------------------------------|----------|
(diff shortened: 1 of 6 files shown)


samiwel:
[Set up GTM on Author](https://trello.com/c/5Gdtb0Ni/1264-set-up-gtm-on-author)

samiwel:
@rmccar Yeah you're right Rich. On runner the preview ID is being passed in, but on Author it was hard coded. I've just introduced another environment variable to get away from hard coding it.

@heofs Good spot I will fix.

PR #565, opened 2019-08-07
Prevent validation errors from disappearing in nav

heofs:
This PR fixes the issue observed in the user session, with the validation errors disappearing half way behind the question nav.

### How to review 
1. Checkout master
1. Create a questionnaire and add a very long title to it.
1. Add an answer but with no content, so the validation error appears.
1. See that the error is half way behind the nav.
1. Checkout this branch and see that the error is fixed.
1. Also check that the fix does not affect any other components.


heofs:
[Bug - Validation errors disappears behind left navbar](https://trello.com/c/i7WcXmCP/1263-bug-validation-errors-disappears-behind-left-navbar)

PR #566, opened 2019-08-08
Enable deletion of last character in validation

heofs:
This PR fixes a bug that prevented a user from deleting the last character or number when creating a custom validation rule.

The following change was necessary because 0 was evaluated to false.
`{enabled && !isNaN(parseInt(value, 10)) && <Detail>{value}</Detail>}`

### How to review 
1. Create a new questionnaire with a number answer.
1. Create a new min or max validation rule on the number.
1. Fill in a number and try delete it. This should be possible on this branch, not on master.
1. Try with different answer types, for example percentage or currency.

### DIFF
--- a/eq-author/src/App/page/Design/Validation/AnswerValidation.js
+++ b/eq-author/src/App/page/Design/Validation/AnswerValidation.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { kebabCase, get, startCase } from "lodash";
+import { kebabCase, get, startCase, isNull } from "lodash";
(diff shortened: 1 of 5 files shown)


samiwel:
[Bug - Not being able to delete last number on answer validation](https://trello.com/c/C0yibTl5/1266-bug-not-being-able-to-delete-last-number-on-answer-validation)

heofs:
[Bug - Not being able to delete last number on answer validation](https://trello.com/c/C0yibTl5/1266-bug-not-being-able-to-delete-last-number-on-answer-validation)

PR #567, opened 2019-08-08
Fix answer validation button flickering

heofs:
This PR fixes the button on answer validation so that it no longer jumps back and forth when you click it. The fix is done by preventing the component to do unnecessary updates.

_I still need to write the test for this, so if you guys are okay with this fix, I will go ahead and do that._

### How to review 
1. Create a new questionnaire with an answer that can validated.
1. Check that the component behaves at it should.
1. Check that when you switch between `Previous answer` and `Custom` that there is no flickering of the buttons.

### DIFF
--- a/eq-author/src/App/page/Design/Validation/AnswerValidation.js
+++ b/eq-author/src/App/page/Design/Validation/AnswerValidation.js
@@ -128,7 +128,7 @@ const validations = [
   {}
 );
 
-class AnswerValidation extends React.Component {
+class AnswerValidation extends React.PureComponent {
   state = {
     startingTabId: null,
     modalIsOpen: false,


pricea1:
> If all this is doing it checking that props have changed and only re-rendering in that case we could just make it a PureComponent?

I think PureComponents check shallow changes to props but this looks like it delved deeper. Worth checking though as PureComponent should be used over shouldComponentUpdate if possible as it performs better.

heofs:
I tried making it a purecomponent and it would not work properly.

heofs:
First thing I did was make it a PureComponent and figured out it didn't work. And I was about to test it again to tell you what was not working properly, but it seems it actually works fine now 🙃 I have committed the change, so if you guys could also give it a good test, then this seems to be the best approach for the moment!

SamGodwin2:
Pure component should work here as the on the first re-render the props aren't being changed to the new validation type, then after its being re-rendered with new props causing the flash. 
If we stop that first re-render with Pure it should prevent the flash. I would kinda like to know what causing the first re-render but this is probably a good enough fix for now.

PR #574, opened 2019-08-14
Implemented publish Survey for admins only

SamGodwin2:
Added the button to publish a survey

To spin up locally: 

1. Checkout runner and remove https://github.com/ONSdigital/eq-survey-runner/blob/413b66e5b8278f9708ab67c4cd9fecddfd189fd1/docker-compose.yml#L27 to line 33.
2. Spin up runner using `docker-compose up`
3. Make sure Author, the api and publisher are all running.
4. Checkout https://github.com/ONSdigital/eq-survey-register/pull/2 and spin it up using `docker-compose up`
5. Make yourself an admin by following the instructions in https://github.com/ONSdigital/eq-author-app/pull/359
6. See the new publish button in the header, click it and follow the link! 

### How to review 
Tests pass and pipeline works.

### DIFF
--- a/eq-author-api/schema/resolvers/base.js
+++ b/eq-author-api/schema/resolvers/base.js
@@ -18,6 +18,7 @@ const {
 const GraphQLJSON = require("graphql-type-json");
 const uuid = require("uuid");
 const { withFilter } = require("apollo-server-express");
+const fetch = require("node-fetch");
 
 const pubsub = require("../../db/pubSub");
 const { getName } = require("../../utils/getName");
@@ -161,6 +162,17 @@ const Resolvers = {
     },
     me: (root, args, ctx) => ctx.user,
     users: () => listUsers(),
+    triggerPublish: async (root, { questionnaireId }) => {
+      const result = await fetch(
+        `${process.env.SURVEY_REGISTER_URL}${questionnaireId}`,
+        { method: "put" }
+      )
+        .then(res => res.json())
+        .catch(e => {
+          throw Error(e);
+        });
+      return { id: questionnaireId, launchUrl: result.publishedSurveyUrl };
+    },
   },
 
   Subscription: {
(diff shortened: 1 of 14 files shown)


SamGodwin2:
[Publish and launch from a registry (9)](https://trello.com/c/8nyY3EJv/1198-publish-and-launch-from-a-registry-9)

SamGodwin2:
Did you have publisher running? Just realised I forgot to mention it in the PR and that would explain the error on 9000?

pricea1:
> Did you have publisher running? Just realised I forgot to mention it in the PR and that would explain the error on 9000?

No - working ok now with Publisher running

eq-author/src/components/EditorLayout/Header/index.test.js
```diff
@@ -100,6 +103,71 @@ describe("Header", () => {
...
+              },
+            };
+          },
+        },
+      ];
+      const { getByTestId } = renderWithContext(<Header {...props} />, {
+        mocks,
+      });
+
+      const publishSurveyButton = getByTestId("btn-publish");
```
heofs: Have you switched to the dark side and started using test ids now? 😌

eq-author-api/schema/resolvers/base.js
```diff
@@ -161,6 +162,16 @@ const Resolvers = {
     },
     me: (root, args, ctx) => ctx.user,
     users: () => listUsers(),
+    triggerPublish: async (root, { questionnaireId }) => {
+      const result = await fetch(
+        `${process.env.SURVEY_REGISTER_URL}${questionnaireId}`
```
samiwel: I think this is fine as a first pass, but I think when we look at adding `eq_id` and `form_type` in the next story, this should be changed to POST to the registry instead. It's more semantically correct. The use of HTTP GET verb implies that we're querying, when actually we're adding something to the registry.

SamGodwin2: Yeah you're right it really should be a fetch, it's a pretty small change both here and in the register so I'll just add it here.

PR #575, opened 2019-08-14
Added validation errors to calculated summary page

pricea1:
Validation errors were not being displayed on calculated summary page

### How to review 
1. Create calculated summary page
2. Ensure validation error & navigation menu item error indicator appears for the title field


pricea1:
[Calculated summary is being validated without showing the errors (2)](https://trello.com/c/UGu6712l/1228-calculated-summary-is-being-validated-without-showing-the-errors-2)

samiwel:
LGTM and works as expected. I wonder if we should also add validation for the case when there are no answers selected. There wasn't any acceptance criteria for this on the card, so we could deal with it seperately, but not selecting any answers on calculated summary pages fails schema validation when previewing in runner.

pricea1:
> LGTM and works as expected. I wonder if we should also add validation for the case when there are no answers selected. There wasn't any acceptance criteria for this on the card, so we could deal with it seperately, but not selecting any answers on calculated summary pages fails schema validation when previewing in runner.

I've added validation to the question too.

samiwel:
There's a failing test because the answer count has changed. Also when I tested, the red box displays as expected but there was no validation message.

image

I wonder if this should be consistent with question pages?

image

eq-author-api/package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "ajv": "latest",
     "ajv-errors": "latest",
+    "ajv-keywords": "^3.4.1",
```
samiwel: Can we change this to "latest"?

PR #594, opened 2019-08-21
Fixed properties panel scrolling when banner displayed

pricea1:
Properties panel does not scroll to bottom if banner is being displayed.

### How to review
On master: 
1. Create several number/currency answers on page so properties panel is longer than browser window.
2. Trigger the banner e.g. disable network
3. Observe that the last property in panel cannot be fully scrolled to
4. repeat steps in this branch and check last property can be scrolled to

### DIFF
--- a/eq-author/src/components/EditorLayout/index.js
+++ b/eq-author/src/components/EditorLayout/index.js
@@ -30,7 +30,7 @@ const Margin = styled.div`
 const Container = styled.div`
   display: flex;
   flex-direction: column;
-  height: 100vh;
+  height: 100%;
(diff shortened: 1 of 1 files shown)


pricea1:
[Content not scrollable when red banner is present on smaller screens](https://trello.com/c/1yfYV3X5/1288-content-not-scrollable-when-red-banner-is-present-on-smaller-screens)

heofs:
Fixes the issue but it does create an empty space there at the bottom..
Screenshot 2019-08-21 at 16 11 31

PR #614, opened 2019-08-28
Dont push history if at same path

heofs:
Prevent pushing to the same path. Causing lots of warnings in the console.

I am happy for this to be merged if it is approved while I am away.

### How to review 
See that everything still works and warnings are gone on all pages.

### DIFF
--- a/eq-author/src/App/MeContext.js
+++ b/eq-author/src/App/MeContext.js
@@ -25,13 +25,14 @@ const signIn = (setSignInSuccess, history, user) => {
       if (!res.ok) {
         throw Error(`Server responded with a ${res.status} code.`);
       }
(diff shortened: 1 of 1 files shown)


heofs:
Do you have any suggestions for how it can be fixed then? I think at the moment it makes for a really bad DX.

SamGodwin2:
I think the opposite is true, if I click a link and nothing happens because we're already on that page how do I know the link works? The warning here gives that feedback and proves that the correct path has been pushed.

heofs:
I don't disagree that the warning should display in those circumstances. But in our case the console is spammed with the warning... So why keep printing console warnings for a piece of our code thats tested and we know is already working?

PR #626, opened 2019-09-03
Changed default inclusivity on number validation to true

SamGodwin2:
There's a bit of an issue around number validation where due to the way Runner works if a number is set to a previous answer and inclusivity is false then validator will fail as there is a possible range of 0->0 which gives no valid answers. This PR defaults inclusivity to true to hopefully make a better user experience.  

### How to review 
Tests pass, code looks good and the min and max value validation is set to inclusive by default.

### DIFF
--- a/eq-author-api/schema/tests/validation.test.js
+++ b/eq-author-api/schema/tests/validation.test.js
@@ -136,7 +136,7 @@ describe("validation", () => {
         minValue: {
           id: minValueId,
           enabled: false,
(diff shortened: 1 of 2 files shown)


SamGodwin2:
[Numeric answer validation — min value not clear to users (1)](https://trello.com/c/AQc75FbO/1213-numeric-answer-validation-min-value-not-clear-to-users-1)

PR #655, opened 2019-09-23
Added warning for conflicting decimal places in validation refs

SamGodwin2:
Added a warning for when there is a validation rule that references a previous answer and the decimal places do not match.

### How to review 
Check the error appears when the decimals don't match on validation references. And tests pass.


SamGodwin2:
[Validating Decimal Places in Properties Panel (6)](https://trello.com/c/3L5KnIDQ/1177-validating-decimal-places-in-properties-panel-6)

PR #657, opened 2019-09-26
Validating conflicting number currency percentage and unit

zimboflyman:
**### What is the context of this PR?**
When adding a new Question:
We needed to display an error message if any of the following Minimums are GREATER than the Maximum values in the following answer types:

Number:
Percentage
Currency:

**### How to review** 
**1*** - Create a new Question in any section.
Give the Question a title.
Add an answer type -  Number.
Min and Max inputs will appear on the left side-bar.
Enter a min and max value - making sure the Min is GREATER than max.

This should display error message as per Trello attachment.
The error should disappear if either the min or max vals are changed to  make the min val LESS than max...

Repeat **1*** for Currency and Percentage answer types....


zimboflyman:
[Validating conflicting number currency percentage and unit (4)](https://trello.com/c/G6jDb0T5/1179-validating-conflicting-number-currency-percentage-and-unit-4)

PR #667, opened 2019-09-30
Works on max value now

SamGodwin2:
Testing showed that the conflicting decimal validations on previous answer did not correctly work on max value. This PR fixes the issue. 

### How to review 
Set up a conflicting decimal validation on max value previous answer and assert it throws an error back at you.


SamGodwin2:
[Validating Decimal Places in Properties Panel (6)](https://trello.com/c/3L5KnIDQ/1177-validating-decimal-places-in-properties-panel-6)

PR #672, opened 2019-10-03
Min/Max val text reads 'must be (higher/lower) than' when given val is included

K9Morphed:
Link to Trello story: https://trello.com/c/VCD3FsH7/1353-min-max-text-on-numerical-values-3

Modified AnswerValidation to make it so that the 'Min value' and 'Max value' text shows 'must be higher than' and 'must be lower than', respectively, when the given value is specified to be included. When it is not included, the word 'value' is now lowercase (previous uppercase). Tests made and are passing when evaluating every number type (number, percentage, etc.) for both min and max value phrases.

### How to review 
The code should be reviewed on GitHub to make sure it's clean and nice (linter passed locally, yay); run tests to ensure they pass too.

### DIFF
--- a/eq-author/src/App/page/Design/Validation/AnswerValidation.js
+++ b/eq-author/src/App/page/Design/Validation/AnswerValidation.js
@@ -38,6 +38,9 @@ import {
   UNIT,
 } from "constants/answer-types";
 
+export const MIN_INCLUSIVE_TEXT = "must be more than";
+export const MAX_INCLUSIVE_TEXT = "must be less than";
+
 const formatValue = (value, { type, properties }) => {
   if (typeof value !== "number") {
     return null;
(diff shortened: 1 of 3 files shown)


K9Morphed:
Just realised that the logic on this is backwards; the text should show when the value is set to be excluded, not included. I'll make @heofs requested change and fix this today.

K9Morphed:
> Just realised that the logic on this is backwards; the text should show when the value is set to be excluded, not included. I'll make @heofs requested change and fix this today.

Fixed. PR ready to be reviewed.

heofs:
[Min/max text on numerical values (3)](https://trello.com/c/VCD3FsH7/1353-min-max-text-on-numerical-values-3)

PR #673, opened 2019-10-07
Added total piping to variable piping

SamGodwin2:
When we moved to the new content picker a few weeks back the picker responsible for piping the total variable into the calculated summary page got left out. This PR re-adds the functionality on this page type. 

### How to review 
Check you can pipe total into the calculated summary page title and all the tests pass.

### DIFF
--- a/eq-author/src/components/ContentPickerv2/VariablePicker.js
+++ b/eq-author/src/components/ContentPickerv2/VariablePicker.js
@@ -1,0 +1,140 @@
+import React from "react";
+import PropTypes from "prop-types";
+import CustomPropTypes from "custom-prop-types";
+
+import styled from "styled-components";
+
+import { colors } from "constants/theme";
+import {
+  SubMenuItem,
+  MenuItemType,
+  MenuItemTitle,
+  MenuItemSubtitle,
+} from "./Menu";
+import Truncated from "components/Truncated";
+
+import ScrollPane from "components/ScrollPane";
+
+const ModalTitle = styled.div`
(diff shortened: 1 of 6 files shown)


SamGodwin2:
[Filter unit answers on calculated summary page by their unit types (2)](https://trello.com/c/N4BcqKwL/1260-filter-unit-answers-on-calculated-summary-page-by-their-unit-types-2)

eq-author/src/components/ContentPickerv2/VariablePicker.js
```diff
@@ -0,0 +1,140 @@
...
+  width: 100%;
+`;
+
+const RightPositioner = styled.div`
+  float: right;
+`;
+
+const TableHeader = styled.div`
+  display: table-header-group;
+  background: #f4f5f6;
```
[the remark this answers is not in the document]
SamGodwin2: It's a good point, might need to get Joe to make and maintain a theme file with all the colors used in Author. Theres a lot of instances of hard-coded colors in our codebase.

eq-author/src/components/ContentPickerv2/VariablePicker.js
```diff
@@ -0,0 +1,140 @@
...
+  font-size: 11px;
```
heofs: Do you think this would be better as `em` instead of `px`?

SamGodwin2: That was the size in the design and 11px is a little neater than 0.6875em.

PR #674, opened 2019-10-07
Create history section for existing questionnaire functionality

jasehumphr:
We added a History nav item so we can navigate to a new History page. This History page contains history data. Initially we just have the information for when the Questionnaire was created and by whom. History item in the nav is only available for those who have admin access 

### How to review 
Click on the History button in the nav (if you have admin access) view the item. It should show the title of the Questionnaire, State that it was Created, by whom and when.


SamGodwin2:
Screenshot 2019-10-07 at 10 43 45
Dosen't quite match design page should take up the entire page space.

eq-author/src/App/history/HistoryItem.test.js
```diff
@@ -0,0 +1,18 @@
...
+  let props = {
+    questionnaireTitle: "my wonderful thing",
+    userName: "Me",
+    createdAt: "2019-09-26T18:22:58.461Z",
+  };
+
+  it("should format date correctly", () => {
+    const { getByText, debug } = render(<HistoryItem {...props} />);
+
+    debug();
```
heofs: Remove this debug

eq-author/src/App/history/HistoryPage.js
```diff
@@ -0,0 +1,72 @@
...
+    return <Loading height="100%">Questionnaire history loading…</Loading>;
+  }
+
+  if (error) {
+    return <Error>Oops! Something went wrong</Error>;
+  }
+
+  if (data && !loading) {
+    createdAt = data.questionnaire.createdAt;
+
```
heofs: Maybe get rid of this newline?

eq-author/src/App/history/HistoryPage.js
```diff
@@ -0,0 +1,72 @@
...
+  if (loading) {
+    return <Loading height="100%">Questionnaire history loading…</Loading>;
+  }
+
+  if (error) {
+    return <Error>Oops! Something went wrong</Error>;
+  }
+
+  if (data && !loading) {
```
SamGodwin2: Don't think we need this if statement here, if loading and errors are false and data doesn't  we want it to throw as it means something is very wrong somewhere.

Also if we take it out theres no need for the decelerations on 36

eq-author/src/App/history/HistoryPage.js
```diff
@@ -0,0 +1,72 @@
...
+  display: flex;
+  flex-direction: column;
+  padding: 1em;
+  border: 1px solid #d6d8da;
+  margin: 0.5em 1em;
+  overflow: hidden;
+  max-width: 60em;
+`;
+
+export const UnwrappedHistoryPageContent = ({ match }) => {
```
SamGodwin2: Any reason behind calling this `UnwrappedHistoryPageContent` as it doesn't look like there's a WrappedHistoryPageContent anywhere?

eq-author/package.json
```diff
@@ -103,6 +103,7 @@
     "graphql-tools": "latest",
     "jsonwebtoken": "latest",
     "lodash": "latest",
+    "moment": "^2.24.0",
```
SamGodwin2: Can we set this to latest?

eq-author/src/App/history/HistoryPage.js
```diff
@@ -0,0 +1,67 @@
...
+  const { loading, error, data } = useQuery(questionnaireHistoryQuery, {
+    variables: { input: { questionnaireId: match.params.questionnaireId } },
+  });
+
+  if (loading) {
+    return <Loading height="100%">Questionnaire history loading…</Loading>;
+  }
+
+  if (error) {
+    return <Error>Oops! Something went wrong</Error>;
```
K9Morphed: I think the error text needs to be a little more descriptive than this; at the moment it doesn't give any indication as to what went wrong or where. This could make solving potential issues down the line a bit harder.

Could we include any information from the `error` var in the message?

eq-author/src/App/history/HistoryItem.js
```diff
@@ -0,0 +1,48 @@
...
+`;
+
+const formatDate = unformattedDate =>
+  moment(unformattedDate).format("DD/MM/YYYY [at] HH:mm");
+
+const HistoryItem = ({ questionnaireTitle, userName, createdAt }) => {
+  return (
+    <StyledItem>
+      <QuestionnaireTitle>
+        {questionnaireTitle} - <strong>Questionnaire created</strong>
```
K9Morphed: Might be better practice to put `Questionnaire created` in a const value at the top of the page? Makes it easier to find later on if we need to change it and separates business logic from output text.

eq-author/src/App/QuestionnaireDesignPage/NavigationSidebar/NavigationHeader.js
```diff
@@ -116,6 +120,13 @@ export class UnwrappedNavigationHeader extends React.Component {
...
+            {me.admin && (
+              <IconListItem>
+                <RouteButton variant="tertiary-light" small to={historyUrl}>
+                  <IconText icon={HistoryIcon}>History</IconText>
+                </RouteButton>
+              </IconListItem>
+            )}
           </IconList>
         </QuestionnaireLinks>
         <QuestionnaireContent>
```
K9Morphed: If putting the other piece of text into a constant is done, `history` and `Metadata` should be too.

eq-author/src/App/history/HistoryItem.js
jasehumphr: I agree and this is something we need to do throughout the whole app. Normally I've got all output text strings in one file that can be translated and various languages can be used.

K9Morphed: > I agree and this is something we need to do throughout the whole app. Normally I've got all output text strings in one file that can be translated and various languages can be used.

Love that idea, having them all in the same file. Would make it so much easier to control and change. We'll need to remember to mention it at the retro as something to introduce.

K9Morphed: > > I agree and this is something we need to do throughout the whole app. Normally I've got all output text strings in one file that can be translated and various languages can be used.
> 
> Love that idea, having them all in the same file. Would make it so much easier to control and change. We'll need to remember to mention it at the retro as something to introduce.

Made a card for it on the retro board: https://trello.com/c/z0Q39jCv

eq-author/src/App/history/HistoryPage.js
heofs: I agree that it should be a more descriptive message. But this is message we are displaying when an error occurs in the app (at the moment) :/

heofs:
[Create history section for existing questionnaire functionality (4)](https://trello.com/c/HFZHI8zN/1313-create-history-section-for-existing-questionnaire-functionality-4)

PR #687, opened 2019-10-16
Add a note to the history page

SamGodwin2:
This PR allows the adding of notes to the questionnaire history page.

### How to review 
Tests pass and you can create notes as expected on the history page for any questionnaire.


heofs:
[Add a note to the History page (6)](https://trello.com/c/7CeQ2jgb/1314-add-a-note-to-the-history-page-6)

heofs:
I get this prop type error when adding a note
Screenshot 2019-10-18 at 09 37 29

eq-author/src/App/history/HistoryPage.test.js
```diff
@@ -8,14 +8,27 @@ import { MeContext } from "App/MeContext";
...
 import questionnaireHistoryQuery from "./questionnaireHistory.graphql";
+import createNoteMutation from "./createNoteMutation.graphql";
+
+//eslint-disable-next-line react/prop-types
+jest.mock("components/RichTextEditor", () => ({ onUpdate }) => {
+  const handleInputChange = event => {
+    onUpdate({
+      value: event.target.value,
+    });
+  };
```
heofs: This could be an implicit return

eq-author/src/App/history/createNoteMutation.graphql
```diff
@@ -0,0 +1,15 @@
...
+  createNote(input: $input) {
+    id
+    publishStatus
+    questionnaireTitle
+    bodyText
+    user {
+      id
+      email
+      name
+      displayName
```
heofs: Is it enough to query only one of `name` and `displayName`? Is email needed here at all?

SamGodwin2: Technically it is but if you don't query everything Apollo will whine as the different user objects in the cache won't quite match up.

heofs:
Code looks good now, but I do get this `w` character before the title in the notes.
Screenshot 2019-10-23 at 10 53 02

Also, if you, for example, create a note with multiple points, the `Add Note` button disappears, and you can not scroll it into view.

PR #694, opened 2019-10-18
Validations fix in staging

SamGodwin2:
Hopefully should fix our staging smoke tests.

### How to review 
Spin this change up then run the cypress smoke tests 20-30 time against it. If you get any fails let me know!


eq-author/src/App/page/Design/Validation/AnswerValidation.js
```diff
@@ -228,49 +225,4 @@ UnwrappedAnswerValidation.propTypes = {
...
-          }
-        }
-      }
-    }
-  }
-  ${ValidationErrorInfo}
-`;
-
-export default withValidations(UnwrappedAnswerValidation, VALIDATION_QUERY);
+export default UnwrappedAnswerValidation;
```
heofs: Should this be called `AnswerValidation` instead? As it is no longer being wrapped.

PR #709, opened 2019-10-28
Authors can now see the History page

MattElcock:
I removed the boolean checking to see if a user is an admin (`{ me.admin && ... }`) so that the History page would be displayed to everyone.

https://trello.com/c/mEWcK8pH/1360-make-history-page-visible-to-authors

### How to review 
Have a look at the code and see what you think; it shouldn't take long. Ensure the tests pass and you can view the History page both as an admin and a normal author.


MattElcock:
> All good but I thought we were going to leave this until the edit & delete PR is done.

I'll check at the stand up before merging.

jasehumphr:
> > All good but I thought we were going to leave this until the edit & delete PR is done.
> 
> I'll check at the stand up before merging.

Make History Great Again

MattElcock:
> Does this mean we don't need to wrap it in the with me provider any more?

Yeah; I've actioned this and removed the `me` proptype.

PR #713, opened 2019-10-30
Reconfigured the RTE boxes

SamGodwin2:
Just reconfigures the settings of the RTE's on the question page.

### How to review 
All new functionality works and makes its way to Publisher.


eq-author/src/App/page/Design/QuestionPageEditor/MetaEditor.js
```diff
@@ -27,16 +27,18 @@ const titleControls = {
 };
 
 const descriptionControls = {
-  bold: true,
   emphasis: true,
   piping: true,
 };
```
MattElcock: Is this needed now?

PR #714, opened 2019-11-01
Author now understands multiple versions of the same questionnaire

MattElcock:
I have changes the API so that:

- Questionnaires are built with an initial version of 1
- TriggerPublish sends the version number of a questionnaire to the registry for storage
- createMutation increments a new version number IF the publish status was set to 'Published'
- History page will update when a questionnaire has been published and unpublished
- History items now show the version of the questionnaire at the time of the history item being added

I have also removed some remnants of discarded code relating to publishing a questionnaire

Describe what you have changed and why, link to other PRs or Issues as appropriate.

### How to review 
Test alongside https://github.com/ONSdigital/eq-survey-register/pull/4; ensure all tests pass and that you can publish multiple versions. Ensure that the history page updates as per the designs.

### DIFF
--- a/eq-author-api/schema/tests/questionnaire.test.js
+++ b/eq-author-api/schema/tests/questionnaire.test.js
@@ -32,6 +32,12 @@ const {
   createHistoryNote,
 } = require("../../tests/utils/contextBuilder/questionnaire");
 
+const {
+  createAnswer,
+  updateAnswer,
+} = require("../../tests/utils/contextBuilder/answer");
+const { NUMBER } = require("../../constants/answerTypes");
+
 const defaultUser = require("../../tests/utils/mockUserPayload");
 const { createUser } = require("../../utils/datastore");
 
@@ -39,11 +45,13 @@ describe("questionnaire", () => {
   let ctx, questionnaire;
   const surveyId = "123";
   const formType = "321";
+  const surveyVersion = "1";
 
   afterEach(async () => {
     if (!questionnaire) {
       return;
     }
+
     await deleteQuestionnaire(questionnaire.id);
     questionnaire = null;
   });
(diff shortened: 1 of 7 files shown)


MattElcock:
https://trello.com/c/INgFJmI1/1318-versioning-in-register-4

eq-author-api/schema/tests/basicAnswer.test.js
```diff
@@ -398,4 +398,54 @@ describe("basic answer", () => {
       expect(page.id).toEqual(questionnaire.sections[0].pages[0].id);
     });
   });
+
+  describe("Versioning", () => {
```
heofs: I think this test should be in `questionnaire.test.js`. It doesnt seem very specific to a `basicAnswer`.

eq-author-api/schema/tests/questionnaire.test.js
```diff
@@ -410,20 +411,30 @@ describe("questionnaire", () => {
...
   });
+
+  describe("versioning", () => {
+    ctx = buildContext(null);
+
+    it("should initialise new questionnaires at version 1", () => {
+      expect(ctx.questionnaire.surveyVersion).toEqual("1");
+    });
+
+    it("can increment the version number of a questionnaire if a published survey is modified and becomes unpublished", () => {});
```
heofs: This test is not doing anything

eq-author-api/schema/tests/basicAnswer.test.js
MattElcock: That makes sense, I've moved it across.

eq-author-api/schema/tests/questionnaire.test.js
MattElcock: Ah I forgot about that, haha. Removed and replaced!

PR #716, opened 2019-11-07
Enable publish button when questionnaire becomes unpublished

heofs:
Fixed an issue where the publish button was still disabled after the questionnaire went from `Published` to `Unpublished`.

This is done by adding a subscription that sends the updated `publishStatus` if the status goes from `Published` to `Unpublished`. 

This subscription was added as a hook to the header component and was causing a lot of tests to break.

### How to review 
Create a new questionnaire and publish it. Make sure publish status on the front page is now `Published`. Edit the questionnaire and notice that the publish button automatically is enabled again.

### DIFF
--- a/eq-author-api/schema/tests/subscription.test.js
+++ b/eq-author-api/schema/tests/subscription.test.js
@@ -1,11 +1,17 @@
 const gql = require("graphql-tag");
 
 const { NUMBER } = require("../../constants/answerTypes");
+const {
+  PUBLISHED,
+  UNPUBLISHED,
+  AWAITING_APPROVAL,
+} = require("../../constants/publishStatus");
 
 const { buildContext } = require("../../tests/utils/contextBuilder");
 const executeSubscription = require("../../tests/utils/executeSubscription");
 
 const {
+  updateQuestionnaire,
   deleteQuestionnaire,
 } = require("../../tests/utils/contextBuilder/questionnaire");
 const {
@@ -14,6 +20,18 @@ const {
 
 const wait = timeout =>
   new Promise(resolve => setTimeout(() => resolve("timeout"), timeout));
+
+jest.mock("node-fetch");
+const fetch = require("node-fetch");
+
+fetch.mockImplementation(() =>
+  Promise.resolve({
+    json: () => ({
+      questionnaireId: "test",
+      publishedSurveyUrl: "https://best.url.com",
+    }),
+  })
+);
 
 describe("subscriptions", () => {
   let ctx, questionnaire, iterator;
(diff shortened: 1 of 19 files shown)


heofs:
[Approve content](https://trello.com/c/ExJa5ANr/1199-approve-content)

eq-author-api/schema/tests/subscription.test.js
```diff
@@ -157,6 +175,64 @@ describe("subscriptions", () => {
         title: "",
       });
 
+      const p = iterator.next();
```
MattElcock: Extremely small thing, feel free to ignore, but could this variable be a little more descriptive? Personally, I'm not a big fan of one letter variables.

eq-author-api/schema/tests/subscription.test.js
```diff
@@ -157,6 +175,64 @@ describe("subscriptions", () => {
...
+      iterator = await executeSubscription(publishStatusSubscription, {
+        id: questionnaire.id,
+      });
+
+      await updateQuestionnaire(ctx, {
+        id: questionnaire.id,
+        publishStatus: AWAITING_APPROVAL,
+      });
+
       const p = iterator.next();
```
MattElcock: If the variable above is changed, maybe do this one at the same time.

heofs: I agree, one letter vars is not nice... But tbh I dont know what to call this, not sure what it does really. Just copied it from the above tests. I could call it something like `nextIterator` or `iteratorNext`?

heofs: I have renamed to `iteratorNext`

eq-author-api/schema/tests/subscription.test.js
heofs: I have renamed to `iteratorNext`

eq-author-api/schema/tests/subscription.test.js
```diff
@@ -120,8 +138,8 @@ describe("subscriptions", () => {
         id: pageId,
         title: "",
       });
-      const p = iterator.next();
-      const result = await Promise.race([p, wait(50)]);
+      const iteratorNext = iterator.next();
```
SamGodwin2: p is fine here, it's pretty accepted shorthand for a promise, if we want to be more descriptive a name with the word promise in it would work.

heofs: I will change it to `promise` 👍

eq-author-api/schema/typeDefs.js
```diff
@@ -735,6 +735,7 @@ input CreateQuestionnaireInput {
   type: QuestionnaireType!
   shortTitle: String
   isPublic: Boolean
+  publishStatus: String
```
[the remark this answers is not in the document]
heofs: contextBuilder needs publishStatus on CreateQuestionnaireInput. Removing any of those results in subscription tests failing.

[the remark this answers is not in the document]
SamGodwin2: Ugh don't like that. I thought the context builder fixes works now by triggering publish mutations, we really shouldn't be changing our schema to get tests to pass.

PR #722, opened 2019-11-13
1332 edit and delete a note

jasehumphr:
Allows a user to delete or edit their own history comments. Also allows admin accounts to edit and delete other users comments.

### How to review 
Go to History page and make a comment with the existing Rich Text Editor. There will be a new message that if permissions allow will show a bar with the edit and delete icons. Clicking the edit icon will open up a RTE and then on saving via the save button the message will be changed. The delete icon "X" will delete this message. You will need to login both as an admin and non admin to test

### DIFF
--- a/eq-author-api/migrations/addTypeToHistoryEvent.js
+++ b/eq-author-api/migrations/addTypeToHistoryEvent.js
@@ -1,0 +1,17 @@
+const { getQuestionnaireMetaById, saveModel } = require("../utils/datastore");
+const { QuestionnaireModel } = require("../db/models/DynamoDB");
+
+module.exports = async function addTypeToHistoryEvent(questionnaire) {
+  const metadata = await getQuestionnaireMetaById(questionnaire.id);
+
+  metadata.history.map(item => {
+    if (!item.type) {
+      item.type = item.hasOwnProperty("bodyText") ? "note" : "system";
+    }
+    return item;
+  });
+
+  await saveModel(new QuestionnaireModel(metadata));
+
+  return questionnaire;
+};
--- a/eq-author-api/schema/resolvers/base.js
+++ b/eq-author-api/schema/resolvers/base.js
@@ -79,10 +79,13 @@ const {
   getQuestionnaire,
   getUserById,
   listUsers,
-  addEventToHistory,
-  getHistoryById,
+  createHistoryEvent,
+  getQuestionnaireMetaById,
   createComments,
+  saveModel,
 } = require("../../utils/datastore");
+
+const { QuestionnaireModel } = require("../../db/models/DynamoDB");
 
 const {
   createDefaultBusinessSurveyMetadata,
@@ -162,7 +165,10 @@ const Resolvers = {
       });
     },
     questionnaire: (root, args, ctx) => ctx.questionnaire,
-    history: (root, { input }) => getHistoryById(input.questionnaireId),
+    history: async (root, { input }) =>
+      getQuestionnaireMetaById(input.questionnaireId).then(
+        ({ history }) => history
+      ),
     section: (root, { input }, ctx) => getSectionById(ctx, input.sectionId),
     page: (root, { input }, ctx) => getPageById(ctx, input.pageId),
     answer: (root, { input }, ctx) => getAnswerById(ctx, input.answerId),
@@ -254,7 +260,46 @@ const Resolvers = {
       return createQuestionnaire(newQuestionnaire, ctx);
     },
     createHistoryNote: (root, { input }, ctx) =>
-      addEventToHistory(input.id, noteCreationEvent(ctx, input.bodyText)),
+      createHistoryEvent(input.id, noteCreationEvent(ctx, input.bodyText)),
+
+    updateHistoryNote: async (root, { input }, ctx) => {
+      const user = ctx.user;
+      const metadata = await getQuestionnaireMetaById(input.questionnaireId);
+
+      const noteToUpdate = metadata.history.find(({ id }) => id === input.id);
+
+      if (!user.admin && user.id !== noteToUpdate.userId) {
+        throw new Error("User doesnt have access");
+      }
+      if (noteToUpdate.type === "system") {
+        throw new Error("Cannot update system event message");
+      }
+
+      noteToUpdate.bodyText = input.bodyText;
+
+      await saveModel(new QuestionnaireModel(metadata));
+      return metadata.history;
+    },
+
+    deleteHistoryNote: async (root, { input }, ctx) => {
+      const user = ctx.user;
+      const metadata = await getQuestionnaireMetaById(input.questionnaireId);
+      const history = metadata.history;
+      const noteToDelete = history.find(item => item.id === input.id);
+
+      if (!user.admin && user.id !== noteToDelete.userId) {
+        throw new Error("User doesnt have access");
+      }
+      if (noteToDelete.type === "system") {
+        throw new Error("Cannot delete system event message");
+      }
+
+      remove(metadata.history, item => item.id === noteToDelete.id);
+
+      await saveModel(new QuestionnaireModel(metadata));
+      return metadata.history;
+    },
+
     createSection: createMutation((root, { input }, ctx) => {
       const section = createSection(input);
       ctx.questionnaire.sections.push(section);
@@ -598,7 +643,7 @@ const Resolvers = {
             throw Error(e);
           });
 
-        await addEventToHistory(ctx.questionnaire.id, publishStatusEvent(ctx));
+        await createHistoryEvent(ctx.questionnaire.id, publishStatusEvent(ctx));
 
         ctx.questionnaire.publishStatus = PUBLISHED;
       }
@@ -610,7 +655,7 @@ const Resolvers = {
 
         ctx.questionnaire.publishStatus = UPDATES_REQUIRED;
 
-        await addEventToHistory(
+        await createHistoryEvent(
           ctx.questionnaire.id,
           publishStatusEvent(ctx, input.reviewComment)
         );
--- a/eq-author-api/schema/typeDefs.js
+++ b/eq-author-api/schema/typeDefs.js
@@ -64,12 +64,17 @@ type Questionnaire {
   publishDetails: PublishDetails
   totalErrorCount: Int!
 }
+enum HistoryEventTypes {
+  system
+  note
+}
 
 type History {
   id: ID!
   publishStatus: String!
   questionnaireTitle: String!
   bodyText: String
+  type: HistoryEventTypes!
   user: User!
   time: DateTime!
 }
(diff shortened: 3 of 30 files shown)


eq-author-api/schema/resolvers/base.js
```diff
@@ -246,6 +248,39 @@ const Resolvers = {
...
+
+    updateHistoryNote: async (root, { input }, ctx) => {
+      const user = ctx.user;
+      const historyItems = await getHistoryById(input.questionnaireId);
+
+      const noteToUpdate = historyItems.find(item => item.id === input.id);
+
+      if (!user.admin && user.id !== noteToUpdate.userId) {
+        throw new Error("User doesnt have access");
+      }
```
MattElcock: I think this should be an OR condition; otherwise admins can't update other peoples notes as it'll evaluate to false when they're trying to do so on other peoples.

eq-author-api/schema/resolvers/base.js
```diff
@@ -246,6 +248,39 @@ const Resolvers = {
...
+      );
+    },
+    deleteHistoryNote: async (root, { input }, ctx) => {
+      const user = ctx.user;
+      const historyItems = await getHistoryById(input.questionnaireId);
+      const noteToDelete = historyItems.find(item => item.id === input.id);
+
+      if (!user.admin && user.id !== noteToDelete.userId) {
+        throw new Error("User doesnt have access");
+      }
```
MattElcock: Same as above, I think this should be an OR condition.

eq-author-api/schema/typeDefs.js
```diff
@@ -70,6 +70,7 @@ type History {
   publishStatus: String!
   questionnaireTitle: String!
   bodyText: String
+  type: String
```
MattElcock: Should this be required instead of optional?

eq-author/src/App/history/HistoryItem.test.js
```diff
@@ -4,10 +4,21 @@ import HistoryItem from "./HistoryItem";
 
 describe("History page", () => {
   let props = {
+    questionnaireId: "123456789",
+    itemId: "987654321",
+    userId: "432156789",
+    currentUser: {
+      id: "123",
+      displayName: "Rick Sanchez",
+      email: "wubbalubba@dubdub.com",
```
MattElcock: Nice.

jasehumphr: That will be @SamGodwin2

eq-author-api/utils/datastore.js
```diff
@@ -140,6 +140,40 @@ const addEventToHistory = (questionnaireId, historyEvent) =>
...
+        if (err) {
+          reject(err);
+        }
+        remove(questionnaire.history, history => history.id === eventId);
+
+        await saveModel(new QuestionnaireModel(questionnaire));
+        resolve(questionnaire.history);
+      }
+    );
+  });
```
SamGodwin2: If you mutate the history objects inside the resolvers then pass that to the save model we should be able to delete both these functions.

eq-author/src/App/history/HistoryItem.js
```diff
@@ -1,20 +1,50 @@
...
+  padding: 0 2em 0 0;
   border: 1px solid ${colors.lightGrey};
   background-color: ${colors.lighterGrey};
   width: 100%;
-  margin: 0.5em 1em 0 0;
+  margin: 0;
   justify-content: space-between;
 `;
 
+const EditThis = styled.div`
```
SamGodwin2:
```suggestion
const EditHeader = styled.div`
```
?

eq-author/src/App/history/HistoryItem.js
```diff
@@ -37,27 +75,179 @@ const EventText = styled.div`
...
 
+const HistoryButtonGroup = styled.div`
+  display: flex;
+  flex-direction: column;
+  flex-direction: row;
+  justify-content: flex-end;
+
+  :not(:last-child) {
+    margin-right: -1.5em !important;
+  }
```
SamGodwin2: Whats going on here, do we need anything more than 
```
  display: flex;
  justify-content: flex-end;
```

eq-author/src/App/history/HistoryItem.js
```diff
@@ -37,27 +75,179 @@ const EventText = styled.div`
...
-    </StyledItem>
+    <EditableItem>
+      {(currentUser.id === userId || currentUser.admin) &&
+        (type === "note" && (
+          <EditThis>
+            <HistoryButtonGroup horizontal align="right">
+              <EditButtonIcon onClick={handleEdit} />
+              <HistoryDeleteButton onClick={() => handleDelete()} />
+            </HistoryButtonGroup>
+          </EditThis>
```
SamGodwin2: This doesn't match designs, bar should be grey when inactive, blue only when selected.

eq-author/src/App/history/HistoryItem.js
```diff
@@ -37,27 +75,179 @@ const EventText = styled.div`
...
+  }
+`;
+
+const SaveButton = styled(Button)`
+  font-size: 0.75em;
+  margin-right: -1em;
+  margin-bottom: 1em;
+`;
+
+const EditButtonIcon = styled.button`
```
SamGodwin2: We already have a component for this, `IconText` should do the trick.

eq-author/src/App/history/HistoryItem.js
```diff
@@ -37,27 +75,179 @@ const EventText = styled.div`
...
+                  list: true,
+                  bold: true,
+                }}
+                multiline
+              />
+            </RTEWrapper>
+            <ActionButtons horizontal align="right">
+              <SaveButton onClick={handleSave}>Save</SaveButton>
+            </ActionButtons>
+          </StyledGrid>
```
SamGodwin2: Can we extract this into a function, its a little unclear whats going on here.

eq-author-api/schema/typeDefs.js
SamGodwin2: Should also be an enum of its available options.

eq-author/src/App/history/HistoryItem.js
```diff
@@ -37,27 +75,179 @@ const EventText = styled.div`
...
+  const handleDelete = () => {
+    deleteNote({
+      variables: {
+        input: {
+          id: itemId,
+          questionnaireId,
+        },
+      },
+    });
+  };
```
SamGodwin2: Might be nice to animate this if possible, looks a little clunky atm, look at how answers are deleted as an example.

eq-author/src/App/history/HistoryItem.js
```diff
@@ -37,27 +75,179 @@ const EventText = styled.div`
...
-      </QuestionnaireUserName>
-      <BreakLine />
-      {bodyText && <EventText dangerouslySetInnerHTML={{ __html: bodyText }} />}
-    </StyledItem>
+    <EditableItem>
+      {(currentUser.id === userId || currentUser.admin) &&
+        (type === "note" && (
+          <EditThis>
+            <HistoryButtonGroup horizontal align="right">
+              <EditButtonIcon onClick={handleEdit} />
```
heofs:
```suggestion
              <EditButtonIcon onClick={() => setActiveEdit(true))} />
```
And you can remove the function.

eq-author/src/App/history/HistoryItem.js
```diff
@@ -37,27 +75,179 @@ const EventText = styled.div`
...
+              <HistoryDeleteButton onClick={() => handleDelete()} />
```
heofs:
```suggestion
              <HistoryDeleteButton onClick={handleDelete} />
```

eq-author/src/App/history/HistoryPage.js
```diff
@@ -4,6 +4,8 @@ import { useQuery, useMutation } from "@apollo/react-hooks";
 
 import CustomPropTypes from "custom-prop-types";
 
+import { useMe } from "App/MeContext";
```
heofs: can we move this import to right below `useQuery...`

eq-author/src/App/history/deleteHistoryNoteMutation.graphql
```diff
@@ -0,0 +1,8 @@
+#import "./historyNoteFragment.graphql"
+
+mutation DeleteHistoryNote($input: deleteHistoryNoteInput!) {
+  deleteHistoryNote(input: $input) {
+    id
```
heofs: Why is `id` needed here? I think we are getting it of the `HistoryNoteInfo` fragment.

eq-author-api/migrations/addTypeToHistoryEvent.js
```diff
@@ -0,0 +1,11 @@
+//This is an auto-generated file.  Do NOT modify the method signature.
+
+module.exports = function addTypeToHistoryEvent(questionnaire) {
+  questionnaire.history.map(item => {
+    if (!item.type) {
+      item.type = item.hasOwnProperty("bodyText") ? "note" : "system";
+    }
+    return item;
+  });
+  return questionnaire;
```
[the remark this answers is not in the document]
SamGodwin2: Good point 😳

eq-author-api/schema/resolvers/base.js
```diff
@@ -246,6 +248,39 @@ const Resolvers = {
...
+      if (!noteToDelete.hasOwnProperty("bodyText")) {
+        throw new Error("Cannot delete system event message");
+      }
```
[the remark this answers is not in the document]
jasehumphr: Yes missed this one

eq-author-api/schema/resolvers/base.js
heofs: It needs to be `&&` here. If we use and it basically means, `if you are not admin OR you are not the note owner THEN throw the error`. But we want it to throw only if both conditions are true.

jasehumphr:
> WE ARE NOTE (Looks good to me)

I HAS TICK

eq-author-api/migrations/addTypeToHistoryEvent.js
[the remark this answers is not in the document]
jasehumphr: Works now

eq-author/src/App/history/HistoryItem.js
```diff
@@ -45,37 +76,130 @@ const translations = {
...
-      </QuestionnaireTitle>
-      <QuestionnaireUserName>
-        {userName} - {formatDate(createdAt)}
-      </QuestionnaireUserName>
-      <BreakLine />
-      {bodyText && <EventText dangerouslySetInnerHTML={{ __html: bodyText }} />}
-    </StyledItem>
+    <EditableItem>
+      {(currentUser.id === userId || currentUser.admin) &&
+        (type === "note" && (
```
SamGodwin2: Can we extract this logic into a `isEditable` variable the conditional rendering here is a little tricky to follow.

PR #735, opened 2019-11-19
Fix scrollpane ghosting text issue

heofs:
Fixes issue with text ghosting when using scrollpane.

### How to review 
Check that nothing has changed except for ghosting is now gone. You can test this on the history page by adding multiple notes till the page starts scrolling.


heofs:
[Fix scrollpane ghosting text issue](https://trello.com/c/UOr5f1tn/1402-fix-scrollpane-ghosting-text-issue)

PR #774, opened 2019-12-05
Upgrading styled components

SamGodwin2:
Updated styled components and rewrote a test or two. 

### How to review 
Give the entire app a once over, make sure all the styles are the same as before and nothings messed up


SamGodwin2:
[Upgrade react to 16.11](https://trello.com/c/C4NygDpL/1282-upgrade-react-to-1611)

heofs:
Screenshot 2019-12-05 at 11 40 39
😭

eq-author/src/App/QuestionnairesPage/QuestionnairesView/QuestionnairesTable/Row/index.js
```diff
@@ -8,7 +8,6 @@ import CustomPropTypes from "custom-prop-types";
 
 import IconButtonDelete from "components/buttons/IconButtonDelete";
 import DuplicateButton from "components/buttons/DuplicateButton";
-import FadeTransition from "components/transitions/FadeTransition";
```
heofs: Why no more fade transition on deleting questionnaire?

SamGodwin2: Well spotted 😉 I couldn't get it to work in a reasonable way, there was a slightly breaking change and getting it to work was starting to take longer than was worth it, especially as you never really see it due to the modal being in the way.

heofs: Okay thats fine. I think we should have a session where we decide what animations etc we should have, as I dont think we have any consistent use of them. It also needs design work to decide what the animation should do.

PR #775, opened 2019-12-05
403 properties max character length

jchapman68:
Added text properties box to set maximum characters for Textfield and Textarea answers.  The value should be between 10 and 2000 inclusive.  Properties box border turns red when navigating away and an error messages will show if value is outside this range.

### How to review 
Paired with Matthew Elcock
Run `yarn test` for both author and api

### Trello ticket
https://trello.com/c/Ry2LUQ1h


eq-author/src/App/page/PropertiesPanel/GroupedAnswerProperties/AnswerProperties/Properties/updateAnswerMutation.graphql
```diff
@@ -0,0 +1,21 @@
...
+      validationErrorInfo {
+        id
+        errors {
+          id
+          type
+          field
+          errorCode
+        }
+        totalCount
+      }
```
jasehumphr: We have a fragment for this

eq-author-api/src/validation/index.test.js
```diff
@@ -269,6 +271,124 @@ describe("schema validation", () => {
...
+                          type: TYPE,
+                          label: "Desc",
+                          properties: { maxLength: "9" },
+                        },
+                      ],
+                    },
+                  ],
+                },
+              ],
+            };
```
heofs: This questionnaire object is set almost the same way on every test. So it could be moved into a before each, and then just mutate the object the way you want it in every test.

eq-author-api/src/validation/index.test.js
```diff
@@ -269,6 +271,124 @@ describe("schema validation", () => {
       });
     });
 
+    describe("textarea and textfield answers", () => {
+      [TEXTAREA, TEXTFIELD].forEach(TYPE => {
```
heofs:
```suggestion
      beforeEach(() => {
        questionnaire = {
          ...
        }
      })
      [TEXTAREA, TEXTFIELD].forEach(TYPE => {
```

eq-author/src/App/page/PropertiesPanel/GroupedAnswerProperties/AnswerProperties/Properties/TextProperties.test.js
```diff
@@ -0,0 +1,113 @@
...
+                __typename: "BasicAnswer",
+              },
+            },
+          };
+        },
+      },
+    ];
+  });
+
+  it("on first render default value is 2000", () => {
```
heofs: I think the pattern we follow for writing test text is something like 
```suggestion
  it("should render default value as 2000", () => {
```

eq-author/src/App/page/PropertiesPanel/GroupedAnswerProperties/AnswerProperties/Properties/TextProperties.test.js
```diff
@@ -0,0 +1,113 @@
...
+    const { getByTestId } = renderTextProperties(props);
+    const inputBox = getByTestId("maxCharacterInput");
+    expect(inputBox.value).toBe("2000");
+  });
+
+  it("when input box is cleared value should return to default of 2000", async () => {
```
heofs:
```suggestion
  it("should set input box value to 2000 when cleared", async () => {
```
Same for other tests here

eq-author/src/App/page/PropertiesPanel/GroupedAnswerProperties/index.js
```diff
@@ -42,6 +45,7 @@ const AnswerPropertiesContainer = styled.div`
 const ValidationWarning = styled(IconText)`
   color: ${colors.red};
   margin-top: 0.5em;
+  justify-content: normal;
```
heofs: `justify-content` has to be given to an element that has `display: flex`
```suggestion
```

heofs:
This doesnt seem to make a difference when running the questionnaire though. Max length seems to always be 2000.

eq-author-api/src/validation/customKeywords/textLengthInRange.js
```diff
@@ -0,0 +1,35 @@
+module.exports = function(ajv) {
+  ajv.addKeyword("textLengthInRange", {
```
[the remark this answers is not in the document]
SamGodwin2: I had a quick look but it might be a bit of a pain. The issue is Graphql stringifys JSON types so we can't use numeric validation on it and I couldn't find a non-messy place to do a type coercion.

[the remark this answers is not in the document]
heofs: Might this help? https://ajv.js.org/coercion.html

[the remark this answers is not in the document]
heofs: https://github.com/epoberezkin/ajv#coercing-data-types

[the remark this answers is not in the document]
SamGodwin2: Yeah I just wasn't a fan of the validator modifying the underlying data but it's not too problematic.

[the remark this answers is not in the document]
heofs: Yeah, I agree. We can just stick with this keyword then 👍

MattElcock:
> This doesnt seem to make a difference when running the questionnaire though. Max length seems to always be 2000.

Fixed this; I forgot to do the Publisher changes with Josh, aha.

eq-author/src/App/page/PropertiesPanel/GroupedAnswerProperties/index.test.js
```diff
@@ -301,4 +307,104 @@ describe("Grouped Answer Properties", () => {
       );
     });
   });
+
+  describe("Text answers", () => {
+    const ERR_MAX_LENGTH_TOO_LARGE = "ERR_MAX_LENGTH_TOO_LARGE";
+    const ERR_MAX_LENGTH_TOO_SMALL = "ERR_MAX_LENGTH_TOO_SMALL";
+
+    it(`Should render 'Enter a character less than x' error on ${TEXTAREA} answers`, () => {
```
heofs: could just write textarea instead of using a constant here. I think you would normally use a constant in the description for when you are looping through several tests.

heofs:
If you create two textboxes on one question page. You are only able to set the max length of the first box and not the second. Second one default to 2000.
Screenshot 2019-12-20 at 15 22 23
Screenshot 2019-12-20 at 15 22 29

MattElcock:
> If you create two textboxes on one question page. You are only able to set the max length of the first box and not the second. Second one default to 2000.
> Screenshot 2019-12-20 at 15 22 23
> Screenshot 2019-12-20 at 15 22 29

With this one, I had a look over Christmas and I think the mutation that's used to update the maxLength thingy should be `updateAnswersOfType` instead of `updateAnswer` -- not tested that though so not 100% sure.

MattElcock:
> If you create two textboxes on one question page. You are only able to set the max length of the first box and not the second. Second one default to 2000.
> Screenshot 2019-12-20 at 15 22 23
> Screenshot 2019-12-20 at 15 22 29

Fixed

eq-author-api/src/validation/index.test.js
```diff
@@ -515,10 +588,11 @@ describe("schema validation", () => {
       const section = questionnaire.sections[0];
       section.title = "";
 
-      const validationErrors = validation(questionnaire);
-      const sectionErrors = validationErrors.sections[section.id].errors;
-      expect(sectionErrors).toHaveLength(1);
-      expect(sectionErrors[0]).toMatchObject({
+      const validationpageErrors = validation(questionnaire);
+      const sectionpageErrors =
+        validationpageErrors.sections[section.id].errors;
```
jasehumphr: validationpageErrors & sectionpageErrors look really odd.
The lack of proper camelcase disturbs me
How about? validationPageErrors & sectionPageErrors

Obviously not a show stopper

eq-author-api/src/validation/index.test.js
```diff
@@ -114,18 +119,18 @@ describe("schema validation", () => {
...
+        validationpageErrors.confirmation[confirmationId].errors[0]
       ).toMatchObject({
         errorCode: "ERR_VALID_REQUIRED",
         field: "label",
         id: "positive-positive-label",
         type: "confirmationoption",
       });
       expect(
-        validationErrors.confirmationoption.positive.errors[0]
+        validationpageErrors.confirmationoption.positive.errors[0]
```
jasehumphr: This looks odd as well. confirmationoption

Obviously it was there before

eq-author-api/src/validation/index.test.js
```diff
@@ -155,9 +160,9 @@ describe("schema validation", () => {
       };
       questionnaire.sections[0].pages[0].answers = [answer];
 
-      const validationErrors = validation(questionnaire);
+      const validationpageErrors = validation(questionnaire);
```
jasehumphr: This const looks odd. Camel validationpageErrors to validationPageErrors

eq-author-api/src/validation/index.test.js
jchapman68: Sorted

PR #790, opened 2019-12-10
Fixes bug on metadata page

SamGodwin2:
Due to one of the recent updates that fixed one of our dependancies the metadata page broke. This PR fixes the bug.

### How to review 
Metadata page works again.

### DIFF
--- a/eq-author/src/App/metadata/MetadataPage.js
+++ b/eq-author/src/App/metadata/MetadataPage.js
@@ -46,7 +46,7 @@ const StyledMainCanvas = styled(MainCanvas)`
 export const UnwrappedMetadataPageContent = ({
   loading,
   error,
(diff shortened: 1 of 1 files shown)


eq-author/src/App/metadata/MetadataPage.js
```diff
@@ -46,15 +46,15 @@ const StyledMainCanvas = styled(MainCanvas)`
...
+  data,
   onAddMetadata,
   onDeleteMetadata,
   onUpdateMetadata,
 }) => {
   if (loading) {
     return <Loading height="100%">Questionnaire metadata loading…</Loading>;
   }
-
+  const { questionnaire } = data;
```
heofs: I think this should be below the error check

PR #823, opened 2019-12-19
Enable publish for editors and owners

heofs:
Enabled publish button for editors and owners of questionnaires.

### How to review 
Check that you can now publish a questionnaire as an owner and editor.

### DIFF
--- a/eq-author/src/App/publish/PublishPage.js
+++ b/eq-author/src/App/publish/PublishPage.js
@@ -6,7 +6,6 @@ import CustomPropTypes from "custom-prop-types";
 import { useMutation } from "@apollo/react-hooks";
 import { map, isEmpty, some } from "lodash";
 
-import { useMe } from "App/MeContext";
 import { useQuestionnaire } from "components/QuestionnaireContext";
 
 import { colors } from "constants/theme";
@@ -94,7 +93,6 @@ const PublishPage = ({ match, history }) => {
   const questionnaireId = match.params.questionnaireId;
   const [surveyId, setSurveyId] = useState("");
   const [formTypes, setFormTypes] = useState({});
-  const { me } = useMe();
   const { questionnaire } = useQuestionnaire();
 
   const [triggerPublish] = useMutation(triggerPublishMutation);
@@ -113,10 +111,17 @@ const PublishPage = ({ match, history }) => {
   };
 
   const publishStatus = questionnaire && questionnaire.publishStatus;
+
+  const canPublish = () => {
+    if (!questionnaire) {
+      return true;
+    }
+    return questionnaire.permission === "Write";
+  };
   if (
-    !me.admin ||
     publishStatus === AWAITING_APPROVAL ||
-    publishStatus === PUBLISHED
+    publishStatus === PUBLISHED ||
+    !canPublish()
   ) {
     return <Redirect to={`/q/${match.params.questionnaireId}`} />;
   }
--- a/eq-author/src/App/publish/PublishPage.test.js
+++ b/eq-author/src/App/publish/PublishPage.test.js
@@ -20,8 +20,8 @@ describe("Publish page", () => {
         </QuestionnaireContext.Provider>
       </MeContext.Provider>,
       {
-        route: `/q/${questionnaire.id}/page/2`,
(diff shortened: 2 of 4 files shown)


eq-author/src/App/publish/PublishPage.js
```diff
@@ -113,10 +113,21 @@ const PublishPage = ({ match, history }) => {
...
+
+  const canPublish = () => {
+    if (!questionnaire) {
+      return true;
+    }
+    return (
+      me.admin ||
+      questionnaire.createdBy.id === me.id ||
+      questionnaire.editors.some(el => el.id === me.id)
+    );
```
SamGodwin2: Think we can just check to see if `questionnaire.permission === "Write"` here. Should make it easier to keep the permission logic consistent.

eq-author/src/App/publish/PublishPage.js
```diff
@@ -113,10 +113,21 @@ const PublishPage = ({ match, history }) => {
   };
 
   const publishStatus = questionnaire && questionnaire.publishStatus;
+
+  const canPublish = () => {
+    if (!questionnaire) {
+      return true;
```
SamGodwin2: Should we allow publishing if the questionnaire is falsy?

eq-author/src/App/publish/PublishPage.test.js
```diff
@@ -133,10 +133,23 @@ describe("Publish page", () => {
...
 
-  it("should redirect to questionnaire when it is not Unpublished", async () => {
+  it("should redirect to questionnaire when it is not Unpublished", () => {
     questionnaire.publishStatus = AWAITING_APPROVAL;
     const { history } = renderPublishPage();
-    await flushPromises();
     expect(history.location.pathname).toBe(`/q/${questionnaire.id}`);
   });
+
+  it("should not redirect to questionnaire when user is owner", async () => {
```
SamGodwin2: Does this need to be `async`?

eq-author/src/components/EditorLayout/Header/index.js
```diff
@@ -97,24 +97,32 @@ export const UnconnectedHeader = props => {
...
-        data-test="btn-publish"
-      >
-        <IconText icon={publishIcon}>Publish</IconText>
-      </RouteButton>
-    );
+
+    if (
+      me.admin ||
+      questionnaire.createdBy.id === me.id ||
+      questionnaire.editors.some(el => el.id === me.id)
```
SamGodwin2: `questionnaire.permission === Write`?

eq-author/src/components/EditorLayout/Header/index.js
```diff
@@ -97,24 +97,32 @@ export const UnconnectedHeader = props => {
...
+    ) {
+      return (
```
SamGodwin2: Judging by the trello card I think the button is supposed to be disabled, not hidden when you can't publish.

eq-author/src/App/publish/PublishPage.js
heofs: questionnaire should not stay falsy. It was just an alternative to using the loading page.

SamGodwin2: But why not return false so we stop them from seeing the publish page while the questionnaire is loading?

heofs: This is for redirecting. Could return null or something on the page to prevent them from seeing the publish page while it is waiting for the questionnaire object to become available. But I dont mind showing the publish page instantly, after all it is the page you are going to. Redirect will kick in if they cant publish.

heofs:
> Looks good; however, if I am not an admin and have a questionnaire that I submitted for review be approved, I cannot return to the publish page to publish the survey.
> 
> These are the steps I took:
> 
> 1. From an account that is not an admin, create a new survey;
> 2. From the same account, navigate to the publish page and submit that page for approval;
> 3. From an account that is an admin, approve the survey;
> 4. From the same account in steps 1 and 2, try to navigate to the publish page - the button will remain disabled.

The publish button is supposed to remain disabled until you make a change to a published survey. I tried this and the button became enabled again as soon as I made a change to the questionnaire.

PR #827, opened 2019-12-20
Upgraded Draft-js-plugin-editor version

martyncolmer:
There is a version mis-match between draft-js and draft-js-plugin-editor.
draft-js 0.11.0 has a peer dependancy on editor plugin 3.0.0 (currently 2.1.1)
This mis-match is possibly causing our i is not a function error in sentry.

### DIFF
--- a/eq-author/package.json
+++ b/eq-author/package.json
@@ -92,7 +92,7 @@
     "draft-convert": "latest",
     "draft-js": "latest",
     "draft-js-block-breakout-plugin": "latest",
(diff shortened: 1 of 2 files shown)


martyncolmer:
> Works fine, but is there a dependancy here? Couldn't find one in `draft.js`'s package.json?

Worded that backwards, draftjs editor plugin depends on draftjs. v2.2.1 doesn't work fully with the latest version of draftjs.

PR #851, opened 2020-01-08
Increase margin to make space for error message

heofs:
Fixes the inline error message so it no longer overlaps with other content.

Design changes confirmed with Joe :)

### How to review 
1. To see the issue, reduce the screen size to about 1200x500.
2. Checkout this branch and check if fix works.
3. Should work for checkboxes answers, radio answers and confirmation questions.


heofs:
[Improving the way error messages are displayed](https://trello.com/c/DDkLOOCh/1376-improving-the-way-error-messages-are-displayed)

PR #857, opened 2020-01-10
Fix missing border lines and overflow of alias input text

heofs:
Fix issue with part of border lines disappearing when reducing zoom on chrome. This typically happened at zoom 90%. This was fixed by changing border size to `thin` instead of `1px`. Setting it to `1px` was forcing Chrome to do subpixel calculation, which had some strange side effects.

Also fixed the input box to only cover the visible area. Previously it would overflow behind the character count.

### How to review 
* Check that the border lines stay the same when zooming in and out.
* Check that entering shortcode on calc summary page, question page and section still works.
* Check that shortcode text doesnt overlap with character counter.


heofs:
[Improving the way error messages are displayed](https://trello.com/c/DDkLOOCh/1376-improving-the-way-error-messages-are-displayed)

PR #863, opened 2020-01-13
Update to restrict new users by email

martyncolmer:
This PR is to stop users with a certain email from getting into author.

A new env variable is required ALLOWED_EMAIL_LIST.


eq-author-api/docker-compose.yml
```diff
@@ -35,6 +35,7 @@ services:
       - JAEGER_SAMPLER_PARAM=1
       - REDIS_DOMAIN_NAME=redis
       - REDIS_PORT=6379
+      - ALLOWED_EMAIL_LIST="@ons.gov.uk, @ons.ext.gov.uk"
```
heofs: Will this require us to use ONS email addresses for development? Can we set this variable in the production environments instead?

eq-author-api/docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
       - DYNAMO_COMMENTS_TABLE_NAME=dev-author-comments
       - NODE_ENV=development
       - RUNNER_SESSION_URL=http://localhost:5000/session?token=
-      - PUBLISHER_URL=http://localhost:9000/publish/
+      - PUBLISHER_URL=http://host.docker.internal:9000/publish/
```
heofs: I think this will break "View survey" for people who are not running runner with docker-compose.

eq-author-api/middleware/identification/upsertUser.test.js
```diff
@@ -33,4 +38,23 @@ describe("Upsert User", () => {
...
+    expect(res.status).toHaveBeenCalledWith(401);
+  });
+
+  it("should allowed user with valid email to save", async () => {
+    process.env.ALLOWED_EMAIL_LIST = "@ons.gov.uk";
+    await upsertUser(req, res, next);
+    const user = await getUserByExternalId(req.user.externalId);
+    expect(user).toMatchObject(req.user);
+    expect(res.json).toHaveBeenCalledWith({ status: "OK" });
+  });
```
heofs: Maybe throw in a test here for an empty email list, for example `process.env.ALLOWED_EMAIL_LIST = ""`. I guess if it is empty all emails should be allowed.

eq-author-api/docker-compose.yml
heofs: Not sure if it has any other implications?

eq-author-api/docker-compose.yml
martyncolmer: Changing this to be an empty string

eq-author-api/docker-compose.yml
martyncolmer: Left as docker-compose version

eq-author-api/middleware/identification/upsertUser.test.js
martyncolmer: Added test to generate user when the allowed_email_list is set to empty string

heofs:
[Prevent unauthorised users from creating accounts on Author](https://trello.com/c/cFlQjWJz/1351-prevent-unauthorised-users-from-creating-accounts-on-author)

eq-author-api/docker-compose.yml
```diff
@@ -35,6 +35,7 @@ services:
       - JAEGER_SAMPLER_PARAM=1
       - REDIS_DOMAIN_NAME=redis
       - REDIS_PORT=6379
+      - ALLOWED_EMAIL_LIST=""
```
MattElcock: This variable should be set; otherwise, we'll need to manually enter it to ensure behaviour matches the real thing, where it should be like this by default.

eq-author-api/README.md
```diff
@@ -38,6 +38,7 @@ Changes to the application should hot reload via `nodemon`.
 | `PORT`                  | The port which express listens on (defaults to `4000`)                             | No       |
 | `NODE_ENV`              | Sets the environment the code is running in                                        | No       |
 | `ENABLE_IMPORT`         | When enabled it exposes a post endpoint for importing questionnaires               | No       |
+| `ALLOWED_EMAIL_LIST`    | Comma seperated string defining allowed email domains                              | No       |
```
heofs:
```suggestion
| `ALLOWED_EMAIL_LIST`    | Comma separated string defining allowed email domains in format "@domain.uk"                              | No       |
```

Misspelled `separated`. You can also add the format of the list to the readme.

PR #867, opened 2020-01-15
Added exclusion of std label if optional label is used

jchapman68:
Have included exclusion for std label "Select all that apply" if an optional label has an entry.  This exclusion is used for both the author preview and AnswerHelper label on main page.

### How to review 
Create survey with a question and a mutually checkbox answer.
Leave the optional label for the first checkbox empty
The "Select all that apply" helper text should be visible above the first checkbox and this label is used in the preview tab in Author.
Go back to survey and add an optional label
The helper text should clear and the option label should only be used on the preview tab.

### DIFF
--- a/eq-author/src/App/page/Design/answers/MultipleChoiceAnswer/index.js
+++ b/eq-author/src/App/page/Design/answers/MultipleChoiceAnswer/index.js
@@ -26,13 +26,6 @@ const AnswerWrapper = styled.div`
   width: 85%;
 `;
 
-const AnswerHelper = styled.div`
-  margin-bottom: 0.5em;
-  font-size: 0.9em;
-  font-weight: 600;
-  color: ${colors.text};
-`;
-
 const ExclusiveOr = styled.div`
   padding-bottom: 1em;
   font-size: 1em;
@@ -137,9 +130,6 @@ export class UnwrappedMultipleChoiceAnswer extends Component {
         type={answer.type}
       >
         <AnswerWrapper>
-          {answer.type === CHECKBOX && (
-            <AnswerHelper>Select all that apply</AnswerHelper>
-          )}
           <TransitionGroup
             component={Options}
             data-test="multiple-choice-options"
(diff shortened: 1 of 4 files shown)


jchapman68:
[Change preview when label of checkboxes is changed](https://trello.com/c/vqKM3KN5/1439-change-preview-when-label-of-checkboxes-is-changed)

eq-author/src/components/preview/Answers/MultipleChoiceAnswer.js
```diff
@@ -144,7 +144,7 @@ const MultipleChoiceAnswer = ({ answer }) => {
   return (
     <Field>
       <Legend>{answer.label}</Legend>
-      {answer.type === CHECKBOX && (
+      {answer.type === CHECKBOX && answer.label === "" && (
```
heofs:
```suggestion
      {answer.type === CHECKBOX && !answer.label && (
```

eq-author/src/App/page/Design/answers/MultipleChoiceAnswer/index.js
```diff
@@ -137,7 +137,7 @@ export class UnwrappedMultipleChoiceAnswer extends Component {
         type={answer.type}
       >
         <AnswerWrapper>
-          {answer.type === CHECKBOX && (
+          {answer.type === CHECKBOX && answer.label === "" && (
             <AnswerHelper>Select all that apply</AnswerHelper>
```
heofs: I think the card suggest completely removing 'select all that apply' on design page.

jchapman68: Yes, checked with Sarah, have removed the AnswerHelper block.

eq-author/src/components/preview/Answers/MultipleChoiceAnswer.js
jchapman68: Done

PR #868, opened 2020-01-15
Replaced removing all checkbox labels with just removing the second

jchapman68:
### Mutually exclusive optional label not showing in runner
When creating mutually exclusive checkboxes al the labels were removed which allowed runner to set an Or for the final checkbox but it meant the first checkbox group was also stripped of the label so runner used the Std one.  The blanket delete was removed and a delete for the 2nd label in the array - this potentially will have issues if other answer types precede the checkboxes.

### How to review 
Create a survey with mutually exclusive checkboxes leaving optional label blank.
Preview the survey in runner and observe the std label of "Select all that apply" is used.
Add an optional label for the checkboxes
Preview survey
Observe the new label is use.


jchapman68:
[Mutually exclusive checkbox label shows wrong text on Runner](https://trello.com/c/PpRn10X7/1356-mutually-exclusive-checkbox-label-shows-wrong-text-on-runner)

PR #891, opened 2020-01-29
Use new content picker for destination

heofs:
This pr changes the content picker to v2 when picking a routing destination.

### How to review 
Check that the new content picker follows the designs and still does what it should do. 

Steps to test:
1. Create a new questionnaire with multiple sections and multiple questions in each section.
1. Try and change the routing destination and make sure it follows the rules below.
1. Try and change the else destination.
1. Try and add multiple rules and change their destination.
1. Resize window to see that it fits different screen sizes.

**It should follow the following rules:** 
* Only questions should be displayed in the picker.
* Only questions **after** the current question should be displayed.
* In sections after the current section, only the first question should be displayed in the picker. 
* Selecting another sections question, routes to the section not the question.


eq-author/src/components/ContentPickerModal/index.test.js
```diff
@@ -138,76 +136,11 @@ describe("ContentPickerModal", () => {
...
-  it("should show an error message when there is no question", () => {
-    const wrapper = shallow(
-      <ContentPickerModal {...props} questionData={[]} />
-    );
-    const tabs = wrapper.find(BaseTabs).prop("tabs");
-    const QuestionTab = tabs[1].render;
-    const questionTabWrapper = shallow(<QuestionTab />);
-    expect(questionTabWrapper).toMatchSnapshot();
-  });
-
```
zimboflyman: I'm assuming these removed tests from this file are now covered in the other tests file like RoutingDestinationContentPicker.test.js or the functionality has been removed in this PR?

eq-author/src/App/page/Routing/DestinationSelector/RoutingDestinationContentPicker.test.js
```diff
@@ -25,70 +80,132 @@ describe("RoutingDestinationContentPicker", () => {
...
-
-    wrapper = render(props);
   });
 
   it("should render", () => {
-    expect(wrapper).toMatchSnapshot();
+    const { getByText, getByTestId } = render(
+      <UnwrappedRoutingDestinationContentPicker {...props} />
+    );
+    const openButton = getByTestId("content-picker-select");
```
zimboflyman: Just wondering if there is a way to create this variable as part of props so it only needs creating once?
I do the same too, so just a general question really!

eq-author/src/components/ContentPickerv2/DestinationPicker/DestinationEnd.js
```diff
@@ -0,0 +1,88 @@
...
+  align-items: center;
+`;
+
+// eslint-disable-next-line no-unused-vars
+export const Icon = styled(({ isActive, ...rest }) => <IconPage {...rest} />)`
+  margin-right: 1em;
+  .path {
+    fill: ${props => (props.isActive ? "white" : colors.darkGrey)};
+  }
+`;
```
zimboflyman: Do we need to pass isActive in here (line 27) in order for line 30 to work?

eq-author/src/components/EditorLayout/Header/PageTitle.js
```diff
@@ -31,7 +31,7 @@ const PageTitle = ({ children }) => (
 );
 
 PageTitle.propTypes = {
-  children: PropTypes.node.isRequired,
+  children: PropTypes.node,
```
zimboflyman: Great! this will fix a consol error for commentsPanel too :)

eq-author/src/components/ContentPickerModal/index.test.js
heofs: Yes, this functionality has been removed. These tests were testing the old content picker. If you look in `src/components/ContentPickerModal/index.js`, I removed lots of stuff there because it wasn't being used. So these tests are not relevant anymore.

eq-author/src/App/page/Routing/DestinationSelector/RoutingDestinationContentPicker.test.js
heofs: I could do something like this ?
```js
    let openButton, destinationPicker;
    beforeEach(() => {
      destinationPicker = render(
        <UnwrappedRoutingDestinationContentPicker {...props} />
      );
      openButton = destinationPicker.getByTestId("content-picker-select");
    });

    it("should only display questions after the current question in the section", () => {
      const { getAllByText, queryByText } = destinationPicker;
      fireEvent.click(openButton);
      expect(queryByText("s1q1")).toBeFalsy();

      getAllByText("s1q3");
      getAllByText("s1q4");
    });

    it("should only display first question in other sections", () => {
      const { getAllByText, queryByText, getByText } = destinationPicker;
      fireEvent.click(openButton);

      const section2Button = getByText("section2");
      fireEvent.click(section2Button);
      getAllByText("s2q1");
      expect(queryByText("s2q2")).toBeFalsy();

      const section3Button = getByText("section3");
      fireEvent.click(section3Button);
      getAllByText("s3q1");
      expect(queryByText("s3q2")).toBeFalsy();
    });
```

eq-author/src/components/ContentPickerv2/DestinationPicker/DestinationEnd.js
heofs: Good question. This is a fix to prevent `isActive` being passed down to the svg dom element. `isActive` is not a valid svg attribute, so it was throwing errors. What line 27 does is destructure `isActive` from the props, and we only pass the `rest` of them to IconPage. Thus removing the unwanted prop but still being able to use it for styling.

eq-author/src/components/EditorLayout/Header/PageTitle.js
heofs: Yes, it was quite annoying 😁

eq-author/src/App/page/Routing/DestinationSelector/RoutingDestinationContentPicker.test.js
heofs: I pushed that up now :) Shaved off 2 lines 😉

eq-author/src/components/ContentPickerv2/DestinationPicker/DestinationEnd.js
zimboflyman: ahhh - clever!

eq-author/src/components/EditorLayout/Header/PageTitle.js
jasehumphr: Skillz

eq-author/src/components/ContentPickerModal/index.test.js
zimboflyman: coolio!

PR #953, opened 2020-02-18
Changes comments resolver to root query

martyncolmer:
This PR is an attempt to change the comments resolver to a root query
Some attention is needed around pusub.

### DIFF
--- a/eq-author-api/schema/resolvers/base.js
+++ b/eq-author-api/schema/resolvers/base.js
@@ -239,7 +239,7 @@ const Resolvers = {
       subscribe: () => pubsub.asyncIterator(["publishStatusUpdated"]),
     },
     commentsUpdated: {
-      resolve: async ({ componentId, questionnaire }, args, ctx) => {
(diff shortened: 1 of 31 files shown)


eq-author-api/schema/resolvers/base.js
```diff
@@ -223,12 +239,19 @@ const Resolvers = {
       subscribe: () => pubsub.asyncIterator(["publishStatusUpdated"]),
     },
     commentsUpdated: {
-      resolve: async ({ pageId, questionnaire }, args, ctx) => {
-        ctx.questionnaire = questionnaire;
-        const page = await getPageById(ctx, pageId);
-        return page;
+      resolve: async ({ componentId, questionnaire }) => {
+        // ctx.questionnaire = questionnaire;
```
jasehumphr: One more to remove

martyncolmer: has been updated since

PR #962, opened 2020-02-24
Comments root resolver

zimboflyman:
**What is the context of this PR?**
Here we are adding commenting/Replies to the following pages:

1. Introduction page
2. Section Pages
3. Confirmation Pages

In order for the resolvers to handle returns - Each query now needs a key/value pair instead of just an Id (eg: introductionID: "id") so the resolver can identify the type of page we are dealing with and return the relevant results since the data structure is different when dealing with these different types of pages.

**How to review**
Create a new BUSINESS survey.
Commenting has been added to the PREVIEW part of each of the pages. so you will probably need to add a page title in order to activate the preview tab (next to Design tab)

On each of the pages above:
you will see a comments panel in the right hand side.
Please test that you can add/delete/edit comments and the same for replies.

To see a confirmation page - hit "add" on a question page.

To test that subscriptions are working -
Open a chrome incognito instance of Author and log in with a different user -
At the same time as logged in on different user on another chrome tab. (on another screen)
When updating comments/replies - you should automatically see the other instance of author refreshing (as long as you are on the same page) -
so multiple users logged into the same page will be able to send comments in real time!

### DIFF
--- a/eq-author-api/schema/resolvers/base.js
+++ b/eq-author-api/schema/resolvers/base.js
@@ -239,7 +239,7 @@ const Resolvers = {
       subscribe: () => pubsub.asyncIterator(["publishStatusUpdated"]),
     },
     commentsUpdated: {
-      resolve: async ({ componentId, questionnaire }, args, ctx) => {
+      resolve: async ({ componentId, questionnaire }) => {
         // ctx.questionnaire = questionnaire;
         const questionnareComments = await getCommentsForQuestionnaire(
           questionnaire.id
@@ -896,16 +896,16 @@ const Resolvers = {
 
   Reply: {
     user: ({ userId }) => getUserById(userId),
-    parentComment: async ({ parentCommentId, componentId }, args, ctx) => {
-      const questionnaire = ctx.questionnaire;
-      const questionnaireComments = await getCommentsForQuestionnaire(
-        questionnaire.id
-      );
-      const parentComment = questionnaireComments.comments[componentId].find(
-        ({ id }) => id === parentCommentId
-      );
-      return parentComment;
-    },
+    // parentComment: async ({ parentCommentId, componentId }, args, ctx) => {
+    //   const questionnaire = ctx.questionnaire;
+    //   const questionnaireComments = await getCommentsForQuestionnaire(
+    //     questionnaire.id
+    //   );
+    //   const parentComment = questionnaireComments.comments[componentId].find(
+    //     ({ id }) => id === parentCommentId
+    //   );
+    //   return parentComment;
+    // },
   },
 
   QuestionnaireInfo: {
--- a/eq-author-api/schema/resolvers/base.js
+++ b/eq-author-api/schema/resolvers/base.js
@@ -82,6 +82,8 @@ const {
   getQuestionnaireMetaById,
   createComments,
   saveModel,
+  saveComments,
+  getCommentsForQuestionnaire,
 } = require("../../utils/datastore");
 
 const { QuestionnaireModel } = require("../../db/models/DynamoDB");
@@ -148,6 +150,13 @@ const createNewQuestionnaire = input => {
   };
 };
 
+const publishCommentUpdates = (questionnaire, componentId) => {
+  pubsub.publish("commentsUpdated", {
+    questionnaire,
+    componentId,
+  });
+};
+
 const Resolvers = {
   Query: {
     questionnaires: async (root, args, ctx) => {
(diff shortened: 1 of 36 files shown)


zimboflyman:
[Enabling commenting for survey introduction and section introduction pages and confirmation questions](https://trello.com/c/AloXLVyN/1435-enabling-commenting-for-survey-introduction-and-section-introduction-pages-and-confirmation-questions)

eq-author-api/schema/resolvers/base.js
```diff
@@ -701,6 +719,141 @@ const Resolvers = {
...
+      const { componentId } = input;
+      const questionnaire = ctx.questionnaire;
+      const questionnaireComments = await getCommentsForQuestionnaire(
+        questionnaire.id
+      );
+
+      const newReply = {
+        id: uuid.v4(),
+        parentCommentId: input.commentId,
+        commentText: input.commentText,
```
MattElcock: You should include `commentId` and `commentText` in the destructuring of `input`, then you wont need to fo `input.`

eq-author-api/schema/resolvers/base.js
```diff
@@ -701,6 +719,141 @@ const Resolvers = {
...
+        userId: ctx.user.id,
+        createdTime: new Date(),
+      };
+      let parentComment = questionnaireComments.comments[componentId].find(
+        ({ id }) => id === input.commentId
```
MattElcock: If you do the above, this would become

```suggestion
        ({ id }) => id === commentId
```

eq-author-api/schema/tests/subscription.test.js
```diff
@@ -209,168 +209,135 @@ describe("subscriptions", () => {
...
+    //   await deleteComment(ctx, {
+    //     pageId: CalSumPageId,
+    //     commentId: commentId,
+    //   });
+    //   const result = await iterator.next();
+    //   expect(result.value.data.commentsUpdated).toMatchObject({
+    //     id: CalSumPageId,
+    //     comments: [],
+    //   });
+    // });
```
MattElcock: Looks like this got left in by mistake

eq-author-api/schema/resolvers/base.js
```diff
@@ -701,6 +719,141 @@ const Resolvers = {
...
+        throw new Error("No replies found!");
+      }
+
+      const replyToEdit = replies.find(({ id }) => id === input.replyId);
+      replyToEdit.commentText = input.commentText;
+      replyToEdit.editedTime = new Date();
+      await saveComments(questionnaireComments);
+
+      publishCommentUpdates(componentId);
+      return replyToEdit;
```
MattElcock: Same as above reg `input.____`

eq-author-api/schema/tests/comments.test.js
```diff
@@ -147,112 +115,113 @@ describe("comments", () => {
...
+  //     commentText: "an edited comment",
+  //   });
+
+  //   expect(queryEditedComment).toMatchObject({
+  //     id: commentId,
+  //     commentText: "an edited comment",
+  //     editedTime: expect.any(String),
+  //   });
+  // });
+
```
tjamesmac: Check if this needs to go

eq-author-api/schema/tests/comments.test.js
```diff
@@ -319,60 +259,108 @@ describe("comments", () => {
...
+    //     commentId: commentId,
+    //     replyId: replyId,
+    //   });
+
+    //   const queriedComment = await queryComments(ctx, {
+    //     pageId: createdCalSumPage.id,
+    //   });
+
+    //   expect(queriedComment.comments[0].replies).toHaveLength(0);
+    // });
```
tjamesmac: Check this too

eq-author-api/schema/tests/subscription.test.js
```diff
@@ -209,168 +209,135 @@ describe("subscriptions", () => {
       });
       questionnaire = ctx.questionnaire;
       createdQuestionPage = questionnaire.sections[0].pages[0];
-      questionPageId = createdQuestionPage.id;
-      createdCalSumPage = questionnaire.sections[0].pages[1];
-      CalSumPageId = createdCalSumPage.id;
+      componentId = createdQuestionPage.id;
+      // createdCalSumPage = questionnaire.sections[0].pages[1];
+      // CalSumPageId = createdCalSumPage.id;
```
tjamesmac: Check these

eq-author/src/components/CommentsPanel/index.test.js
```diff
@@ -6,6 +6,7 @@ import COMMENT_SUBSCRIPTION from "./commentSubscription.graphql";
 
 import mocks from "./setupTests";
 import CommentsPanel from "./";
+// import { async } from "rxjs/internal/scheduler/async";
```
tjamesmac: Can be removed if not needed

eq-author-api/schema/resolvers/base.js
zimboflyman: Thanks matt - much cleaner and I've repeated in the comments resolver too!

eq-author-api/schema/tests/comments.test.js
zimboflyman: thanks - not sure how i missed that ):

PR #994, opened 2020-03-03
Update logger styling

MattElcock:

The logs produced by Author and the API can be confusing to read when debugging as they aren't styled very well. I've introduced a prettier that'll only run locally with the hope that it'll make debugging easier.

I've excluded publisher from this as we're currently rebuilding it.

### How to review 

* Make the code looks OK
* Make sure the logs look OK

### DIFF
--- a/eq-author-api/server.js
+++ b/eq-author-api/server.js
@@ -132,8 +132,8 @@ const createApp = () => {
   const httpServer = http.createServer(app);
   server.installSubscriptionHandlers(httpServer);
 
-  logger.info(`🚀 Server ready at ${server.graphqlPath}`);
-  logger.info(`🚀 Subscriptions ready at ${server.subscriptionsPath}`);
+  logger.info(`🚀  Server ready at ${server.graphqlPath}`);
+  logger.info(`🚀  Subscriptions ready at ${server.subscriptionsPath}`);
   return httpServer;
 };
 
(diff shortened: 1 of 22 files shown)


eq-author-api/server.js
```diff
@@ -135,8 +131,8 @@ const createApp = () => {
   const httpServer = http.createServer(app);
   server.installSubscriptionHandlers(httpServer);
 
-  logger.info(`🚀 Server ready at ${server.graphqlPath}`);
-  logger.info(`🚀 Subscriptions ready at ${server.subscriptionsPath}`);
+  logger.info(`🚀  Server ready at ${server.graphqlPath}`);
+  logger.info(`🚀  Subscriptions ready at ${server.subscriptionsPath}`);
```
tjamesmac: thank you

eq-author-api/utils/logger.js
```diff
@@ -0,0 +1,29 @@
+const noir = require("pino-noir");
+
+let logger;
+
+switch (process.env.NODE_ENV) {
+  case "development": {
+    logger = require("pino")({
+      prettyPrint: {
+        colorize: true,
+        ignore: "pid,hostname",
```
tjamesmac: I like that

martyncolmer:
Screenshot 2020-03-03 at 17 02 47

eq-author-api/utils/logger.js
```diff
@@ -0,0 +1,29 @@
...
+    break;
+  }
+  default: {
+    logger = require("pino")();
+  }
+}
+
+const expressLogger = require("express-pino-logger")({
+  logger,
+  serializers: noir(["req.headers.authorization"], "[Redacted]"),
```
tjamesmac: I think it might be something to do with this

PR #1000, opened 2020-03-06
Automated checks now use GitHub Actions

MattElcock:

Way back when, Henning raised that moving from Travis to GitHub Actions may resolve CodeCov being a bit weird at times. He never got to finish it before he left, so I have carried it on in some downtime here and there.

The idea was that GitHub Actions would generate the coverage files that CodeCov uses differently and therefore might make it get less confused. Honestly, I've not reallt seen much of a difference thus far but that might change as we go forward.

However, as a side effect, trying GitHub Actions has shown that the automated checks stage of our Continuous Integration pipeline can, essentially, be ran in half the time. This is because, currently, Travis runs the linter, unit tests and uploads to CodeCov sequentually for each of Author, the API and Publisher. Whereas, GitHub Actions runs in parallele which allows us to do all of this at the same time. Thus, making checks run equal to the longest library, which is an average 9 minutes, contrasting to Travis which runs in average 15 minutes.

### How to review 

* Have a look over the checks made on this PR and see if you understand them
* Have a look over the checks made on this PR and contrast them to Travis to see if I have left anything out
    * It builds docker images at the end of the build, so deploying to staging etc shouldn't be effected

I have deleted the Travis.yml file but this Pull Request will still be expecting it until we turn it off in our repo settings.

When I was running my tests, etc, I think I confused GitHub as there's now 12 checks. This isn't normal; I was just running them so much that I think it's duplicated a few. Normally, there's 3 GitHub Actions checks.


MattElcock:
When I was running my tests, etc, I think I confused GitHub as there's now 12 checks. This isn't normal; I was just running them so much that I think it's duplicated a few.

PR #1033, opened 2020-03-26
1507 implement firestore layer

MattElcock:

This task involved creating a new datastore so that the API can interact and store data within Firestore instead of DynamoDB. Functionally, Author (front end) has not changed as part of this.

This change does not require a migration, however in future we will need to decide upon a strategy for moving questionnaires from DynamoDB into Firestore.

### How to review 

Ensure that the `DATABASE` environment variable within the `docker-compose` file is set to `firestore`. This will change the database from DynamoDB to Firestore and allow the right datastore file to be used.

### DIFF
--- a/eq-author-api/db/datastore/datastore-firestore.test.js
+++ b/eq-author-api/db/datastore/datastore-firestore.test.js
@@ -1,11 +1,6 @@
 const { defineFeature, loadFeature } = require("jest-cucumber");
 const { createQuestionnaire } = require("./datastore-firestore");
 const { v4: uuidv4 } = require("uuid");
-const { Firestore } = require("@google-cloud/firestore");
-
-const creatingAQuestionnaire = loadFeature(
-  "./features/createQuestionnaire.feature"
-);
 
 jest.mock("@google-cloud/firestore", () => {
   class Firestore {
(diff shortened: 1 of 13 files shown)


eq-author-api/db/datastore/datastore-firestore.test.js
```diff
@@ -0,0 +1,529 @@
...
+      expect(questionnaire.id).toBeFalsy();
+      questionnaire.id = "123";
+      expect(questionnaire.id).toBeTruthy();
+      const questionnaireFromDb = await createQuestionnaire(questionnaire, ctx);
+      expect(questionnaireFromDb.id).toMatch("123");
+    });
+  });
+
+  describe("Getting the latest questionnaire version", () => {
+    it("Should should handle when an ID is not provided", () => {
```
tjamesmac: extra should

eq-author-api/db/datastore/datastore-firestore.test.js
```diff
@@ -0,0 +1,529 @@
...
+        "123",
+        mockHistoryEvent
+      );
+
+      expect(questionnaireHistory[0] === mockHistoryEvent).toBeTruthy();
+    });
+  });
+
+  describe("Saving a base questionnaire", () => {
+    it("Should handel when an ID cannot be found within the given base questionnaire", () => {
```
tjamesmac: 👉

PR #1093, opened 2020-04-14
Migrate Runner and Go launcher to GCP

MattElcock:
The EQAuthor team have been working on moving our product from V2 to V3 and moving everything from AWS to GCP.

This PR introduces a new endpoint to the API so that it can natively retrieve Author schemas in the Runner V3 format. This is necessary as Publisher V3 is no longer connected to the Author database and therefore is no longer able to retrieve Author schemas that need 'publishing'.

### How to review 

* Ensure all the tests pass and that, locally, your AWS ("old") set up works fine. 
* Switch to the GCP version and ensure the tests still pass and that the "new" set up works fine in the same way.
* * To do this, down V2 Publisher and up V3 instead
* * Change the DATABASE env variable in the API to firestore
* * Down V2 of Runner and up V3 instead
* * Update env vars where appropriate on the API
* Pop onto the hosted GCP version of Author and see if you can launch a survey from the "view survey" feature
* Pop onto the hosted GCP version of Launcher and see if you can launch a survey from there.

All links to hosted versions are on the Trello tickets.


eq-author-api/middleware/convertQuestionnaire.js
```diff
@@ -0,0 +1,46 @@
...
+  const CONVERSION_URL = process.env.CONVERSION_URL;
+  if (!usingGCP) {
+    res
+      .status(503)
+      .json({ error: "Endpoint not supported when using AWS configuration" });
+  } else if (!CONVERSION_URL) {
+    res.status(500).json({
+      error: "Cannot convert questionnaire without CONVERSION_URL env var set",
+    });
+  } else {
```
martyncolmer, line 16: Minor point. Don't need the "else" and the "else if" as the module would exit when you call res.status.json anyway.

PR #1173, opened 2020-05-13
Ear 212 add label to shortcode field fix size

jchapman68:

The Shortcode box should scale as window browser width is changed as it sits in the left hand half of a grid organising the the toolbar elements.

### How to review 


MattElcock:
> changing to grid has made the button hover sizes change to very large (delete, duplicate, move) as they are filling up the entire space (top to bottom)

We'll need to add a height of 2.125em to the buttons.

PR #1174, opened 2020-05-13
Added answer type to piping entity

MattElcock:

Piping DateRange answer types into question text has them be displayed as `[Deleted Answer]` if you leave and return to the page, or refresh. This is because the code which replaces piped answers in Author with `[Deleted Answer]` was triggering as, earlier down the line, another piece of code could not find the DateRange question ID from the Ritch Text Editor.

### How to review 

* Create a new questionnaire
* Have the first question be a date range answer
* Pipe either the first or second part of the date range answer into the text of the second question
* Refresh the page; if you do not see `[Deleted Answer]` then it works


MattElcock:
> On the author side everything looks great. When you try to look at the piped answers in Runner though, nothing shows up. Is this out of the scope of the ticket?

Yeah that's out of scope I think

PR #1188, opened 2020-05-20
Relative NextPage routing destination now displays as 'Next page'

MattElcock:

When adding a routing rule to a page for the first time, it creates a blank rule with the destination being the next page in the questionnaire. However, this is displayed to the user as the name of the question. The purpose of this task was to change the text displayed to the user to "Next page".

### How to review 

* Create a new questionnaire with a routable answer as question one and any other type as question two.
* Create a new routing rule on question one; the destination fields should show "Next page" instead of the destination question name. 
* Ensure the routing still works.


MattElcock:
> Does "NextPage" need to be a constant?

It can be, although I'm not sure if there's much value as it'll only be used once which is where it is now.

PR #1202, opened 2020-05-27
fix for broken routing

martyncolmer:
Create mulitple questions and sections.
In the final section add rules that reference earlier questions in both the same section and different sections.
Delete an answer and check rule shows correct message, 
Delete an question and check rule shows correct message, 
Delete an section and check rule shows correct message,


martyncolmer:
I'm happy that it refers to the question only. As it is the question that is used in the calc. Well answer really, but question sounds nicer.
Also the validation for left navigation will be another ticket

PR #1223, opened 2020-06-05
Ear 235 openclose all

tjamesmac:

~~Allows all the section accordions to be opened and closed with a single click.~~

~~The current method involves drilling props down a few layers which works, but I wish the context method would work.~~

~~I did try to use useContext to pass the props down the several layers. But the tests kept breaking even when wrapped in the context provider...~~

So this has been reworked ever so slightly since everyone looked at it last. Basically I noticed there were significant performance issues when you started to add loads of sections. I think they were being caused by lots of rerenders when the open/close state of the accordion was updated.  Which would cause the whole nav to render again including the section-title.

There are several things I've done to address these performance issues:
- Only update the label of the button if one of the accordions is closed
- Removed the useEffect at the top level -  could be done in an event handler instead
- Wrapped a few Callback hooks around certain functions in various components. The callback hook only instantiates a new function if one of the dependencies change in the callback hook. Wrapping this around the SectionTitle helped minimize the amount of stuff being rendered each time

### How to review 
0.5 Open chrome dev tools and click this button in the top left
Screenshot 2020-06-11 at 12 06 54

It should open up tabs on the left like this
Screenshot 2020-06-11 at 12 07 21

The verbose tab at the bottom shows when certain actions are taking longer than expected

1. Open questionnaire
2. Create as many sections as you want but more than 1 (Try like 20 if you can)
3. Hit Open all
    - should change to close all
    - if individual section is opened, they should all change to the expected state instead of the opposite of the current state 
4. Adding a new section and see if that works
5. When all sections are open, the button should say 'Close all'
6. When one section is open, the button should say 'Open all'
7. When there are a lot of sections, you should be able to scroll down the nav and the open all button should be fixed under the add content button.

### DIFF
--- a/eq-author/src/App/QuestionnaireDesignPage/NavigationSidebar/index.js
+++ b/eq-author/src/App/QuestionnaireDesignPage/NavigationSidebar/index.js
@@ -74,10 +74,10 @@ const actionTypes = {
 function reducer(state, action) {
   switch (action.type) {
     case actionTypes.updateGroup: {
-      const { identity } = action.payload;
+      const { identity, isOpen } = action.payload;
       const updateAccordionState = state.controlGroup
         .filter(x => x.identity !== identity)
-        .concat({ identity, isOpen: !action.payload.isOpen });
+        .concat({ identity, isOpen: !isOpen });
 
       const allOpen = updateAccordionState.every(x => x.isOpen === false);
       return { controlGroup: updateAccordionState, label: allOpen };
--- a/eq-author/src/App/QuestionnaireDesignPage/NavigationSidebar/index.js
+++ b/eq-author/src/App/QuestionnaireDesignPage/NavigationSidebar/index.js
@@ -79,7 +79,7 @@ export const reducer = (state, action) => {
       return { label: !state.label, isOpen: toOpen };
     }
     default:
-      throw new Error(`${action.type} isn't a valid dispatch type`);
+      throw new Error(`${action.type} is not a valid dispatch type`);
   }
 };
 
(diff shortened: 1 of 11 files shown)


eq-author/src/App/QuestionnaireDesignPage/NavigationSidebar/index.js
```diff
@@ -36,8 +38,18 @@ const NavList = styled.ol`
   list-style: none;
 `;
 
-export class UnwrappedNavigationSidebar extends Component {
-  static propTypes = {
+const AccordionGroupToggle = styled(Button).attrs({
```
MattElcock, line 41: This has legit blown my mind.

eq-author/src/App/QuestionnaireDesignPage/NavigationSidebar/index.js
```diff
@@ -47,58 +59,127 @@ export class UnwrappedNavigationSidebar extends Component {
...
+        isOpen: state.label,
+      }));
+
+      return { label: !state.label, controlGroup: toggleAll };
+    }
+    case actionTypes.setInitial: {
+      return { label: true, controlGroup: action.payload };
+    }
+    default:
+      throw new Error(`${action.type} isn't a valid dispatch type`);
```
MattElcock, line 97: We might need to escape the apostrophe here for when/if this error is ever shown. Think our linter throws a bit of a wobbly. Not sure if that's just locally though as in production it shouldn't matter.

jchapman68:
I get verbose errors when creating sections but none when opening or closing.
Screenshot 2020-06-12 at 12 21 02

MattElcock:
Closed as it's been done in another PR

PR #1242, opened 2020-06-17
Implement Skip Conditions Back End

MattElcock:
### Context

This pull request introduces the back end functionality for supporting Skip Conditions. Utilising what already exists for Routing, where possible, types and resolvers have been reused to ensure that A) the schema used in both are identical and B) the front end content picker maintains the same interface when talking to the API via resolvers.

This **does not** include publisher changes for translating between Author and EqRunner.

### How to test

Firstly, ensure the code looks OK and that Routing remains unchanged in functionality.

Some notes regarding the Skip Conditions schema:

- `skipConditions` is an array of objects, each defining a condition whereby the question on which the aforementioned field appears should be skipped. Multiple skip conditions are evaluated as OR rules (if any of the conditions evaluate to true, then the question is skipped).
- `expressions` is an array of objects, each defining a unique rule that must be satisfied in order to cause the condition they are part of to evaluate as true and consequently cause the question to be skipped. Multiple expressions are evaluated as AND rules. Each expression is a `BinaryExpression2` identical to what is found in Routing.

Then, boot up your prefered tool for sending API requests (e.g. Altair and Postman) and test out the following mutations:

**Note** - Skip conditions live on the question that they will cause to be skipped if evaluated to true.

- `createSkipCondition` - it should accept a single input, `pageId`, which is the identifier for the page to enable skip conditons on. If adding a skip condition for the first time, it should introduce the `skipConditions` property onto the Page's JSON blob and initialise an empty, default, skip conditon containing an empty, default expression. It should return the skip condition blob it created.

- `deleteSkipCondition` - it should accept a single input, `id`, which is the identifier for the condition you want to delete. If deleting the only skip condition on the page, the entire `skipConditions` property should be removed from the page JSON blob. It should return the page it removed the skip condition from.

- `createBinaryExpression2` - it should accept a single input, `expressionGroupId`, which is the identifier for the skip condition you created. It should create an empty, default expression within the `expressions` array in the JSON blob for the target skip condition.

- `deleteBinaryExpression` - it should accept a single input, `id`, which is the identifier for the skip condition you want to delete. It should delete the expression from the `expressions` array on the parent skip condition. If it is the only skip condition in the array, `expressions` should be empty.

- `updateBinaryExpression2` - it should accept two inputs, `id` and `condition`, which are the identifier for the expression you wish to update and the logical condition (e.g. Equals, NotEquals, Unanswered) to update on the expression, respectively. It is known that the name of this mutation is weird and a note has been made to consider renaming it when we refactor Routing. When called, it should update the condition on the target expression, return an error if the condition is not allowed for the answer type on the left of the expression or an error if there is no left side of the expression.

- `updateLeftSide2` - it should accept two inputs, `expressionId` and `answerId`, which are the identifiers for the expression you wish to update the left side of and the answer to use as the new left side. 

- `updateRightSide2` - it should accept two inputs, `expressionId` and `selectedOptions` OR `customValue` (depending on the left hand side of the expression), which is an identifier for the expression to update and the value to use when evaluating if the page should be skipped. If you give it both `customValue` and `selectedOption` it  should return an error.

You can test these in any order. However, to build a valid Skip Condition, you could call them like this:

**Prerequesite** - Build a question that is a multiple choice answer and a second one which is of any type. The first question will be used as the answer to evaluate whether the second question should be skipped.

1. `createSkipCondition`
2. `createBinaryExpression2`
3. `updateLeftSide2`
4. `updateBinaryExpression2`
5. `updateRightSide2`

### DIFF
--- a/eq-author-api/schema/tests/skipConditions.test.js
+++ b/eq-author-api/schema/tests/skipConditions.test.js
@@ -1,0 +1,91 @@
+const { buildContext } = require("../../tests/utils/contextBuilder");
+const { RADIO } = require("../../constants/answerTypes");
+const {
+  createSkipCondition,
+  deleteSkipCondition,
+  deleteSkipConditions,
+} = require("../../tests/utils/contextBuilder/skipConditions");
+const {
+  createBinaryExpression,
+} = require("../../tests/utils/contextBuilder/routing");
+
+const { queryPage } = require("../../tests/utils/contextBuilder/page");
+
+const config = {
+  metadata: [{}],
+  sections: [
+    {
+      title: "title-1",
+      alias: "alias-1",
+      position: 0,
+      pages: [
+        {
+          title: "page-1",
+          parentSection: "title-1",
+          answers: [
+            {
+              type: RADIO,
+            },
+          ],
+        },
+      ],
+    },
+  ],
+};
+
+describe("skip conditions", () => {
+  describe("expression group", () => {
+    it("should create an expression group and default expresiion", async () => {
+      const ctx = await buildContext(config);
+      const { questionnaire } = ctx;
+      const page = questionnaire.sections[0].pages[0];
+
+      createSkipCondition(ctx, page);
+      const result = await queryPage(ctx, page.id);
+      expect(result.skipConditions[0].expressions[0].left.reason).toBe(
+        "DefaultSkipCondition"
+      );
+    });
+
+    it("should delete an existing expression group", async () => {
+      const ctx = await buildContext(config);
+      const { questionnaire } = ctx;
+      const page = questionnaire.sections[0].pages[0];
+      const { id } = await createSkipCondition(ctx, page);
+
+      expect(page.skipConditions.length).toBe(1);
+      await deleteSkipCondition(ctx, id);
+      const result = await queryPage(ctx, page.id);
+      expect(result.skipConditions).toBeNull();
+    });
+
+    it("should delete all existing expression groups on a page", async () => {
+      const ctx = await buildContext(config);
+      const { questionnaire } = ctx;
+      const page = questionnaire.sections[0].pages[0];
+      createSkipCondition(ctx, page);
+      createSkipCondition(ctx, page);
+
+      expect(page.skipConditions.length).toBe(2);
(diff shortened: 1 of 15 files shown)


eq-author-api/schema/tests/skipConditions.test.js
```diff
@@ -0,0 +1,91 @@
...
+  });
+  describe("binary Expression", () => {
+    it("should add a binary exporession to an existing expression group", async () => {
+      const ctx = await buildContext(config);
+      const { questionnaire } = ctx;
+      const page = questionnaire.sections[0].pages[0];
+
+      await createSkipCondition(ctx, page);
+      const expressionGroup =
+        ctx.questionnaire.sections[0].pages[0].skipConditions[0];
```
tjamesmac, line 83: Super minor, questionnaire destructured above

tjamesmac:
Just wanted to add that what you've written up is super helpful. Perhaps it could be stored in the wiki for future reference because it is a fairly good run down of how it works. It helps give the idea of how binaryExpressions work etc.

If we do keep it, could we swap the ordering of the mutations to reflect the order they work in. Tried updating the binary expression before the left hand side before I scrolled to the bottom of the PR

MattElcock:
> Just wanted to add that what you've written up is super helpful. Perhaps it could be stored in the wiki for future reference because it is a fairly good run down of how it works. It helps give the idea of how binaryExpressions work etc.
> 
> If we do keep it, could we swap the ordering of the mutations to reflect the order they work in. Tried updating the binary expression before the left hand side before I scrolled to the bottom of the PR
> 
> Also some of the field id's are missing s's like the `selectedOption` one. I used the Apollo client in the dev tools and it worked really well

Cheers man! Yeah, I'll spend some time this sprint writing up how the mutations work on the wiki and try to give a run down of routing also. Glad it was helpful.

The notes about the s's etc I'll keep hold of and mention when we come to refactor routing. Stuff like that might need to be changed in the mutations on the front end, too. I have a few notes too on stuff that should be renamed etc, like the `updatebinaryExpression2` mutation you mentioned; it only updates the condition so we should rename it to reflect that.

Thanks for the feedback dude.

PR #1246, opened 2020-06-19
Ear-226 date range duration validation

tjamesmac:

Added validation to duration on date range questions.

This was a bit of a pain in the bum because all the validations were generated dynamically and didn't include a way to include another property error. 

Had to separate the way the date range validation buttons are generated. It felt a bit messy when I did it so would appreciate any comments on the front end part.

The validator logic makes no sense to me but for some reason it works.

### How to review 

1. Create a new questionnaire with a date range question
2. Add validation for both earliest/latest date and duration
3. Make the max bigger than the min
4. Error messages should appear for both

Look out for:
- Validation rendering in order as expected
- Error messages disappearing when conditions are corrected
- Error messages actually appearing
- Date questions working
- Multiple various questions working


eq-author-api/src/validation/index.test.js
```diff
@@ -446,53 +462,131 @@ describe("schema validation", () => {
...
+                    },
+                  },
+                },
+              ];
 
-            expect(pageErrors.validation).toMatchObject({});
-            expect(pageErrors.totalCount).toBe(0);
+              const pageErrors = validation(questionnaire);
+
+              console.log(pageErrors, "what is up yo");
```
jasehumphr, line 585: Yeah this will have to go

PR #1268, opened 2020-06-26
update for skip conditions

martyncolmer:
Update for skip conditions.
Will need the front end changes merged first to test publish route.


MattElcock:
Code looks good. Will give it a tick after I've tested it alongside the front end changes once they're merged in.

PR #1276, opened 2020-06-29
Add validation notification to Design Tab

zimboflyman:

add validation notification to Design Tab

### How to review 

GIVEN I have an error on the design tab
THEN I see a red dot on the label for the design tab

this should be visible when i'm on the preview and logic tabs

Same applies for confirmation and also calc sum page tabs


eq-author/src/components/EditorLayout/Tabs/index.js
```diff
@@ -62,7 +77,9 @@ const TABS = [
 ];
 
 export const UnwrappedTabs = props => {
-  const { match } = props;
+  const { match, page } = props;
```
martyncolmer, line 80: Why pass in the whole page when you can just pass in the validation errors.

eq-author/src/components/EditorLayout/Tabs/index.js
```diff
@@ -73,9 +90,14 @@ export const UnwrappedTabs = props => {
                 otherProps: { to: url(match), activeClassName },
               }
             : { Component: DisabledTab };
-
           return (
             <Component data-test={key} key={key} {...otherProps}>
+              {key === "design" &&
```
martyncolmer, line 95: Going to have an issue once we have logic page errors are included in with the page errors. 
Will have to filter the page errors in some fashion,

PR #1277, opened 2020-06-29
Add validation to empty routing answers

tjamesmac:

Add validation to routing pages. This focuses on no answer being selected when initialising routing rule. Picking no answer causes runner to 500.

There are a series of following tickets to add validation in the same area so I've tried to find a way that will make it easy to incorporate those rules

### How to review 

- Create questionnaire with a routing rule added
- When no question is picked a warning should appear
- Delete the rule and add a new one to see if that works


eq-author-api/src/validation/customKeywords/routingAnswer.js
```diff
@@ -0,0 +1,35 @@
+const {
+  ERR_ANSWER_NOT_SELECTED,
+} = require("../../../constants/validationErrorCodes");
+
+// not sure if this gonna cause problems
+const errorTypes = {
+  isNull: "Null",
+};
```
MattElcock, line 8: I don't think this will cause problems because it's a null value represented as a string, so the interpreter wont treat it as a null value. If that makes sense?

PR #1290, opened 2020-07-03
Ear 645 new routing tab show val error count

zimboflyman:

Let authors know if they have an error on the routing logic tab.

### How to review 

create a new routing rule and make sure you have not yet selected an routing answer
you should see a red validation dot on the routing tab with the number of errors for the routing page.

add a few more rules, but don't select any routing pages for each
you should see the correct number of validation errors showing in the red dot

you can look at the graphQL calls to make sure the error number corresponds to the routing errors


martyncolmer:
All the reduce code can be replaced with the lodash function (inside TAB.map)
`const errors = filter(page.validationErrorInfo.errors, error => error.id.includes(key) );`

You'll need to add a loading to the props and include 
`if (loading) {
      return <Loading height="20em">Loading logic page</Loading>;
 }`
This will stop the page erroring until the page data is available.

zimboflyman:
> All the reduce code can be replaced with the lodash function (inside TAB.map)
> `const errors = filter(page.validationErrorInfo.errors, error => error.id.includes(key) );`
> 
> You'll need to add a loading to the props and include
> `if (loading) { return <Loading height="20em">Loading logic page</Loading>; }`
> This will stop the page erroring until the page data is available.

Thanks Martyn - much simpler!

eq-author/src/App/page/Logic/index.js
```diff
@@ -67,36 +69,80 @@ const LogicLink = styled(NavLink)`
...
+    const TABS = [
+      {
+        key: `routing`,
+        label: "Routing logic",
+      },
+      {
+        key: `skip`,
+        label: "Skip logic",
+      },
+    ];
```
tjamesmac, line 119: By getting the data on line 100 this way, everytime you change the tab it completely blanks and shows the loading component for a split second before reloading

tjamesmac: Also the TABS are static so I think they might be able to be taken out of the rendeContent method

eq-author/src/App/page/Logic/index.js
```diff
@@ -67,36 +69,80 @@ const LogicLink = styled(NavLink)`
...
-              </li>
-              <li>
-                <LogicLink exact to={`skip`} activeClassName="active" replace>
-                  Skip logic
-                </LogicLink>
-              </li>
+              {TABS.map(({ key, label }) => {
+                const errors = filter(page.validationErrorInfo.errors, error =>
+                  error.id.includes(key)
+                );
```
tjamesmac, line 130: If instead you check page has loaded correctly here, the page remains intact on tab changes and still safeguards when the component is loading.

Something like this
```suggestion
           let errors;
                if (page) {
                  errors = filter(page.validationErrorInfo.errors, error =>
                    error.id.includes(key)
                  );
                }
```

martyncolmer: You could just put the loading test on the badge icon only
```
{!loading &&
  errors !== undefined &&
  errors !== null &&
  errors.length > 0 ? (
  <Badge data-test="badge-withCount">
  {errors.length}
   </Badge>
  ) : null}
```
and then delete the loading test on line 102.
This would keep the static part of the labels there then.

martyncolmer: This works well. I don't why. but it works. lol

eq-author/src/App/page/Logic/index.js
```diff
@@ -67,36 +67,75 @@ const LogicLink = styled(NavLink)`
...
+  color: white;
+  padding: 0.1em 0.3em;
+  font-weight: normal;
+  z-index: 2;
+  margin-left: auto;
+  line-height: 1;
+  font-size: 0.9rem;
+  pointer-events: none;
+  width: 20px;
+  height: 20px;
```
MattElcock, line 83: Could these be changed to em values rather than px, please? If you don't already know, I think 1em = 16px in our app so 20px would be 1.25em.

zimboflyman: Thanks Matt

PR #1291, opened 2020-07-03
Add warning error for empty right expression

tjamesmac:

Adds warning error and message to number, checkbox and radio answers when the right hand side is empty

### How to review 

1. Create questionnaire
2. Add three questions of types number, checkbox and radio
3. Go to each logic tab for each and add routing and skip condition using each answer type
4. Red border should surround the answer picker with either Enter a number (number) or Select at least one option
(radio, checkbox)

### DIFF
--- a/eq-author/src/App/page/Logic/BinaryExpressionEditor/MultipleChoiceAnswerOptionsSelector.js
+++ b/eq-author/src/App/page/Logic/BinaryExpressionEditor/MultipleChoiceAnswerOptionsSelector.js
@@ -26,8 +26,6 @@ const MultipleChoiceAnswerOptions = styled.div`
   flex-flow: row wrap;
   display: flex;
   width: 100%;
-  /* margin-top: 0.75em; */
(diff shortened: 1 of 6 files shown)


martyncolmer:
So, as the validation errors are passed with the expression to each of the different right hand components anyway. each component can then do its own validation styling and bespoke error message. no changes should be needed to the binaryExpressionEditor.

As the error component and icon would now be used in multiple components you could make it it's own module.

You can use lodash to find your error rather that a loop.
`some( { errorCode: "ERR_NO_RIGHT_VALUE_NUMBER" }, expression.validationErrorInfo.errors )` return true if the error exists
`filter( { errorCode: "ERR_NO_RIGHT_VALUE_NUMBER" }, expression.validationErrorInfo.errors )` can return an array of the error if required
`find( { errorCode: "ERR_NO_RIGHT_VALUE_NUMBER" }, expression.validationErrorInfo.errors )` can return the first error

PR #1301, opened 2020-07-06
Validate right side of routing and skip

tjamesmac:

Adds warning error and message to number, checkbox and radio answers when the right hand side is empty

### How to review 

1. Create questionnaire
2. Add three questions of types number, checkbox and radio
3. Go to each logic tab for each and add routing and skip condition using each answer type
4. Red border should surround the answer picker with either Enter a number (number) or Select at least one option
(radio, checkbox)

Codecov is back
Codacy is lame

### DIFF
--- a/eq-author/src/App/page/Logic/BinaryExpressionEditor/MultipleChoiceAnswerOptionsSelector.js
+++ b/eq-author/src/App/page/Logic/BinaryExpressionEditor/MultipleChoiceAnswerOptionsSelector.js
@@ -1,19 +1,22 @@
 import React from "react";
 import styled from "styled-components";
-import { includes, get } from "lodash";
+import { includes, get, some, find } from "lodash";
 import { PropTypes } from "prop-types";
 import { TransitionGroup } from "react-transition-group";
 
+import CheckboxChip from "./CheckboxChip";
 import CheckboxChipTransition from "./CheckboxChipTransition";
+import CheckboxOptionPicker from "./CheckboxOptionPicker";
+import Popover from "./CheckboxSelectorPopup";
+import ValidationError from "./ValidationError";
+
+import { rightSideErrors } from "constants/validationMessages";
 import { colors } from "constants/theme";
 import { RADIO } from "constants/answer-types";
 import { Select } from "components/Forms";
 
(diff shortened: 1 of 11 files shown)


eq-author/src/App/page/Logic/BinaryExpressionEditor/NumberAnswerSelector.js
```diff
@@ -67,59 +79,81 @@ class NumberAnswerSelector extends React.Component {
     this.props.onConditionChange(value);
   };
 
+  handleError = () => {
+    const { expression } = this.props;
+
+    const { errorCode } = find(expression.validationErrorInfo.errors, {
+      errorCode: "ERR_NO_RIGHT_VALUE",
+    });
```
MattElcock, line 87: This should use the constant that you've created rather than being hard coded.

eq-author/src/App/page/Logic/BinaryExpressionEditor/MultipleChoiceAnswerOptionsSelector.js
```diff
@@ -141,87 +153,118 @@ class MultipleChoiceAnswerOptionsSelector extends React.Component {
     this.setState({ showPopup: false });
   };
 
-  renderRadioOptionSelector() {
+  handleError = () => {
+    const { expression } = this.props;
+
+    const { errorCode } = find(expression.validationErrorInfo.errors, {
+      errorCode: "ERR_NO_RIGHT_VALUE",
+    });
```
MattElcock, line 161: This should use the constant that you've created rather than being hard coded.

eq-author/src/App/page/Logic/BinaryExpressionEditor/NumberAnswerSelector.js
martyncolmer: The constant is for the message not the errorcode passed in validationErrorInfo. Unless we add a constant.

martyncolmer: Nevermind, found it

PR #1307, opened 2020-07-08
comment out code todo with publish btn and tests

zimboflyman:

Hide the Publish button in the current version of Author because it does not currently work.
I've commented out code re: Publish button and tests
### How to review 

Check the publish btn is no longer in the menu

### DIFF
--- a/eq-author/src/App/QuestionnaireDesignPage/MainNavigation/index.js
+++ b/eq-author/src/App/QuestionnaireDesignPage/MainNavigation/index.js
@@ -1,7 +1,9 @@
 import React, { useState } from "react";
 import styled from "styled-components";
 import PropTypes from "prop-types";
-import { flowRight, get } from "lodash";
+import { flowRight } from "lodash";
+// import { flowRight, get } from "lodash";
+
 import { withRouter } from "react-router-dom";
 import gql from "graphql-tag";
 import { useSubscription } from "react-apollo";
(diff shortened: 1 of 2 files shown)


zimboflyman:
> I'm not sure how I feel about us commenting out code to hide a feature. Someone new might join the team and be confused as to why this happened, or we might forget, and it looks a bit untidy.
> 
> I'm happy to be outvoted, but I think, instead, we should either remove the code entirely or add a property into the CSS for the button to hide it.
> 
> What does everyone else think?

I don't feel great about deleting code that is going to be used later - As it's sometimes hard to trace your way back to finding that code when it comes time to adding it back in!
But, happy for a team decision and we can use that moving forward

MattElcock:
> > I'm not sure how I feel about us commenting out code to hide a feature. Someone new might join the team and be confused as to why this happened, or we might forget, and it looks a bit untidy.
> > I'm happy to be outvoted, but I think, instead, we should either remove the code entirely or add a property into the CSS for the button to hide it.
> > What does everyone else think?
> 
> I don't feel great about deleting code that is going to be used later - As it's sometimes hard to trace your way back to finding that code when it comes time to adding it back in!
> But, happy for a team decision and we can use that moving forward

That's completely fair! It's not overly important and, like you say, it'll be helpful for when we add it back in. I'll change mine to a tick.

PR #1310, opened 2020-07-09
Updated validation on checkboxes

martyncolmer:
Added custom validation to deal with exclusive checkboxes.


eq-author-api/src/validation/customKeywords/validateExclusiveCheckbox.js
```diff
@@ -0,0 +1,88 @@
...
+const {
+  getOptionById,
+  getExpressionGroupByExpressionId,
+} = require("../../../schema/resolvers/utils");
+const { filter } = require("lodash");
+
+module.exports = function(ajv) {
+  ajv.addKeyword("validateExclusiveCheckbox", {
+    validate: function isValid(
+      otherFields,
```
tjamesmac, line 15: Think you can delete this as it's not used

eq-author-api/src/validation/customKeywords/validateExclusiveCheckbox.js
```diff
@@ -0,0 +1,88 @@
...
+      entityData,
+      fieldValue,
+      dataPath,
+      parentData,
+      fieldName,
```
tjamesmac, line 20: And this one

tjamesmac: Well that's not helpful. I think `otherFields` and `fieldName` are the ones that aren't used

eq-author-api/src/validation/customKeywords/validateExclusiveCheckbox.js
martyncolmer: I did originally, but the arguments are positional.

tjamesmac: ah right

PR #1322, opened 2020-07-16
Ear 690 routing validate when later answer is set as left

zimboflyman:
When a valid routing answer (left) is moved _AFTER_ the routing Question - That routing rule is no longer valid and an error message must be shown

### How to review 

create 2 questions A and B in that order
B will have a routing rule that uses an Answer from A

_1. Move a question used as a previous answer in routing earlier_

GIVEN I've created a routing rule on question B which uses a respondent's answer to question A to determine the destination
WHEN I move question B earlier than question A
AND questions A and B are still in the same section
THEN I get an error message on question B letting me know that my routing is no longer valid because you can only route based on answers to the current question or previous questions
AND the content is Select an answer that is not later on in the questionnaire

_2. Move a question using a previous answer in routing later_

GIVEN I've created a routing rule on question B which uses a respondent's answer to question A to determine the destination
WHEN I move question A later than question B
AND questions A and B are still in the same section
THEN I get an error message on question B letting me know that my routing is no longer valid because you can only route based on answers to the current question or previous questions
AND the content is Select an answer that is not later on in the questionnaire


eq-author-api/src/validation/index.js
```diff
@@ -115,6 +115,7 @@ module.exports = questionnaire => {
   const errorMessages = validate.errors.filter(
     err => err.keyword === "errorMessage"
   );
+
```
martyncolmer, line 118: Random added space?

eq-author/src/constants/validationMessages.js
```diff
@@ -59,6 +59,13 @@ export const binaryExpressionErrors = {
   },
 };
 
+export const leftSideErrors = {
+  ERR_LEFTSIDE_NO_LONGER_AVAILABLE: {
+    errorCode: "ERR_LEFTSIDE_NO_LONGER_AVAILABLE",
+    message: "Select an answer that is not later on, in the questionnaire",
```
MattElcock, line 65:
```suggestion
    message: "Select an answer that is not later on in the questionnaire",
```

Rogue comma

zimboflyman: well spotted!

PR #1325, opened 2020-07-17
Update routing to not remove destination

martyncolmer:
Update to not remove the destination when a page or section is deleted.

To Test,
Add a second question and route to it from the first question.
Delete second question and check first question routing shows select a destination.

Repeat by adding a second section, routing to it and deleting it.


eq-author-api/schema/tests/routing.test.js
```diff
@@ -829,8 +823,7 @@ describe("routing", () => {
       firstPage = ctx.questionnaire.sections[0].pages[0];
 
       // Only rule to exist after deletion of third page
```
tjamesmac, line 825: Are the comments till relevant?

PR #1341, opened 2020-07-23
Err msg now displays correctly on smaller screens

MattElcock:

Jordan noticed that the error message for deleting a page used as the destination in a routing rule (and related error messages) were styled weirdly. This PR fixes that.

### How to review 

Set up a routing rule that goes to an absolute destination and then delete the page used as the destination. An error message should pop up. Reduce the screen size and the message should break onto two lines, instead of being overlaped which is what it was before.


jasehumphr:
Love the BBC way codecov pokes it nose in to a css change

PR #1350, opened 2020-07-30
Ear 364 unique qcodes

tjamesmac:

This PR refactors the Qcode table to make it a bit easier to follow. It also provides validation message around the qCode input if there are duplicate qCodes present within the table.

### How to review 

- Create questionnaire
- Add all the answer types to one question page
- Then create individual questions for each answer type
- Go to qCode page and add a qcode to each input
- Refresh the page to check they all save (i've been using 1-x to check for gaps)
- Start adding duplicates and check the validation message appears
- Validation shouldn't fire when the answer is `""` or `null`.

Not quite sure why the code coverage is less, not sure where I need to add coverage. I don't seem to be able to add coverage to the memo'd Row component.

Also I've tried rebasing this PR and it's pretty awkward. If it's okay, I would prefer not to please.


eq-author/src/App/qcodes/QCodesTable/index.js
```diff
@@ -38,304 +64,314 @@ const StyledTableBody = styled(TableBody)`
...
+        input: {
+          ...inputValues,
+        },
+      },
+    };
+  };
 
-  return content;
+  const { id, qCode } = payload;
+  if (questionMatrix[type] === "Confirmation question") {
```
MattElcock, line 238: Hi Tom, could you change "Confirmation question" to use the const you've made up above, please?

eq-author/src/App/qcodes/QCodesTable/index.js
```diff
@@ -38,304 +64,314 @@ const StyledTableBody = styled(TableBody)`
...
+    updateConfirmation(mutationVariables({ id, qCode }));
+  } else if (questionMatrix[type] === "Calculated summary") {
```
MattElcock, line 240: The same here, too, with "Calculated summary"

eq-author/src/App/qcodes/QCodesTable/index.js
```diff
@@ -38,304 +64,314 @@ const StyledTableBody = styled(TableBody)`
...
+          <>
+            <SpacedTableColumn>{questionMatrix[type]}</SpacedTableColumn>
+            <SpacedTableColumn>{label}</SpacedTableColumn>
+            <SpacedTableColumn>
+              <ErrorWrappedInput
+                value={qCode}
+                onChange={e => setQcode(e.value)}
+                onBlur={() => handleBlur(id, type, qCode)}
+                name={`${id || ""}-qcode-entry`}
+                data-test={`${id || ""}-test-input`}
```
MattElcock, line 298: Could you use something other than the empty string in these, please? I think it would make it cleaner. Or, would we ever not have `id` available here? It doesn't look like `handleBlur` accounts for if an `id` isn't available.

PR #1397, opened 2020-08-17
Ear 520 validate total merge test

tjamesmac:

Adds validation to totals

### How to review 

1. Create questionnaire
2. Add at least two number answers
3. Can use percentage, duration, number, unit
4. Add total validation
5. Check whether the message appears in the modal as well as in the right hand panel

EDIT: Okay so to get the coverage okay I added `hasStyleRule` to the custom editor test. Problem is, it passes no matter what. Should I get rid of it to have less coverage or keep it even though it's a false positive?


MattElcock (approved):
Noiceeeee

jasehumphr (approved):
Cheers Dude

PR #1404, opened 2020-08-20
Remove scroll bar from comment textArea's

tjamesmac:

Made the comment textArea expand by removing the row props

Needed to add a scroll bar to the pane itself though otherwise it was impossible to get to the bottom of the container to hit the add button.

**Help**
Looks kinda naff when you add another comment, need to show Joe

### How to review 

1. Create questionnaire
2. Add a comment
3. Check if there's a scroll bar


tjamesmac:
Joe mentioned last week that the comments section might be reworked and that would break this

MattElcock:
> Joe mentioned last week that the comments section might be reworked and that would break this

I spoke to Jordan and he said we can leave the comments box alone, keeping it with a scroll bar, for that reason.

MattElcock:
Closing this as it's over a month old.

PR #1414, opened 2020-08-21
819 and 820 - one scrollbar to replace right panel and editorLayout

zimboflyman:

This PR combines EAR 819 and 820
remove individual scrollbars from the right hand panel and editorlayout
create a single scrollbar to contain both

this will work the same for:

- design page
- preview page
- confirmation page
- calcsum page

### How to review 

create a question with enough answers to extend both the editor panel and the right hand panel
This should show a single scrollbar on the far right controlling both
The header will remain sticky
the scroll bar will not affect the nav panel on the left
If you shrink screen height below 680px - a global scroll bar will take over and the entire page including header and left nav will become scrollable
enjoy!


MattElcock (changes requested):
Scrollbar works perfectly. Nice one! My only reservation is that the properties panel cuts off at the bottom and doesn't span the entire page when you scroll. Can you check this with Joe and see if he wants it fixed as part of this ticket?

image

MattElcock (approved):

jasehumphr (approved):

jchapman68:
Scroll bar only visible in the header when screen height is shrunk so whole page is scrollable.

PR #1417, opened 2020-08-24
Add or to checkbox routing chips

tjamesmac:

Adds an `or` to the checkbox chip when adding options to the routing rules

### How to review 

1. Create questionnaire
2. Create a checkbox answer and add an `or`
3. Go to routing rules and add the or option
4. or should be displayed along with the option label
5. Remove the option label and see if it's replaced with `Unlabelled option`


MattElcock (changes requested):
Code looks good; nice one. 

From what I can see, there are a few minor things that differ from the design; I might be being a bit nitpicky, 'though.

- 'or' is not in bold
- the horizontal line after 'or' does not reach the top and bottom of the chip
- the x in the remove button on the chip is misaligned slightly

tjamesmac:
Not entirely sure the 'or' is actually bold. If I add bold, the rest of the text isn't and it looks weird. 

The X being misaligned is quite confusing because I know what you mean. It doesn't appear to do it consistently to all of them. 

I think I might try and take the chip apart and build it again. That will hopefully fix the line going to the top too

MattElcock:
> Not entirely sure the 'or' is actually bold. If I add bold, the rest of the text isn't and it looks weird.
> 
> The X being misaligned is quite confusing because I know what you mean. It doesn't appear to do it consistently to all of them.
> 
> I think I might try and take the chip apart and build it again. That will hopefully fix the line going to the top too

Weird about the bold... Is there a font-weight anywhere that's custom making it a bit-but-not-quite bold?

jasehumphr (approved):
Looks great. Spot on

MattElcock (approved):
Amazeballs. Nice one.

PR #1429, opened 2020-08-27
Downgraded apollo-cache-inmemory dep

MattElcock:

Upgrading this dependency beyond this version breaks Author. We will need to revert it (this PR) and then make a ticket to investigate why and how to fix it.

### How to review 

Run `yarn && yarn start` on the front end to re-install dependencies and start the front end. Create a new questionnaire with one section, one question page and one answer of any type. Refresh the page. It should not error.

### DIFF
--- a/eq-author/package.json
+++ b/eq-author/package.json
@@ -90,7 +90,7 @@
   "dependencies": {
     "@apollo/react-testing": "latest",
     "@sentry/browser": "latest",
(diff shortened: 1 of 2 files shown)


jasehumphr (approved):
Seems to have fixed the issue

jchapman68 (approved):

PR #1431, opened 2020-08-27
Fix scrollPane after for main content

tjamesmac:

The scrollPane that was sending the panels back to the top was removed. Have re-added it to editor layout to replicate the same functionality.

### How to review

1. Create questionnaire
2. Add a bunch of sections and questions 
3. Switch pages
4. Left nav should stay and main and right should go to the top


MattElcock (approved):
Looks good to me!

jasehumphr (approved):
To be sure

jchapman68 (approved):

PR #1446, opened 2020-09-01
Scroll pane fix three

tjamesmac:

At lower screen heights the main navigation starts to stick out from the rest of the content as shown in the picture below. 

A few months ago I added a media query to stop this at screen sizes below 680px. With the removal of the Scroll panes this has caused problems again with no scrolling functionality. This is a problem because the generic thinkpads used are small enough to encounter this problem.

To rectify the issue i've added... a scroll pane! 

This is definitely a band-aid solution and might need some more thought in the future.

| Problem  | 
| ------------- | 
| Screenshot 2020-09-01 at 08 37 18
  | 

### How to review 

1. Create a questionnaire with lots of questions pages and answers
2. Reduce the screen height
3. At around 576px the left-hand nav should be scrollable


MattElcock:
Closing this as it's almost a month old.

PR #1452, opened 2020-09-04
Dependabot upgrades 34

MattElcock:

Lots of dependabot upgrades. 

This PR also includes a configuration file which tells Dependabot to limit the number of open PRs to 10 (total; 3 for the front end, api and publisher and 1 for the root).

### How to review 

Spin up the app and make sure everything works as expected.

You can use this to validate the Dependabot yml: https://dependabot.com/docs/config-file/validator/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,27 @@
+version: 2
+updates:
+- package-ecosystem: npm
+  directory: "/eq-publisher"
+  schedule:
+    interval: weekly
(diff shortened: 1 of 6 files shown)


zimboflyman (approved):
cool - diggin the dependabot.yml file! be interesting to see if it works over the next few weeks!

jasehumphr (approved):
Nice we could do it for just our repo. Plan B would have been fun though

PR #1469, opened 2020-09-04
Ear 792 num min value validate missing value

zimboflyman:
https://collaborate2.ons.gov.uk/jira/browse/EAR-792

Numeric min and max input validation
show validate error on custom or previous answer, when min or max values are empty

This should work for all numeric types:
number, unit, percentage and currency

Given I've created a validation rule for a number answer type
And left the value box blank
When I'm on the question page
Then the appropriate property should be highlighted and have an associated error message

Given I've created a validation rule for a number answer type
And left the value box blank
When I'm on the question page
Then the design tab should have an associated validation error

Given I've created a validation rule for a number answer type
And left the value box blank
When I'm on the question page
Then the left hand navigation should have an associated validation error on the specific question

Given I've created a validation rule for a number answer type
And left the value box blank
When I'm on the property modal
Then the value box should be highlighted and have an associated error message
and the Done button is disabled


MattElcock (approved):
Slapping a big ol' tick on this; looks really good, dude. Well done. Especially with that AJV JSON edit. That must have been tricky. 

How did you find editing the `createValidationError` function I made as part of the refactor and the outcome of the refactor all together; did it help at all, not, or made things worse?

martyncolmer (approved):

jasehumphr (commented):
ThE FlAsH gOt ThErE bEfOrE mE

PR #1480, opened 2020-09-07
Added label entry to unit types

jchapman68:

> https://collaborate2.ons.gov.uk/jira/browse/EAR-830
>
> Remove default type when adding a unit answer

### How to review 

* Create questionnaire
* Add answer type of unit
* Check unit type box reads "Select a type"
* Set a unit type
* Check "Select a type" is no longer available in the list


eq-author/src/App/page/Design/QuestionPageEditor/AnswersEditor/AnswerEditor/index.js
```diff
@@ -116,13 +116,7 @@ class AnswerEditor extends React.Component {
...
   getAnswerTypeText(answer) {
     if (answer.type === UNIT) {
-      const unitConfig = unitConversion[answer.properties.unit];
-      return (
-        <>
-          {unitConfig.unit}
-          {` (${unitConfig.abbreviation})`}
-        </>
-      );
+      return <></>;
```
tjamesmac, line 119: Need to replace the title of the question with something now that we have removed centimetres as the default. 

The import above might need to be removed as well

jchapman68: I've removed the check for UNIT so default title is used

MattElcock (changes requested):
I might be reviewing this wrong, I can still see a default unit type?

image

MattElcock (dismissed):
Nevermind, I was being stupid. Looks good.

eq-author/src/App/page/Design/QuestionPageEditor/AnswersEditor/AnswerEditor/index.js
```diff
@@ -115,15 +114,6 @@ class AnswerEditor extends React.Component {
...
   getAnswerTypeText(answer) {
-    if (answer.type === UNIT) {
-      const unitConfig = unitConversion[answer.properties.unit];
-      return (
-        <>
-          {unitConfig.unit}
-          {` (${unitConfig.abbreviation})`}
-        </>
-      );
-    }
```
tjamesmac, line 126: You can't remove this block of code because it adds the title to the top of answer. There needs to be a placeholder for when no unit is selected and that needs to be swapped out when the user does select a unit

tjamesmac (commented):
Think you just need to confirm whether it's okay for the answer title to just be `unit` and not add the extra unit type and it's good to go

tjamesmac:
@MattElcock think I did the exact same thing

martyncolmer (changes requested):
The change breaks the preview tab.

MattElcock:
> The change breaks the preview tab.

Nice spot. I didn't see that!

martyncolmer (changes requested):
I want to double check the behaviour of the answer title. Default is Unit, but once you select a unit I’d have expected the old behaviour of setting the Unit type. The ticket isn't clear on this.

eq-author/src/App/page/Design/QuestionPageEditor/AnswersEditor/AnswerEditor/index.js
```diff
@@ -115,15 +114,6 @@ class AnswerEditor extends React.Component {
   }
 
   getAnswerTypeText(answer) {
-    if (answer.type === UNIT) {
```
martyncolmer, line 118: Spoke to Joe and confirmed that the answer title should change to the unit type once a selection is made.
need to put this back in. I've added a line as well to return a default text.

    if (answer.type === UNIT) {	
      if (!answer.properties.unit) {
        return answer.type;
      }
      const unitConfig = unitConversion[answer.properties.unit];	
      return (	
        <>	
          {unitConfig.unit}	
          {` (${unitConfig.abbreviation})`}	
        </>	
      );	
    }

martyncolmer (changes requested):

MattElcock (approved):

martyncolmer (approved):

MattElcock (approved):

martyncolmer (approved):

PR #1481, opened 2020-09-07
GitHub actions update

MattElcock:

GitHub Actions always fails on dependency PRs. This is because Docker uses the branch name from dependabot to build and tag it's images, but the format is incorrect. Irregardless, we do not need to make new Docker images and push them for dependency PRs as when we merge them into master, an image with all the updated dependencies is made anyway.

Additionally, on all PRs, an image is made regardless of if the code actually works.

This PR makes sure that the tests have passed before a Docker image is made. If the PR is a dependency PR (known by the 'dependency' label being present, which is automatically added on Dependabot PRs), it does not run the build scrips for the Docker images.

### How to review 

Ensure all the tests are passing and nothing looks weird.

### DIFF
--- a/.github/workflows/apiPullRequest.yml
+++ b/.github/workflows/apiPullRequest.yml
@@ -3,7 +3,7 @@ name: API Pull Request
 on: [pull_request]
 
 jobs:
-  build:
+  Tests:
     runs-on: ubuntu-latest
 
     strategy:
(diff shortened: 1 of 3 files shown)


zimboflyman (approved):

tjamesmac:
@MattElcock previously if the unit tests failed did the build carry on or end the process there?

Nice that they've separated out into build and test because at a quick glance we can see if it's docker related or not.

tjamesmac (approved):
Marvellous

MattElcock:
> @MattElcock previously if the unit tests failed did the build carry on or end the process there?
> 
> Nice that they've separated out into build and test because at a quick glance we can see if it's docker related or not.

Derp. Unit tests would fail. Ignore my point in the PR around pushing broken code.

tjamesmac:
Oh no that's okay, I wasn't even thinking about that!

PR #1495, opened 2020-09-18
Tom b answer validation refactoring suggestions

MattElcock:

Tom B made some refactoring suggestions for Shane's PR (https://github.com/ONSdigital/eq-author-app/pull/1494/files) -- I made this PR which asks for it to be merged into Shane's branch just so we can see the proposed changes a bit easier.

### How to review 

### DIFF
--- a/eq-author/src/App/page/Design/Validation/AnswerValidation.js
+++ b/eq-author/src/App/page/Design/Validation/AnswerValidation.js
@@ -44,7 +44,6 @@ import {
   PERCENTAGE,
   UNIT,
 } from "constants/answer-types";
-import { join } from "lodash/fp";
 
 const formatValue = (value, { type, properties }) => {
   if (typeof value !== "number") {
@@ -185,19 +184,9 @@ class AnswerValidation extends React.PureComponent {
       const validation = get(answer, `validation.${validationType.id}`, {});
       const errors = get(validation, `validationErrorInfo.errors`, []);
 
-      const { id } = validationType;
       const { enabled, previousAnswer, metadata, inclusive } = validation;
 
-      if (answer.type === "DateRange") {
-        if ((id === "latestDate" || id === "earliestDate") && enabled) {
-          errorsArr.dateRange.range.push(...errors);
-        }
-
-        if ((id === "minDuration" || id === "maxDuration") && enabled) {
-          errorsArr.dateRange.duration.push(...errors);
-        }
-      }
-      errorsArr.other.push(...errors);
+      errorsArr.push(...errors);
 
       const value = enabled ? validationType.preview(validation, answer) : null;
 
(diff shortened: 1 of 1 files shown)


eq-author/src/App/page/Design/Validation/AnswerValidation.js
```diff
@@ -153,20 +138,22 @@ export const SidebarValidation = styled(SidebarButton)`
...
+      key={id}
+      data-test={`sidebar-button-${kebabCase(title)}`}
+      onClick={onClick}
+      hasError={hasError}
+    >
+      {children}
+    </SidebarValidation>;
 
-  handleModalClose = () => this.setState({ modalIsOpen: false });
+function AnswerValidation({answer}) {
```
MattElcock, line 151:
```suggestion
const AnswerValidation = ({answer}) => {
```

Nice one for kicking the class component to the bucket! 😂 Thanks for that. We're slowly trying to phase them out, where we can. So, what you've done here is really useful. 

The only comment I'll make is that we usually try to stick with the ES6 function syntax when writing components (and functions in general, really). I did a similar thing when I first started, too. This is partly because it's the style set from devs in the past so we try to keep things consistent but also as it's a more modern pattern to use and matches help guides and documentation for if we ever need to refer to them.

eq-author/src/App/page/Design/Validation/AnswerValidation.js
```diff
@@ -179,180 +166,88 @@ class AnswerValidation extends React.PureComponent {
...
+    if (type.id === "maxDuration") {
+      validationMessage = DURATION_ERROR_MESSAGE;
+    }
+    else if (type.id === "latestDate") {
+      validationMessage = EARLIEST_BEFORE_LATEST_DATE;
+    }
+    else if (answer.type === "Date") {
+      validationMessage = EARLIEST_BEFORE_LATEST_DATE;
+    }
+    else if (find(validationErrors, error => error.errorCode.include("ERR_NO_VALUE"))) {
```
MattElcock, line 221:
```suggestion
    else if (find(validationErrors, ({errorCode}) => errorCode.include("ERR_NO_VALUE"))) {
```

I only found this out a few weeks ago, but you can destructure objects as they're being used as well as when assigning them to variables.

eq-author/src/App/page/Design/Validation/AnswerValidation.js
```diff
@@ -179,180 +166,88 @@ class AnswerValidation extends React.PureComponent {
...
+    </ValidationContext.Provider>
+  );
 }
 
 AnswerValidation.propTypes = {
   answer: CustomPropTypes.answer,
 };
 
-export default AnswerValidation;
+// Use memoisation to replace previous use of PureComponent
```
MattElcock, line 252:
```suggestion
```

When it comes to showing why you've changed something, try to keep that within the description of a PR instead of in comments within the code. In this example, a few months down the line the comment probably wont mean much as we'll likely forget we used PureComponent here. Nice one at pointing out the change of style, all the same. It's good to make sure people are aware of these things.

tbaggaley:
Thanks for looking over this Matt - ignore this PR for now; I'll re-submit a new PR to merge changes with Shane's code once his is merged as planned.

PR #1519, opened 2020-09-29
Ear 872 validate empty question definition

MattElcock:

> Describe what you have changed as part of this task and why; include a link to the Trello card/GitHub issue, related pull requests, etc as required.
>
> E.g. _"Currently, if you add a new question to Author the entire app crashed. This is because..."_

### How to review 


tjamesmac (commented):
Hi @jchapman68 I have fixed a linting error and made changes to the way the errors are picked up. Could you double check they work for you?

I think the way the errors are generated needs to be refactored a little bit. If there more errors added to this in the future they're just doubling the methods and variables needed to create the errors. 

I don't have an answer quite yet but can help tomorrow

eq-author-api/src/validation/schemas/entities.json
```diff
@@ -96,68 +89,102 @@
...
-        }
-      },
-      "else": {
-        "if": {
-          "properties": {
-            "pageType": {
-              "enum": [
-                "CalculatedSummaryPage"
-              ]
+      "allOf": [
```
martyncolmer, line 92: You've put the tests inside an allOf here, rather than testing each field in the properties. I would only do if this if testing multiple related senarios around the same field.

jasehumphr (approved):

zimboflyman (approved):
looks great !  good job

PR #1524, opened 2020-10-02
EAR 637 - Empty QCode validation - exploring other options

zimboflyman:
This is the 3rd iteration of this ticket!
I first tried to avoid using AJV as it would add to the totalErrorCount that is used in multiple places, but qCode errors shouldn't really be counted there.
secondly - Using react context nearly worked to pass props between parent(MainNaviagtion.js) and child(qCodesTable) but ran into some issues there too with keeping the red dot 'live' on the main Navigation.
Eventually went with the tried and tested AJV method, but having to take qCode errors away from the totalErrors was a necessary evil to keep the design tab errors and the individual answer errors clean.

Enjoy! this PR nearly made me cry :)

https://collaborate2.ons.gov.uk/jira/browse/EAR-637

Given I've created one or more questions in a questionnaire
when I have not entered the Qcodes for each questions
then I see validation on the navigation menu item

Given I've created one or more questions in a questionnaire
when I have not entered the Qcodes for each questions
then I see validation on the Qcodes page stating they are missing but 'View Survey' button will still be active

**Checkbox** (please see the attached example screenshot).

Given I've created a checkbox answer in a questionnaire
When I have not entered Qcodes against all of the options
then I should see validation stating they are missing

  *Please note 2: The 'View Survey' button must not become inactive due to Qcodes missing


tbaggaley (approved):
Nice one Shane - glad the last set of test fixes wasn't too extensive, haha. 

Just a heads up to other reviewers - this also addresses ticket [EAR-922](https://collaborate2.ons.gov.uk/jira/browse/EAR-922) which removes the qCode entry for the plain "Checkbox" question type (as only checkbox options are supposed to have qCodes required rather than the question itself) so both tickets can move through on JIRA when this is merged.

jasehumphr (approved):
SWEET

jasehumphr:
I have found that if you have a dupe qcode you don't get the red dot in the nav

zimboflyman:
> I have found that if you have a dupe qcode you don't get the red dot in the nav

Thanks Jason, well spotted - Will make a ticket to deal with this as Dups wern't part of the ticket

PR #1552, opened 2020-10-19
EAR 883 - Add validation for empty include/exclude section in question editor

tbaggaley:

Previously there were no validation errors associated with enabling the "include/exclude" optional question field and leaving it blank - see [the JIRA ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-883).

This PR adds validation via the AJV schema for the includes/excludes field and updates the front-end to display the appropriate message as per the ticket.

Screenshot 2020-10-19 at 13 36 29

### How to review 

Open a questionnaire in author. Enable the "include/exclude" optional field and leave it blank - the validation message should appear. Enter text and de-focus the element and the validation message should disappear.

### DIFF
--- a/eq-author-api/src/validation/schemas/definitions.json
+++ b/eq-author-api/src/validation/schemas/definitions.json
@@ -4,9 +4,7 @@
     "populatedString": {
       "type": "string",
       "pattern": "\\w+",
-      "errorMessage": {
-        "pattern": "ERR_VALID_REQUIRED"
-      }
+      "errorMessage": "ERR_VALID_REQUIRED"
     }
   }
 }
--- a/eq-author-api/src/validation/index.test.js
+++ b/eq-author-api/src/validation/index.test.js
@@ -111,6 +111,36 @@ describe("schema validation", () => {
         type: "page",
       });
     });
+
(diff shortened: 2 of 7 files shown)


eq-author-api/src/validation/schemas/definitions.json
```diff
@@ -4,9 +4,7 @@
     "populatedString": {
       "type": "string",
       "pattern": "\\w+",
-      "errorMessage": {
-        "pattern": "ERR_VALID_REQUIRED"
-      }
+      "errorMessage": "ERR_VALID_REQUIRED"
```
tbaggaley, line 7: In case you're wondering - I made this change so that using the `populatedString` ref will give the error for null values _as well as_ empty strings (this makes the actual AJV schema simpler because there's no need to also check for null). 

Previously it only gave an error when the `pattern` failed a match and ignored errors on `type`. This handles both now.

Once Shane's PR is through some of the cruft could be removed from the AJV schema file since a lot of the checks for null strings AND populatedStrings can be removed after this change.

martyncolmer (approved):
Looks good, does the Business.

MattElcock (approved):

PR #1568, opened 2020-10-29
EAR 981 - fix confirmation page validation

tbaggaley:

Confirmation pages were reporting an incorrect amount of errors and propagating these errors into the parent question, causing inappropriate errors in the parent. See [JIRA ticket ](https://collaborate2.ons.gov.uk/jira/browse/EAR-981)for a screenshot.

This PR:

- Re-jigs the AJV validation of the `confirmationPage` labels so that the labels are only checked for duplication **if they are not empty**. Previously two errors were generated for each empty label: one for 'label required' and one for 'duplicated value', giving a confusing error count in the nav pane.
- Adjusts the resolver for `questionPage` to stop confirmation page errors being included inside their parent question's `validationErrorInfo`. This is necessary due to the design of confirmation pages as "part" of the parent page rather than an independent page in their own right.

### How to review 

- Create or open a questionnaire
- Create a question or open an existing question
- Add a confirmation page for the question
- Check that the red 'error dot' in the navigation bar is accurate for the question & the confirmation page based on errors visible on the respective pages
- Check that duplicated labels and missing labels on confirmation page are still reported appropriately as errors

Screenshot 2020-10-29 at 15 58 14

### DIFF
--- a/eq-author-api/schema/resolvers/pages/questionPage.js
+++ b/eq-author-api/schema/resolvers/pages/questionPage.js
@@ -74,7 +74,7 @@ Resolvers.QuestionPage = {
   },
   validationErrorInfo: ({ id }, args, ctx) => {
     const pageErrors = ctx.validationErrorInfo.filter(
-      ({ pageId }) => id === pageId
+      ({ pageId, type }) => id === pageId && !type.startsWith("confirmation")
     );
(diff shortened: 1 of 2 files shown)


tjamesmac (approved):
Looks good!

MattElcock (approved):

PR #1589, opened 2020-11-03
Ear 336 history cancel button

farres1:

Implementing a cancel button to cancel an edit being made to a history item.

### How to review 


jasehumphr (dismissed):

eq-author/src/App/history/HistoryItem.js
```diff
@@ -112,6 +112,12 @@ const SaveButton = styled(Button)`
   margin-bottom: 1em;
 `;
 
+const CancelButton = styled(Button)`
```
tjamesmac, line 115:
```suggestion
const ActionButton = styled(Button)`
```

eq-author/src/App/history/HistoryItem.js
```diff
@@ -162,6 +168,15 @@ const HistoryItem = ({
...
+            <CancelButton
+              data-test="cancel-note-btn"
+              onClick={() => {
+                setNoteState(bodyText);
+                setIsEditActive(false);
+              }}
+            >
+              Cancel
+            </CancelButton>
             <SaveButton
```
tjamesmac, line 180:
```suggestion
            <ActionButton
              data-test="cancel-note-btn"
              onClick={() => {
                setNoteState(bodyText);
                setIsEditActive(false);
              }}
            >
              Cancel
            </ActionButton>
            <ActionButton
```

tjamesmac (approved):
Hey @farres1, looks really good! Only one suggestion and it would be to merge the cancel/save button into one component and then use that one twice

eq-author/src/App/history/HistoryItem.js
jasehumphr: Yeah that's a good idea

farres1:
> Hey @farres1, looks really good! Only one suggestion and it would be to merge the cancel/save button into one component and then use that one twice

I forgot to do this, I'll change it now. Thank you!

tjamesmac:
Okay @farres1 that's much better but I have two more things now.

Could you align the buttons so the right edge of the save button is inline with the text box? 

Also the cancel button currently has the wrong styling (checkout the `calSumContentPicker` cancel button)

Then it's good to go!

jasehumphr (approved):
Looks great Sam

tjamesmac (commented):
Looks awesome Sam, thanks for making those changes

tjamesmac (approved):

PR #1595, opened 2020-11-05
basic validation working

zimboflyman:

As an author
when I'm adding a unit answer
and I haven't chosen a type
I want to see an error telling me I need to selection required
so I can choose a type

Given I haven't chosen a type
When I'm on the question page
Then the design tab should have an associated error (dot on Design tab)

Given I haven't chosen a type
When I'm on the question page
Then the left hand navigation should have an associated error on the specific question


tjamesmac (approved):
Noice

MattElcock (approved):

PR #1612, opened 2020-11-12
EAR 917 - Add "insert link" functionality to rich text editor

tbaggaley:

Previously Author had no means to add external links into surveys, requiring manual editing of the runner JSON in order to add external links as HTML.

This PR adds an "Insert link" button to the rich text editor. It is enabled on non-title field rich text editors present on the Section editor page, Introduction editor page and Question editor page per the [JIRA Ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-917).

#### Main changes:

- Adds link button to toolbar
- Adds `LINK` entity type to draft.js `RichTextEditor` component with its associated paraphernalia (e.g. conversion to/from html, strategy/decorator plugin) 
- Creates `LINK` entities when button pushed (using modal) as per prototype - replaces selected text with link or inserts new link text as per AC
- Styling per prototype
- Some minor refactoring has taken place in order to allow multiple draft.js entities to be processed in a modular way within the `RichTextEditor` component (in order for this to co-exist amicably with the existing custom code for piped values).

### How to review 

See extensive AC on [JIRA ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-917).

### Screenshot

Screenshot 2020-11-17 at 13 05 04


jchapman68 (dismissed):

eq-author/src/components/RichTextEditor/LinkPlugin/index.js
```diff
@@ -0,0 +1,98 @@
...
+  [ENTITY_TYPE]: (entity, originalText) => (
+    <a href={entity.data.url} target="_blank" rel="noopener noreferrer">
+      {originalText}
+    </a>
+  ),
+};
+
+const Link = styled.a`
+  background: #d6e4ff;
+`;
```
MattElcock, line 29: Hi Tom. Inside this file, `src/constants/theme.js` there's an object called `colors`. Are there any colours in there that you could use instead of this new one? 

You can speak to Joe if you're not sure -- iirc we had a conversation at a retrospective (before you joined the team I think - so you wouldn't know about this) about trying to reuse as many colours as we can to avoid adding new ones for single uses.

If it turns out none of them are appropriate, you can add your new colour into that object and reference it in the code that you're written. The benefit is that we can centralise our colours so that if we ever need to bulk change them we can do in a single place.

MattElcock (changes requested):

eq-author/src/components/RichTextEditor/LinkPlugin/index.js
tbaggaley: Ah - cheers Matt. I'll take a look.

tbaggaley: `colors.paleBlue` is a pretty good fit; I've pushed the change now.

MattElcock (changes requested):
Hi Tom. Code looks good -- thanks for making that change.

I've found a bug functionally, 'though. When I add a link and go to add another one, the URL from my previous entry is pre-populated in the modal's dialogue box.

tbaggaley:
> Hi Tom. Code looks good -- thanks for making that change.
> 
> I've found a bug functionally, 'though. When I add a link and go to add another one, the URL from my previous entry is pre-populated in the modal's dialogue box.

Cheers for spotting this Matt - fixed now.

MattElcock (approved):

jchapman68 (approved):

PR #1625, opened 2020-11-16
EAR 833 - Improve unit selector

tjamesmac:

This PR involves changing the unit selector for unit answer types from a `<select>` to an new `autocomplete` component.

I've added the accessibility addon for storybook to author to help with accessibility going forward

I've also added `userEvent` to `react-testing-library` to get natural dom events.

Quick run down, `keyDown` events weren't causing the input focus to change in the `rtl` tests. They do with `userEvent`. Seems pretty straightforward and more about it is here:
https://github.com/testing-library/user-event

### How to review

1. Create a questionnaire
2. Make a unit type
3. Select a type by either using `space`, `enter` or `mouse click`
4. Check it saves in the JSON
5. Remove it and check it returns to `""`;
6. Also that validation errors reappear


eq-author/src/components/Autocomplete/SRStatus.js
```diff
@@ -0,0 +1,90 @@
...
+  }, [content]);
+
+  const noResults = length === 0;
+
+  let srContent = null;
+  if (noResults) {
+    srContent = tNoResults();
+  } else {
+    srContent = tResults(length);
+  }
```
MattElcock, line 49:
```suggestion
  let srContent = null;
  if (!length) {
    srContent = tNoResults();
  } else {
    srContent = tResults(length);
  }
```

eq-author/src/components/Autocomplete/SRStatus.js
```diff
@@ -0,0 +1,90 @@
...
+          height: "1px",
+          marginBottom: "-1px",
+          marginRight: "-1px",
+          overflow: "hidden",
+          padding: "0",
+          position: "absolute",
+          whiteSpace: "nowrap",
+          width: "1px",
+        }}
+      >
```
MattElcock, line 68: Hi Tom. Is there a reason why you haven't made this a React component and used Styled Components to apply the style.

MattElcock (changes requested):

eq-author/src/components/Autocomplete/SRStatus.js
tjamesmac: I don't think it really crossed my mind tbh. Replaced it with the visually hidden component now

MattElcock (changes requested):
(I might be wrong on this) but I think the dropdown is meant to show all items if you're focused on it and haven't typed anything. Right now, it doesn't show anything until I start typing.

tjamesmac:
@MattElcock I thought we recently removed all the scroll bars from author and are allowing users to choose whether they want them present or not?

That's the only reason I haven't made it permanent

farres1 (commented):
This is looking really good! There were a few things I noticed.

The first one is that the scrollbar remains visible on the prototypes when not in use, and in the implementation it is not shown when not in use. I saw that you responded to Matt about this though, and so I'm not sure which would be correct.

Something else I noticed is that on the prototype the “Selection required” error message is closer to the right side. However, previous implementations also have the error message close to the left side, the same as you have implemented. I’m not sure which would be correct, so it might be something to discuss with Joe.

---

I noticed something when removing the unit type. Sometimes the error message does not show when there is no unit type entered in the box - upon refresh when this happens, the box displays the most recently entered unit type; when this happens, the JSON also includes the most recently entered unit type instead of an empty string. I’ve listed the steps that I noticed caused this below:

1 - Create a question with a unit type answer.

2 - Select a unit type.

3 - Remove the unit type.

4 - Click out of the unit type input box.

When I refresh the page, the previously entered unit type is in the text box and JSON with no error message. Removing the unit type after this refresh works and displays the error message/updates the JSON, however entering a unit type and removing it without refreshing again after that causes the JSON not to update to include the blank string, and also does not display the error message. From what I've seen, when the unit type has been blank once before refreshing (and therefore the error message displayed), it is not possible to revert it back to blank without refreshing the page. Also, doing the following causes the opposite to happen (error message being displayed and JSON containing blank string when there is a unit type entered):

1 - Start with any unit type entered into the input box.

2 - Refresh the page to make sure the error message will be displayed.

3 - Highlight the unit type in the input box but don't remove it.

4 - Click out of the unit type input box.

5 - The error message is displayed and the JSON has a blank string as the unit. Refreshing the page shows a blank input box.

eq-author/src/App/page/PropertiesPanel/GroupedAnswerProperties/index.js
```diff
@@ -72,17 +73,24 @@ const GroupContainer = styled.div`
...
-  ${({ hasUnitError }) => hasUnitError && `
-  border-color: ${colors.red};
-`}
-`;
+const filterUnitOptions = (options, query) =>
+  Object.values(options)
+    .filter(
+      x =>
+        x.unit.toLowerCase().includes(query.toLowerCase()) ||
+        x.abbreviation.toLowerCase().includes(query.toLowerCase())
```
tbaggaley, line 81:
```suggestion
        x.unit.toLowerCase().includes(query.toLowerCase().trim()) ||
        x.abbreviation.toLowerCase().includes(query.toLowerCase().trim()) ||
        query.toLowerCase().trim().startsWith(x.unit.toLowerCase()) ||
        query.toLowerCase().trim().startsWith(x.abbreviation.toLowerCase())
```

What do you think of this? It makes it a bit more flexible in that whitespace at the start/end of user input is ignored - and it also means that when you type extra text after a matching fragment it still gives you suggestions. For instance, before this when you select an option and then you start editing it, you get a "No suggestions" message: 

Screenshot 2020-12-10 at 09 11 20

After this change the option is still displayed which might be less confusing:

Screenshot 2020-12-10 at 09 18 47

tbaggaley (changes requested):
Looking grand and works well now - couldn't reproduce Sam's bug after your latest commits. I've just made a few wee suggestions

tbaggaley (dismissed):

tjamesmac:
Spoke to Joe, scroll bar is gonna get added and waiting for RAD to decide whether groups are needed

farres1 (approved):

jchapman68 (approved):

farres1 (approved):

PR #1627, opened 2020-11-16
set default message for then routing

zimboflyman:

https://collaborate2.ons.gov.uk/jira/browse/EAR-609

Acceptance criteria (user perspective) 
Given I am creating a routing logic rule
when I add the rule
and select an answer
then I am presented with an empty 'Then' destination box
text should say: "Select a destination"

_note: currently there will be NO validation on this - to be done in seperate ticket - EAR-918_ 
_note: 'Else  Go to' will stay as 'Next page' default.__

### How to review 

create 2 new questions
on question 1, create a routing rule within logic
you should see the 'then'  default selection to say 'Select a destination'
you should see the  'else' default selection to say 'Next page'
routing should work as normal

### DIFF
--- a/eq-author-api/schema/resolvers/logic/routing2/routing2/index.js
+++ b/eq-author-api/schema/resolvers/logic/routing2/routing2/index.js
@@ -62,7 +62,7 @@ Resolvers.Mutation = {
               }),
             ],
           }),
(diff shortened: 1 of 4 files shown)


tjamesmac (approved):
Looks good man!

jasehumphr (approved):

PR #1632, opened 2020-11-19
EAR 1026 - Remove options from skipConditions when deleted

tbaggaley:

Options (of multiple choice questions - e.g. checkbox / radio) were being deleted from routing rules when deleted from the answer - but not from skip conditions. This left dangling references to invalid option IDs in the affected skip condition without any error validation visible to the user (but breaking the survey in runner).

This PR makes deleting options from checkbox / radio answers behave the same for skip conditions as for routing rules - the deleted options have all of their references in skip conditions deleted. The user will now see the standard "Select at least one option" error when all of a skip condition's options have been deleted in this way.

### How to review 

1. Given that I have created a `Radio` type answer and referenced a set of options in a skip condition, deleting this entire set of options from the answer should display a validation error for the logic rule ("Select at least one option")

2. Given that I have created a `Checkbox` type answer and referenced a set of options in a skip condition, deleting this entire set of options from the answer should display a validation error for the logic rule ("Select at least one option")

3. Check that deleted options are still removed from routing rules as per previous behaviour.


eq-author-api/schema/resolvers/base.js
```diff
@@ -519,19 +519,28 @@ const Resolvers = {
...
+            )
+          : [];
+        const rightHandSides = [...routingExprs, ...skipExprs].map(
+          x => x && x.right
+        );
+
+        rightHandSides.forEach(right => {
+          if (right && right.optionIds) {
+            remove(right.optionIds, value => value === removedOption.id);
+          }
```
MattElcock, line 543: You'll want to do this as an OR (`||`) instead of an AND (`&&`). Pages can have routing but not skip conditions, and vice versa. Having it as an AND would mean that you'll need both for this to happen. 

This also means that you'll need to pop the code for handling skip and routing in their own conditional block, too. You can probably keep your fail fast option, 'though, for pages that don't have either.

Your test likely works for skip conditions because the questionnaire JSON has routing rules on the page. It might be worth making some tests for if you have both, skip without routing and routing without skip. Your skip condition test can be repurposed into the one for if you have both.

MattElcock (changes requested):

eq-author-api/schema/resolvers/base.js
tbaggaley: Thanks Matt,

This was a way of unifying the logic for the resolver - otherwise as you'd say you'd need two almost identical blocks of logic which do essentially the same thing (but with slightly different variable names). The logic as is handles both skip conditions and routing - but it obviously depends on there being routing or skip logic to process, hence the initial early `return` if there's nothing to do.

The test doesn't have routing rules & skip logic on the same page - the first page only has routing logic and the second only has skip conditions, just like in an actual questionnaire. It works in practice too in Author - handles deletion of options from both routing (as before) and skip conditions irrespective of whether other pages have skip conditions / logic enabled.

So I think it's alright. Let me know if you disagree further ;-)

MattElcock: Sorry. You're absolutely right. I didn't look into it deep enough. Ignore me!

MattElcock (approved):

jasehumphr (approved):

eq-author-api/schema/resolvers/base.js
martyncolmer: For reference, there is a function in the utils.js that returns all the expressions for a questionnaire in a single array to  iterate over.

tbaggaley: Hmm, thanks Martyn. I'll have a dig. Can perhaps refactor as part of another PR.

martyncolmer: It's creatively named 'getExpressions'

tbaggaley: Very exciting. I might squeeze it into my folders update PR since that touches this file too.

PR #1650, opened 2020-11-30
EAR 903 - Update front-end GraphQL for Folders

tbaggaley:
## What is the context of this PR?

Fixes front-end graphql so that Author works as-is with the back-end folders changes. 
Provides a base from which front-end folder functionality can be implemented on top of.

##### Questionnaires
- Viewing list of questionnaires
- Creating a questionnaire
- Opening a questionnaire
- Deleting a questionnaire
- Duplicating a questionnaire
- Viewing QCodes page
- Viewing Sharing / Settings pages
- Validation error subscription

##### Pages
- Viewing a question page
- Creating a question page using the button on the question design page
- Creating a question page using the nav bar menu (context dependent as before)
- Calculated summary pages
- Duplicating a question page
- Deleting a question page
- Creating a calculated summary page
- Deleting a calculated summary page
- Creating a confirmation page
- Deleting a confirmation page
- Moving pages in same section
- Moving pages between sections - including generating replacement page when empty (now handled by server rather than client)

##### Answers
- Creating all answer types
- Updating answers
- Updating validation for answers

##### Sections
- Creating a section
- Duplicating a section
- Deleting a section, including deleting the final section

##### Logic
- The "No Routing" page
- Adding routing logic
- Adding skip logic
- Choosing routing destinations

### 🌵 The Bad
... Hopefully dealt with for now 🔫

### 🌵 The Ugly
- The move page functionality has a quick fix for now - flattening all folders into one list for the purpose of choosing positions. Will need re-doing properly when we get to the appropriate ticket.
- `optimisticResponse` removed for now from the `withMovePage` HOC - may be re-added later (if required) once that functionality is appropriately adjusted to be target folder aware.

## Back-end changes
- When a folder has all its questions deleted, delete the folder - unless it's the only folder left in the section. This was necessary as it was otherwise impossible to delete an only child question (of a folder) from the questionnaire.
- Resolvers for `QuestionPage.position` and `CalculatedSummaryPage.position` updated to return a page's position within parent folder rather than relative to the section.
- `duplicatePage` resolver updated so that if a question in a disabled folder is duplicated, the copied question is added to a new disabled folder. This is to keep consistency with other behaviour - newly coined questions are created in empty disabled folders by default. This change allows move page logic to work as is on the client and hopefully will simplify the implementation of the create folder UI etc.
- Added ability to query a page's parent `folder`.
- Added ability to query a folder's parent `section`.

## Publisher v2
- Publisher v2 has been modified just enough to work with folders in situ. Folders are mapped to groups in the Runner schema. Disabled folders are merged together to save producing needlessly large amounts of groups.
- Will need future work during the "add skip logic to folders" ticket.
- Matching changes made in PR for v3: https://github.com/ONSdigital/eq-publisher-v3/pull/15

## How to review

- Load up Author, create a questionnaire
- Try everything and see if it breaks
- Check that you can view the questionnaire in Runner

## What to do after everything is green

### DIFF
--- a/eq-author/src/App/QuestionnaireDesignPage/index.js
+++ b/eq-author/src/App/QuestionnaireDesignPage/index.js
@@ -14,12 +14,7 @@ import Loading from "components/Loading";
 import BaseLayout from "components/BaseLayout";
 import QuestionnaireContext from "components/QuestionnaireContext";
 import MainNavigation from "./MainNavigation";
-import {
-  SECTION,
-  PAGE,
-  QUESTION_CONFIRMATION,
-  FOLDER,
-} from "constants/entities";
+import { SECTION, PAGE, QUESTION_CONFIRMATION } from "constants/entities";
 import {
   ERR_PAGE_NOT_FOUND,
   ERR_UNAUTHORIZED_QUESTIONNAIRE,
--- a/eq-author/src/App/QuestionnaireDesignPage/index.js
+++ b/eq-author/src/App/QuestionnaireDesignPage/index.js
@@ -120,7 +120,7 @@ export class UnwrappedQuestionnaireDesignPage extends Component {
       // TODO
 
       default:
-        throw new Error("Unrecognised entity type.");
+        throw new Error("Unrecognised entity name.");
     }
 
     if (pageType === "QuestionPage") {
--- a/eq-author/src/App/QuestionnaireDesignPage/index.js
+++ b/eq-author/src/App/QuestionnaireDesignPage/index.js
@@ -238,7 +238,6 @@ export class UnwrappedQuestionnaireDesignPage extends Component {
                   </MainNav>
                   <NavigationSidebar
                     data-test="side-nav"
-                    loading={loading}
                     onAddSection={this.props.onAddSection}
                     onAddQuestionPage={this.handleAddPage("QuestionPage")}
                     canAddQuestionPage={this.canAddQuestionAndCalculatedSummmaryPages()}
--- a/eq-author/src/App/QuestionnaireDesignPage/index.js
+++ b/eq-author/src/App/QuestionnaireDesignPage/index.js
@@ -357,74 +357,77 @@ export const VALIDATION_QUERY = gql`
         validationErrorInfo {
           ...ValidationErrorInfo
         }
-        pages {
+        folders {
           id
-          validationErrorInfo {
-            ...ValidationErrorInfo
-          }
-          ... on QuestionPage {
-            answers {
-              ... on BasicAnswer {
-                id
-                validation {
-                  ... on NumberValidation {
-                    minValue {
-                      id
-                      validationErrorInfo {
-                        ...ValidationErrorInfo
+          pages {
+            id
+            validationErrorInfo {
+              ...ValidationErrorInfo
+            }
+            ... on QuestionPage {
+              answers {
+                ... on BasicAnswer {
+                  id
+                  validation {
+                    ... on NumberValidation {
+                      minValue {
+                        id
+                        validationErrorInfo {
+                          ...ValidationErrorInfo
+                        }
+                      }
+                      maxValue {
+                        id
+                        validationErrorInfo {
+                          ...ValidationErrorInfo
+                        }
                       }
                     }
-                    maxValue {
-                      id
-                      validationErrorInfo {
-                        ...ValidationErrorInfo
+                    ... on DateValidation {
+                      earliestDate {
+                        id
+                        validationErrorInfo {
+                          ...ValidationErrorInfo
+                        }
+                      }
+                      latestDate {
+                        id
+                        validationErrorInfo {
+                          ...ValidationErrorInfo
+                        }
                       }
                     }
-                  }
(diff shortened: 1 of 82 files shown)


eq-author/src/App/QuestionnaireDesignPage/index.js
```diff
@@ -81,31 +86,48 @@ export class UnwrappedQuestionnaireDesignPage extends Component {
...
-          break;
+    let sectionId, position, selectedPage, selectedSection, selectedFolder;
+
+    switch (entityName) {
+      case SECTION:
+        sectionId = entityId;
+        position = 0;
+        break;
+
+      case QUESTION_CONFIRMATION:
```
MattElcock, line 97:
```suggestion
```

Love a good switch statement 'though. Nice one. 👌😉

eq-author/src/App/QuestionnaireDesignPage/index.js
```diff
@@ -81,31 +86,48 @@ export class UnwrappedQuestionnaireDesignPage extends Component {
...
-      }
+
+        break;
+
+      case FOLDER:
+        // TODO
+        break;
+
+      default:
+      // TODO
```
MattElcock, line 128: This has been left in for when we do the front end work, is it? If so that's fine.

eq-author/src/components/PositionModal/index.js
```diff
@@ -1,149 +1,118 @@
...
+  margin-bottom: 0.25rem;
+  margin-top: 1.25rem;
+`;
+
+const Trigger = styled.button.attrs({ type: "button" })`
+  width: 100%;
+  font-size: 1em;
+  padding: 0.5rem 2em 0.5rem 0.5rem;
+  background: white url('${Icon}') no-repeat right center;
+  border: solid 1px #aeaeae;
```
MattElcock, line 24: Whilst we're in this file, we could swap this colour out for one that's in our Theme library.

MattElcock (changes requested):

eq-author/src/App/QuestionnaireDesignPage/index.js
tbaggaley: Haha, cheers Matt. That case is intentional though - it means that the following block handles both QUESTION_CONFIRMATION and PAGE types. I can add a "// Fall though" comment if you want so that it's clear that it is intentional?

eq-author/src/App/QuestionnaireDesignPage/index.js
tbaggaley: Yeah, just a placeholder so whoever picks it up can see where the new logic needs to go.

eq-author/src/components/PositionModal/index.js
tbaggaley: Sure thing

eq-author/src/App/QuestionnaireDesignPage/index.js
MattElcock: Oh right! Okay. No, that's fine.

eq-author/src/App/QuestionnaireDesignPage/index.js
```diff
@@ -81,31 +81,48 @@ export class UnwrappedQuestionnaireDesignPage extends Component {
...
         }
-      }
+
+        break;
+
+      // case FOLDER:
+      // TODO
+
+      default:
+        throw new Error("Unrecognised entity type.");
```
tjamesmac, line 123: Entity name or type?

tbaggaley: Could be `name` instead if you want. That line should never really get hit in practice - was just needed since ESLint insists on a `default` case (although it might help in dev in the rare case we mess up the name for a new entity type)

tjamesmac (approved):
As far as I can tell it's all working. Nice one!

Also nice one with the createMockQuestionnaire

eq-author/src/App/QuestionnaireDesignPage/index.js
tjamesmac: Just for the sake of future clarity. I'm not particularly bothered but someone else might be

tbaggaley: Okay, no probs. I'll rename

farres1 (commented):
Hi Tom. I've just started testing the PR. I tried creating a Business questionnaire, which worked as intended, but when I made a Social questionnaire, the app crashed. I tried making a Social questionnaire again after restarting the server, and the same crash happened. Making a Business survey after the restart still worked. 

It looks like making a Social questionnaire crashes the app.

tbaggaley:
> Hi Tom. I've just started testing the PR. I tried creating a Business questionnaire, which worked as intended, but when I made a Social questionnaire, the app crashed. I tried making a Social questionnaire again after restarting the server, and the same crash happened. Making a Business survey after the restart still worked.
> 
> It looks like making a Social questionnaire crashes the app.

Cheers Sam! This is fixed now.

MattElcock (approved):

PR #1655, opened 2020-12-09
Add mongodb support to author

martyncolmer:
Support now added for Mongo DB to allow larger JSON sizes.

To test-
Run docker-compose up --build
Create questionnaires and test all features

View tables in Mongo express by going to localhost:8081 in browser

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,4 @@ typings/
 fragmentTypes.json
 
 data/
+eq-author-api/globalConfig.json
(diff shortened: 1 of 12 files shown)


tbaggaley (dismissed):
Looks good. I think the performance is a bit better than before too - and I like the mongo express UI. Will be handy if we ever fiddle with how we represent surveys in the DB.

tjamesmac (dismissed):

MattElcock (approved):
Looks OK. I'll give it a tick as my suggestion is a super small thing. However, I'm not overly fond of the empty function within `datastore-dynamodb`. Hpapy to be overruled on this 'though.

tjamesmac (approved):

farres1 (approved):

jasehumphr (approved):

PR #1656, opened 2020-12-09
EAR-998 break-word in multiple preview places

zimboflyman:

Error occurs in multiple places.

**Introduction preview** - all editible content boxes

**Section Preview** - Section title and section intro titles

**Question Preview** - all different answer types - title, labels and descriptions, and optional checkboxes and radios 

**Question confirmation preview** - Question, +ve and -ve titles and descriptions

This happens in the Preview tab.

If text is really long it probably should be truncated so it doesn't expand out the side of the container

### How to review
Create very long words for titles etc in all of the above places and test that they now word-break as required

### DIFF
--- a/eq-author/src/App/introduction/Preview/index.js
+++ b/eq-author/src/App/introduction/Preview/index.js
@@ -16,7 +16,7 @@ import iconChevron from "./icon-chevron.svg";
 
 const Container = styled.div`
   padding: 0 2em 2em;
-  max-width: 40em;
+  word-break: break-word;
   font-size: 1.1em;
   p {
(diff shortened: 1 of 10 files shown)


tbaggaley (changes requested):
Looks good, just found one last case which isn't covered - when the answer is super long which is chosen as part of a calculated summary page:

Screenshot 2020-12-10 at 10 26 12

tbaggaley (approved):
Cheers Shane; all good now

farres1 (approved):
This looks good, and the calculated summary page is fixed now!

PR #1659, opened 2020-12-11
EAR 1065 - Improve validation performance (large surveys)

tbaggaley:

Currently Author API responds very slowly to graphql requests for large questionnaires - cloning the Low Carbon Survey locally (and using it on prod) allows observing response times as great as three seconds per query.

This PR improves Author-API's response times significantly for large surveys such as LCS by optimising the questionnaire validation code.

For more details see my report here: [Profiling & optimising validation performance](https://tbaggaley.github.io/ajv_perf).

NB: I ran the benchmarks and profiling on Martyn's new MongoDB branch, hence the lower overall times in the benchmarks in the report compared to current `master`; I think the mongo branch is better performing overall.

### How to review

- First open LCS on prod or import it locally using the `master` branch and observe the high response times.
- Switch to this branch and import LCS.
- Refresh the page, switch pages, etc. to generate network activity and observe the graphql response times.
- Should be significantly less (generally 50% less or more).


eq-author-api/src/validation/index.js
```diff
@@ -12,58 +11,58 @@ require("ajv-keywords")(ajv, "select");
...
-    err => err.keyword === "errorMessage"
+  const { dataPath, message } = error;
+
+  const splitDataPath = dataPath.split("/");
+  const field =
+    message === "ERR_QCODE_REQUIRED"
+      ? "qCode"
+      : message === "ERR_SECONDARY_QCODE_REQUIRED"
+      ? "secondaryQCode"
+      : splitDataPath.pop();
```
MattElcock, line 31: Hi Tom. I realise this block was here before, but I feel we could improve it's readability by swapping it to a switch statement. What do you think?

```suggestion      
      let field = ""
      
      switch(message) {
          case "ERR_QCODE_REQUIRED":
             field = "qCode";
             break;
         case "ERR_SECONDARY_QCODE_REQUIRED":
             field = "secondaryQCode";
             break;
         default:
             field = splitDataPath.pop();  
      }
```

MattElcock (changes requested):

eq-author-api/src/validation/index.js
tbaggaley: Sure, agreed :-)

MattElcock (approved):

jasehumphr (approved):
With great power comes great responsibility

PR #1676, opened 2020-12-16
Fix dates breaking when changing relative position

tjamesmac:

The position picker for date answers was recently refactored. When re-adding the relative positions we forgot to capitalise them.

This caused the backend to break as it was expecting a capital.

### How to review

1 - Create a questionnaire
2 - Write a question
3 - Add an answer of type Date
4 - Add answer label
5 - On the right sidebar, select either Set earliest date or Set latest date
6 - Switch on the date validation
7 - Change the before value to after
8 - This shouldn't break anything now

### DIFF
--- a/eq-author/src/App/page/Design/Validation/DateValidation/PositionPicker.js
+++ b/eq-author/src/App/page/Design/Validation/DateValidation/PositionPicker.js
@@ -2,9 +2,9 @@ import React from "react";
 import PropTypes from "prop-types";
 import { RelativePositionSelect } from "./components";
 
-const RELATIVE_POSITIONS = ["before", "after"];
+const RELATIVE_POSITIONS = ["Before", "After"];
 
-const PositionPicker = ({value, onChange, onUpdate}) =>
+const PositionPicker = ({ value, onChange, onUpdate }) => (
   <RelativePositionSelect
     name="relativePosition"
     value={value}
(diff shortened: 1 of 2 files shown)


eq-author/src/App/page/Design/Validation/DateValidation/PositionPicker.js
```diff
@@ -2,9 +2,9 @@ import React from "react";
 import PropTypes from "prop-types";
 import { RelativePositionSelect } from "./components";
 
-const RELATIVE_POSITIONS = ["before", "after"];
+const RELATIVE_POSITIONS = ["Before", "After"];
```
MattElcock, line 5: Inside the map you're doing `position.toLowerCase()`, so I think we should keep it consistent and make the `RELATIVE_POSITIONS` be all lower case too. Or, will it break something?

```suggestion
const RELATIVE_POSITIONS = ["before", "after"];
```

MattElcock (approved):

eq-author/src/App/page/Design/Validation/DateValidation/PositionPicker.js
tjamesmac: Just for clarity sake, the lowercase is causing the issue with the relative position

farres1 (approved):

PR #1687, opened 2020-12-29
EAR 485 - Add metadata fallback options

tbaggaley:

[JIRA Ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-485)

Previously Author had no way to specify fallback metadata values to be displayed when the preferred metadata value was empty/blank. Where desired in surveys this was achieved through manually editing the runner schema to add entries in the style of `first_non_empty(metadata['preferred'], metadata['fallback'])`.

This PR adds a UI for adding fallback keys to metadata within Author and updates publisher to generate the above output when fallback keys are provided.

Screenshot 2020-12-29 at 13 27 19

### Summary of changes
- Changes to `Autocomplete` to make it suitable for use as fallback key selector (adds optional `borderless` prop and makes drop-down menu float over page a la existing `KeySelect`; tested with VoiceOver and seems unchanged accessibility wise). Fixes bug where dynamic changes to `options` prop weren't being reflected in the UI.
- Add fallback selector to metadata page
- Client-side logic to provide list of available fallback metadata (must be same type +/- `_Optional` & not be the same key)
- Back-end changes to persist `fallbackKey` data
- Update tests/coverage for `Autocomplete`
- Add tests for `Row` and `FallbackSelector`
- Update publisher v2 & add test - publisher now checks if a `fallbackKey` is present, and if so, uses the `first_non_empty` modifier to produce the fallback behaviour in runner; this reflects the current approach used in existing surveys where `trad_as` has `ru_name` as fallback.
- Re-fetch the metadata page to trigger re-render of the metadata table on update to rows - allows all other rows to re-calculate possible / available fallback keys & delete those deleted server-side
- Delete fallback keys server-side when referenced key deleted & update UI accordingly

#### UI/API
- Go to the metadata page of a questionnaire
- Should be able to see/choose fallback keys for each metadata entry, default being "None"
- Fallback key selections should persist if you refresh / change page
- Should only be able to choose fallback keys of appropriate type - e.g. Date metadata entries can only have Date fallback entries
- Add a new metadata type or change type of existing elements - available fallback keys should update appropriately
- Delete a metadata key referenced as a fallback key - fallback key should be cleared

#### Publisher/runner
- Add a fallback key to a metadata type - e.g. `ru_name` for `trad_as`
- Delete the `value` of the `trad_as` entry so it's blank
- Pipe in `trad_as` within one of the questionnaire's text boxes
- View the survey in runner and you should see the value populated by whatever was in `ru_name` (the fallback key) rather than the empty `trad_as`.


jasehumphr (dismissed):
Works how expexted

jasehumphr (approved):

tjamesmac (approved):
https://user-images.githubusercontent.com/22731314/103773339-d35b0580-5022-11eb-9328-cbcf01b1496e.mov

PR #1694, opened 2021-01-05
EAR-918 Add destination required validation message

tjamesmac:

Adds a validation message to routing rules then destination picker. Messages informs user that a destination is needed.

### How to review

1. Add 2 questions
2. Add routing rule to the first question
3. See if the "Destination required" error message appears
4. It should disappear once a valid destination is selected
5. Create four more questions
6. Make a routing rule on the 2nd new question to the 3rd question
7. Move the 3rd question before the 2nd
8. Check the error message appears for destination moving `Select a destination that's later in the current section, or is another section`


jasehumphr (approved):
Cool

tbaggaley (approved):
Looking good

PR #1696, opened 2021-01-06
Ear 827 remove default numerical

farres1:
> BEFORE MAKING YOUR PR... There must be no linting errors, all tests must pass and PR is named after JIRA ticket number e.g. EAR-###.

Removing the default "Equal to" operator for numerical routing/skip logic, and replacing it with a "Select an operator" option.

### How to review

Open or create a questionnaire
Add a numerical answer
Create routing and skip logic, and check these display "Select an operator" as default
Check that "Select an operator" option is removed once an operator has been selected.

1. - [ ] Create a questionnaire
2. - [ ] Add numerical answers to questions
3. - [ ] Add routing logic using a numerical answer
4. - [ ] Check that the default operator is "Select an operator", and not "Equal to"
5. - [ ] Check that the routing rule's condition in the JSON is null
6. - [ ] Select any of the operators
7. - [ ] Check that the routing rule's condition in the JSON has updated
8. - [ ] Check that "Select an operator" option has been removed
9. - [ ] Add skip logic using a numerical answer
10. - [ ] Check that the default operator is "Select an operator", and not "Equal to"
11. - [ ] Check that the skip rule's condition in the JSON is null
12. - [ ] Select any of the operators
13. - [ ] Check that the skip rule's condition in the JSON has updated
14. - [ ] Check that "Select an operator" option has been removed


eq-author/src/App/page/Logic/BinaryExpressionEditor/NumberAnswerSelector.js
```diff
@@ -11,6 +11,7 @@ import { rightSideErrors } from "constants/validationMessages";
 import ValidationError from "components/ValidationError";
 
 const conditions = {
+  SELECT: "Select",
```
tbaggaley, line 14:
```suggestion
  SELECT: "",
```

Just a  wee suggestion - if we make the value saved when "Select an operator" is chosen to be an empty string the AJV validation will be more straightforward since we can use the existing keyword we have which checks for populated strings (and an empty string is more readily understood as not being a valid choice). Otherwise looks good, I'll try out in Author in a sec.

tbaggaley (changes requested):

eq-author-api/constants/routingConditions.js
```diff
@@ -1,3 +1,4 @@
+const SELECT = "Select";
```
tbaggaley, line 1:
```suggestion
const SELECT = null;
```

eq-author-api/schema/typeDefs.js
```diff
@@ -544,6 +544,7 @@ type SelectedOptions2 {
 }
 
 enum LogicCondition {
+  Select
```
tbaggaley, line 547:
```suggestion
```

eq-author/src/App/page/Logic/BinaryExpressionEditor/NumberAnswerSelector.js
tbaggaley, line 14:
```suggestion
  SELECT: null,
```

tbaggaley (changes requested):
Hey Sam - tried it out in Author and looks good. Just suggesting that we use `null` for the saved value when "Select an operator" is chosen - makes it consistent with everything else which tends to be `null` when unset / non-existent (and makes AJV validation more straightforward for the next ticket).

eq-author/src/App/page/Logic/BinaryExpressionEditor/NumberAnswerSelector.test.js
```diff
@@ -76,6 +76,7 @@ describe("NumberAnswerSelector", () => {
       defaultProps.expression.condition = null;
```
tbaggaley, line 76:
```suggestion
```

tbaggaley (changes requested):
Hey Sam, looks good now! This line doesn't do anything since you overwrite it with `"Equal"` straight after. Otherwise happy to tick 👍🏻

tbaggaley (approved):

jasehumphr (approved):

PR #1699, opened 2021-01-07
EAR-565 Add spacing when additional guidance panel is closed

tjamesmac:

Added space between `additional guidance panel` and `Description` when the additional guidance panel is closed.

### How to review
1. Check the spacing between the above is 2em when the agp is closed


jasehumphr (approved):

martyncolmer:
I'm wondering whether we should have a margin at the top of the label component. Save doing the if statement. Decision for joe.

martyncolmer (approved):

farres1 (approved):

PR #1706, opened 2021-01-11
Fix storybook

MattElcock:

Our build script for deploying the design system is failing. This is a first attempt at fixing it by upgrading to the latest version of the action, and installing a necessary dependency required to build the design system.

### How to review

Pull down the branch and try spinning up the application by running `yarn storybook`. 

You can also tripple check that the build script still works by running `yarn build-storybook`. However, this should be fine as it isn't broken.

You cannot check that the workflow will work.


jasehumphr (approved):
sweet

tbaggaley (approved):

PR #1711, opened 2021-01-13
EAR 1025 - Fix Author API hanging

tbaggaley:

Author API was hanging for 12-13s at 100% CPU due to AJV stalling sporadically on some requests. This is a work-around to fix the issue - it re-initialises AJV on every request rather than re-using one instance. Increases response times slightly but stops the API server from hanging / being killed.

Relates to [EAR-2015](https://collaborate2.ons.gov.uk/jira/browse/EAR-1025).

### How to review

Open up a couple of large surveys (e.g. FALS and LCS)
Switch rapidly between them
Click like a madman on different pages etc.
Shouldn't break API (previously network requests were sporadically taking 12-13s)

### DIFF
--- a/eq-author-api/src/validation/index.js
+++ b/eq-author-api/src/validation/index.js
@@ -3,14 +3,6 @@ const createValidationError = require("./createValidationError");
 
 const schemas = require("./schemas");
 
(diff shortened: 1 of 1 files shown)


martyncolmer (approved):

jasehumphr (approved):
Sweet

PR #1712, opened 2021-01-13
EAR-678-DateRange-label-validation

zimboflyman:

https://collaborate2.ons.gov.uk/jira/browse/EAR-678

'from' and 'to' date labels in DateRange answer type are not validating missing labels correctly

Enter a date label' - Error shows for the 2nd date label when the From date is missing a label - Even though TO date HAS a label.
If the 1st (from) date label has a label, but not the TO date - then no error is shown - it should show TO date as missing a label

### How to review

**Wrong behaviour (1)**
Create a question with a date range answer type
Fill out the first date label field
The error on the second box disappears 

**Expected behaviour (1)**
Create a question with a date range answer type
Fill out the first date label field
The error on the second box shouldn't disappear 

**Wrong behaviour (2)**
Create a question with a date range answer type
Fill out the second date label field
The error remains on the second box disappears 

**Expected behaviour (2)**
Create a question with a date range answer type
Fill out the first date label field
The error on the second box should NOT disappear

### DIFF
--- a/eq-author/src/App/page/Design/answers/DateRange/index.test.js
+++ b/eq-author/src/App/page/Design/answers/DateRange/index.test.js
@@ -2,25 +2,34 @@ import React from "react";
 import { shallow } from "enzyme";
 import DateRange from "./";
 import createMockStore from "tests/utils/createMockStore";
-
-const answer = {
-  id: "1",
-  label: "Lorem ipsum",
-  childAnswers: [
-    { id: "1from", label: "Period from" },
-    { id: "1to", label: "Period to" },
-  ],
-};
+import { DATE_LABEL_REQUIRED } from "constants/validationMessages";
+import { render as rtlRender } from "tests/utils/rtl";
 
 describe("DateRange", () => {
-  let handleChange;
-  let handleUpdate;
-  let store;
+  let handleChange, handleUpdate, store, props;
 
   beforeEach(() => {
     handleChange = jest.fn();
     handleUpdate = jest.fn();
     store = createMockStore();
+
+    props = {
+      answer: {
+        id: "1",
+        label: "",
+        type: "DateRange",
+        description: "test",
+        guidance: "",
+        secondaryLabel: "",
+        properties: {required: false},
+        displayName: "Untitled answer",
+        qCode: "yuky",
+        validationErrorInfo:
+        {
+          errors: [],
+        },
+      },
+    };
   });
 
   it("should render", () => {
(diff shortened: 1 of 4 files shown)


farres1 (approved):

eq-author/src/App/page/Design/answers/DateRange/index.test.js
```diff
@@ -2,36 +2,80 @@ import React from "react";
...
+        id: "71b5f8a2-5de1-4a9a-8923-acd7fd3b7080",
+        type: "answer",
+        __typename: "ValidationError",
+      },
+    ];
+    props.answer.validationErrorInfo.errors = error;
+
+    const { getAllByText } = rtlRender(<DateRange onUpdate={handleUpdate} {...props} />);
+
+    expect(getAllByText(DATE_LABEL_REQUIRED)).toBeTruthy();
```
MattElcock, line 61: Hi Shane, using `getAllByText` indeed asserts whether the error message is there, but it will return any and all instances of that error message regardless of where it is applied on the page. Therefore, I'm not sure we can't be sure that the error you're expecting is coming from the first label and not the second.

There's a singular varient of `getAllByText()` called `getByText()` you could use. Here's a link to the documentation: 

https://testing-library.com/docs/bs-react-testing-library/examples/#getbytext

https://testing-library.com/docs/dom-testing-library/cheatsheet/#queries

eq-author/src/App/page/Design/answers/DateRange/index.test.js
```diff
@@ -2,36 +2,80 @@ import React from "react";
...
+        id: "71b5f8a2-5de1-4a9a-8923-acd7fd3b7081",
+        type: "answer",
+        __typename: "ValidationError",
+      },
+    ];
+    props.answer.validationErrorInfo.errors = error;
+
+    const { getAllByText } = rtlRender(<DateRange onUpdate={handleUpdate} {...props} />);
+
+    expect(getAllByText(DATE_LABEL_REQUIRED)).toBeTruthy();
```
MattElcock, line 78: The same as the above comment.

MattElcock (changes requested):

eq-author/src/App/page/Design/answers/DateRange/index.test.js
tjamesmac: Or you could combine the tests and use getAllByText and check it has two errors present. One for the label and one for the secondary label

MattElcock: > Or you could combine the tests and use getAllByText and check it has two errors present. One for the label and one for the secondary label

My only issue with this is that it wouldn't assert that you can have one without the other.

zimboflyman: Well spotted, thanks Matt

MattElcock (approved):

farres1 (approved):

PR #1714, opened 2021-01-14
EAR-551 Sort out destination picker

tjamesmac:
## Useful links
[Ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-551)
[Designs](https://overflow.io/s/96LRCTYQ?node=2d584288&prototype)

Sorting out the destination content picker.

It should:
- Allow users to pick questions from the same sections
- Provide only the sections if they are routing outside of the current section
- Provide other destinations if any of the above don't work

### How to review
- Create questionnaire
- Add four question pages in section one
- Add an additional 3 sections
- Go to S1 q2 and create a routing rule
- Click `select a destination` and the modal should display
NOTE: Design of modal being fixed in another ticket
- Should be three tabs on the left (`Current`, `Later`, `Other`)
- Each tab should display correct destination values
- If clicking the `Else` destination selector, `Later` should not appear
NOTE: should be able to switch tabs and choose a destination with `Space` or `Enter`
- Confirm the selection and check it appears in the destination picker


eq-author/src/App/page/Logic/Routing/DestinationSelector/RoutingDestinationContentPicker.js
```diff
@@ -144,48 +80,50 @@ export const UnwrappedRoutingDestinationContentPicker = ({
...
+  });
+
+  return (
+    <>
+      <UnwrappedRoutingDestinationContentPicker
+        data={data}
+        loading={loading}
+        {...otherProps}
+      />
+    </>
```
tbaggaley, line 124: Is the fragment needed?

Also - do we still need to have two components here? We could just have RoutingDestinationContentPicker and do away with the wrapped/unwrapped distinction now you're using useQuery - or was it to make testing easier?

```suggestion
  return (
      <UnwrappedRoutingDestinationContentPicker
        data={data}
        loading={loading}
        {...otherProps}
      />
```

eq-author/src/components/ContentPickerv2/Menu.js
```diff
@@ -56,35 +56,38 @@ export const ParentMenuItem = styled(MenuItem)`
...
-      opacity: 1;
-    }
-    &:hover {
-      background: #f3f3f3;
-    }
-  }
+  ${props =>
+    props["aria-selected"] &&
+    css`
+      background-color: #f3f3f3;
```
tbaggaley, line 62: Is there a colour constant we can use while you're changing this bit?

eq-author/src/components/ContentPickerv2/Menu.js
```diff
@@ -56,35 +56,38 @@ export const ParentMenuItem = styled(MenuItem)`
...
+      &::after {
+        opacity: 1;
+      }
+      &:hover {
+        background: #f3f3f3;
```
tbaggaley, line 67: Le same

eq-author/src/components/ContentPickerv2/Menu.js
```diff
@@ -56,35 +56,38 @@ export const ParentMenuItem = styled(MenuItem)`
...
+    props["aria-selected"] &&
+    css`
+      background-color: ${colors.primary};
+      --color: ${colors.white};
+      --colorSecondary: ${colors.white};
 
-    background: ${colors.primary};
-    border-color: #377090;
-    z-index: 2;
+      border-color: #377090;
```
tbaggaley, line 84: Le same

eq-author/src/components/ContentPickerv2/Menu.js
```diff
@@ -56,35 +56,38 @@ export const ParentMenuItem = styled(MenuItem)`
...
+      z-index: 2;
 
-    &:hover {
-      background: #397596;
-    }
-  }
+      &:hover {
+        background: #397596;
```
tbaggaley, line 88: Le same

tbaggaley (changes requested):
Hey Tom, some super warm and fuzzy refactoring going on, excellent!

Just a couple small comments from code review. Haven't tried in Author yet

eq-author/src/components/ContentPickerv2/Menu.js
jasehumphr: Let's not hard code them innit

jasehumphr: We currently don't have that in the theme but it's a blue so I guess. ${colors.blue} would do it

farres1 (dismissed):

eq-author/src/App/page/Logic/Routing/DestinationSelector/RoutingDestinationContentPicker.js
tjamesmac: Removed unwrapped too

jasehumphr (approved):
S.W.E.E.T.

tbaggaley (approved):
Superfantastic

farres1 (approved):

PR #1737, opened 2021-01-21
EAR 1110 - Fix move page modal w/ folders (staging)

tbaggaley:

Imported surveys to folderised staging were having issues with moving, duplicating and creating questions - positioning was incorrect. The API creates & expects new questions to be in their own disabled folders at present; this PR updates the migration to do this.

Move page modal was then showing outdated page positions (if opened straight after moving the selected page) as it was using outdated data from props; component has been refactored to use questionnaire context instead. It now shows the correct position. Removed some unnecessary props & graphql query after this change.

### How to review

- [ ] Import a questionnaire into localhost
- [ ] Check that moving questions, duplicating & creating new questions results in the correct positioning of the moved / new question(s)
- [ ] Open up the move dialog straight after moving a page & check the page positions in the modal match their updated positions in the questionnaire


eq-author/src/App/page/Design/PageHeader.js
```diff
@@ -35,7 +34,6 @@ export class PageHeader extends React.Component {
     showDeleteConfirmDialog: false,
     showMovePageDialog: false,
   };
-  π;
```
tbaggaley, line 38: This made me smile - a single PI on its own. But it wasn't a linting error :D Weird.

farres1 (approved):

zimboflyman (approved):
Impressive work on refactoring MovePageModalTom!

eq-author/src/App/page/Design/PageHeader.js
tjamesmac: Was saving that for later

PR #1746, opened 2021-01-25
EAR 1114 - Update Author API to node version 14

tbaggaley:

Updates Author API node.js version to 14.

Node.js version 10 was hanging reliably with AJV validation on certain surveys - updating seems to resolve this. Node.js version 10 also goes EOL in April this year.

This PR reverts the previous temporary AJV fix - AJV schema is once again compiled once and re-used for each request; request times go back down to what they were before that fix.

This also lets us use new useful syntax on the back-end, such as the optional chaining operator `?.`.

### How to review

- Run `docker-compose up --build` and make sure the web server builds & runs as before
- If you're running the server locally, `nvm use 14`, then `yarn` and check it still works as before


martyncolmer (approved):
Tested and works fine

MattElcock (approved):

PR #1760, opened 2021-02-01
EAR 993 Fix error when min/max value entered without unit

farres1:
> BEFORE MAKING YOUR PR... There must be no linting errors, all tests must pass and PR is named after JIRA ticket number e.g. EAR-###.

Fix a crash that happens when a unit type answer is assigned a minimum or maximum value when a unit type has not been selected

### How to review

When no unit type is selected and a minimum/maximum value is assigned, check that Author does not crash

The minimum/maximum value should still be displayed in the unit properties, but no unit type should be displayed with the value

Changing unit type should display the new unit type in the unit properties, and removing the unit type again should display no unit type


jasehumphr (dismissed):

eq-author/src/App/page/Design/Validation/AnswerValidation.js
```diff
@@ -57,7 +57,13 @@ const formatValue = (value, { type, properties }) => {
...
   }
   if (type === UNIT) {
-    return `${value} ${unitConversion[properties.unit].abbreviation}`;
+    let valueWithUnit;
+    unitConversion[properties.unit] == null
+      ? (valueWithUnit = `${value}`)
+      : (valueWithUnit = `${value} ${
+          unitConversion[properties.unit].abbreviation
+        }`);
+    return valueWithUnit;
```
tjamesmac, line 66:
```suggestion
    return unitConversion[properties.unit] == null
      ? `${value}`
      :  `${value} ${
          unitConversion[properties.unit].abbreviation
        }`;
    
```

Does that work instead?

tbaggaley, line 66: Hey Sam, tested and working; looks good. Just got a suggestion for cutting down the code:

```suggestion
    return `${value} ${unitConversion[properties.unit]?.abbreviation || ""}`;
```

Gave it a whirl and it works the same. Also just a heads up - when an object's attribute can't be found it'll return `undefined` instead of `null`. Your comparison works because you've used `==` and `null == undefined`.

tbaggaley (changes requested):

tbaggaley (approved):

tjamesmac (approved):
Nice one, good job!

PR #1761, opened 2021-02-01
EAR-1124-Fix unit type selection

tjamesmac:

The unit type selector was persistent across different pages. This meant you could save a type on one page and it would save it in another instead

Handy video courtesy of @farres1 
https://user-images.githubusercontent.com/22731314/106477516-25673d80-64a0-11eb-8e58-9f5eac953c6a.mov

### How to review
- Create a questionnaire
- Create 3 or 4 questions with unit answers
- Should be able to assign unique types to each answer
- Should add/remove error messages consistently


martyncolmer (approved):
Work as expected

jasehumphr (approved):

PR #1779, opened 2021-02-08
Ear 1055 dependabot 3

tjamesmac:

This PR only updates the front-end and publisher. Tried to update the api but ran into issues with all the tests blowing up.

Would like to discuss at the tech session tomorrow or at retro introducing #1771 so we never have to do this again
### How to review

`yarn` and `yarn test` on the front-end and publisher


tbaggaley (approved):

jasehumphr (approved):
All tests pass 

Dependabots Roll Out!!

PR #1804, opened 2021-02-18
Ear 1147 sort comments order in older surveys

zimboflyman:
https://collaborate2.ons.gov.uk/jira/browse/EAR-1147
In Prod - some of the older comments in questionnaires aren't being sorted correctly
Looks like all comments in the DB _before_ the migration to MongoDB - were going to the top
MongoDB saves dates as ISODate('2021-01-18T10:46:41.360Z')

The fix i've added is to convert dates toString() before sorting and doing the sorting in the front end as the replies were done.

This fix will need to be pushed to prod before we can really confirm the bug is ehem 'sorted' 😊

### How to review
make sure new comments and replies are sorting correctly locally


farres1 (dismissed):

jasehumphr (approved):
Comments display as expected. Let's see how they do with some older ones

tjamesmac (changes requested):
Do the changes need to be one the front-end? Can we just move the new sort to the resolvers? Otherwise we are sorting comments twice now

martyncolmer (approved):

eq-author/src/components/CommentsPanel/index.js
```diff
@@ -402,8 +402,13 @@ const CommentsPanel = ({ componentId, me: { id: myId } }) => {
...
 
-  const displayComments = comments.map((item, index) => {
-    const { replies } = comments[index];
+  let sortedComments = comments.slice() || [];
+  sortedComments = sortedComments.sort((a, b) =>
+    b.createdTime.toString().localeCompare(a.createdTime.toString())
+  );
+
+  const displayComments = sortedComments.map((item, index) => {
+    const { replies } = sortedComments[index];
```
tjamesmac, line 411:
```suggestion
    const { replies } = item;
```

tjamesmac (approved):
All good and the change is optional

farres1 (approved):

PR #1824, opened 2021-02-26
Ear 1002 delete a folder

zimboflyman:

GIVEN I'm on a folder
WHEN I press the Delete button
THEN see a warning message which is the same as the Delete section warning, except the warning message is "All questions in this folder will also be removed. This may affect piping and routing rules elsewhere."

GIVEN I'm on a folder
WHEN I press the Delete button and I confirm the deletion
THEN the folder and all questions contained within it are deleted and the focus moves up to previous question or section (whichever is before the deleted folder)_

### How to review

make some folders in various locations and test deletion and redirect (next focus) is working


eq-author/src/App/folder/Design/index.js
```diff
@@ -34,14 +37,34 @@ const StyledPanel = styled(Panel)`
...
   const { loading, error, data } = useQuery(GET_FOLDER_QUERY, {
     variables: { input: { folderId } },
   });
 
+  let sectionId, folderPosition, pages;
+  if (data) {
+    sectionId = data.folder.section.id;
+    folderPosition = data.folder.position;
+    pages = data.folder.pages;
+  }
```
MattElcock, line 52: You can omit the conditional and move lined 49, 50 and 51 to after the (what is currently) line 97, the conditional to check if `data` is `null` or `undefined`.

`sectionId`, `folderPosition` and `pages` will then update in the same way you have now, but without checking to see if `data` is `null` or `undefined` twice.

eq-author/src/App/folder/Design/index.test.js
```diff
@@ -23,30 +29,92 @@ const mockUser = {
...
+
+// useQuery.mockImplementation(() => ({
+//   data: data,
+//   loading: false,
+// }));
+
+// useMutation.mockImplementation(() => ({
+//   data: deletefolderdata,
+//   loading: false,
+// }));
```
MattElcock, line 89:
```suggestion

```

eq-author/src/App/folder/Design/index.test.js
```diff
@@ -112,10 +226,47 @@ describe("Folder design page", () => {
...
+
+    await act(async () => {
+      await flushPromises();
+    });
+
+    await act(async () => {
+      await fireEvent.click(getByTitle("Delete"));
+    });
+
+    // expect(getByTestId("delete-confirm-modal")).toBeVisible();
```
MattElcock, line 261:
```suggestion
```

MattElcock (changes requested):

eq-author/src/App/folder/Design/index.js
zimboflyman: ahhh - of course - thanks Matt!

eq-author/src/App/folder/Design/index.test.js
zimboflyman: wooops !

MattElcock (dismissed):

eq-author/src/tests/utils/createMockQuestionnaire.js
```diff
@@ -2,7 +2,7 @@ import { times } from "lodash";
 
 const DEFAULT_ANSWER_COUNT = 0;
 const DEFAULT_PAGE_COUNT = 1;
-const DEFAULT_FOLDER_COUNT = 1;
+const DEFAULT_FOLDER_COUNT = 2;
```
tbaggaley, line 5:
```suggestion
const DEFAULT_FOLDER_COUNT = 1;
```

Could we keep this as 1 by default? If other tests only need one folder there's no need to inflate all the objects

Can specify `folderCount` when you call `buildQuestionnaire` if you want to override

tbaggaley (changes requested):
Hey Shane, tried it out and works well for first 2 ACs

But when I delete the last remaining Q, the folder gets deleted too instead of it re-creating the page

eq-author/src/App/folder/Design/index.js
```diff
@@ -91,11 +119,17 @@ const FolderDesignPage = ({ match }) => {
...
+          onDelete={() => setShowDeleteModal(true)}
           disableMove
           disableDuplicate
-          disableDelete
+          showDeleteModal={showDeleteModal}
+          handleModalClose={() => setShowDeleteModal(false)}
+          handleModalConfirm={handleModalConfirm}
           key={`toolbar-folder-${folderId}`}
+          displayName={
+            data.folder.alias ? data.folder.alias : "Untitled folder"
```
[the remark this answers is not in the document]
MattElcock: This is fixed now as part of some changes

eq-author/src/tests/utils/createMockQuestionnaire.js
zimboflyman: thanks Tom - that was a bit silly of me! sorted now

MattElcock (approved):

tbaggaley:
Mornin' Shane 👋

When I delete a calc sum page within a folder I get `Unhandled Rejection (TypeError): Cannot read property 'enabled' of undefined`

Screenshot 2021-03-08 at 08 35 15

eq-author-api/schema/resolvers/pages/index.js
```diff
@@ -68,17 +68,23 @@ Resolvers.Mutation = {
...
-      } else {
-        // If this is the section's last folder, re-populate it with a new question
+      } else if (
+        (section.folders.length === 1 &&
+          section.folders[previous.folderIndex].enabled === false) ||
+        section.folders[previous.folderIndex].enabled
+      ) {
         const newPage = createQuestionPage();
         section.folders[previous.folderIndex].pages.push(newPage);
       }
```
tbaggaley, line 90: The logic within deletePage can simplified a fair bit - this works and I've just copy & pasted the whole thing. GH won't let me suggest stuff outside the diff, but the if/else if branching is redundant.

Otherwise happy to tick, honest ;-) 

```
  deletePage: createMutation((_, { input }, ctx) => {
    const section = getSectionByPageId(ctx, input.id);
    const { previous } = getMovePosition(section, input.id, 0);

    const folder = section.folders[previous.folderIndex];
    folder.pages.splice(previous.pageIndex, 1);

    onPageDeleted(ctx, section, previous.page);

    if (!folder.pages.length) {
      if (section.folders.length > 1 && !folder.enabled) {
        const [deletedFolder] = section.folders.splice(previous.folderIndex, 1);
        onFolderDeleted(ctx, deletedFolder);
      } else {
        const newPage = createQuestionPage();
        section.folders[previous.folderIndex].pages.push(newPage);
      }
    }

    deleteFirstPageSkipConditions(ctx);
    deleteLastPageRouting(ctx);

    return section;
  }),
```

zimboflyman: works good - nice shortcut!

tbaggaley (approved):

MattElcock (approved):

PR #1863, opened 2021-03-09
EAR-1004 Move Question Into And Out Of A Folder

jchapman68:

Allows you to move questions into and out of folders. 

This slightly extends the current splicing method we use to add a question into a new position. This method doesn't really work with a non-flat structure that we have with folders now.

It might be reworked in the future depending on what Matt B and Joe discuss.

This PR includes:
- moving down into a folder
- moving up into a folder
- moving down out of a folder
- moving up out of a folder
- jumping below a folder when clicking on a folder and going down
- jumping above a folder when clicking on a folder and going up

**Moving down refers to starting above the position in which you intend to move too**
**Moving up refers to starting below the position in which you intend to move too**

### How to review
- Create a questionnaire with a question, folder with 3 questions and questions after the folder.
- Try to replicate the actions outlined above.

### DIFF
--- a/eq-author/src/App/page/Design/MovePageModal/index.test.js
+++ b/eq-author/src/App/page/Design/MovePageModal/index.test.js
@@ -62,6 +62,8 @@ describe("MovePageModal", () => {
 
   it("should close section select Modals on close", () => {
     wrapper.find("MovePageModal__Trigger").simulate("click");
+    
+    expect(getSectionModal(wrapper).prop("isOpen")).toBe(true);
 
     getSectionModal(wrapper).simulate("close");
 
--- a/eq-author/src/App/page/Design/MovePageModal/index.test.js
+++ b/eq-author/src/App/page/Design/MovePageModal/index.test.js
@@ -1,5 +1,5 @@
 import React from "react";
-import MovePageModal from ".";
+import MovePageModal, { buildPageList } from ".";
 import { shallow } from "enzyme";
 import { byTestAttr } from "tests/utils/selectors";
 import { buildQuestionnaire } from "tests/utils/createMockQuestionnaire";
@@ -9,6 +9,8 @@ jest.mock("components/QuestionnaireContext", () => ({
   __esModule: true,
   useQuestionnaire: jest.fn(),
 }));
+
+jest.mock("components/PositionModal");
 
 const getSectionModal = (wrapper) =>
   wrapper.find(byTestAttr("section-select-modal"));
@@ -58,13 +60,21 @@ describe("MovePageModal", () => {
     expect(getSectionModal(wrapper).prop("isOpen")).toBe(false);
   });
 
+  it("should close section select Modals on close", () => {
+    wrapper.find("MovePageModal__Trigger").simulate("click");
+
+    getSectionModal(wrapper).simulate("close");
+
+    expect(getSectionModal(wrapper).prop("isOpen")).toBe(false);
+  });
+
   it("should update selected section on change", () => {
     const selectedSection = mockQuestionnaire.sections[1];
     getSectionItem(wrapper).simulate("change", {
       value: mockQuestionnaire.sections[1].id,
     });
 
-    expect(wrapper.find("PositionModal").prop("options")).toStrictEqual(
+    expect(wrapper.find("PositionModal").prop("options")).toMatchObject(
       selectedSection.folders.flatMap(({ pages }) => pages)
     );
   });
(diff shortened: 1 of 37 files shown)


jasehumphr (approved):
Works wonderfully

martyncolmer (changes requested):
I have a questionnaire with 
Q1
Q2
F1

Trying to move Q1 to after Q2 then back to start and then after again. It jumps to the end.

jasehumphr:
> I have a questionnaire with
> Q1
> Q2
> F1
> 
> Trying to move Q1 to after Q2 moves it to after F1.

I don't get that

eq-author/src/components/PositionModal/index.js
```diff
@@ -32,64 +32,123 @@ const Trigger = styled.button.attrs({ type: "button" })`
...
+      item: orderedOptions[previousPosition],
+    });
   };
 
-  const handleOpen = () => setIsOpen(true);
-
-  const handleChange = ({ value }) => setSelectedPosition(parseInt(value, 10));
+  const handleChange = ({ value }) =>
+    setOption(({ item: previousItem }) => {
+      let count =
```
MattElcock, line 66:
```suggestion
      const count =
```

I think? I can't see it being modified anywhere else, but I might be wrong.

eq-author/src/components/PositionModal/index.js
```diff
@@ -32,64 +32,123 @@ const Trigger = styled.button.attrs({ type: "button" })`
...
+
+    const positionCalculation = isSame
+      ? sameMove
+      : differentSection
+      ? folderMove
+      : fromFolderToFolder
+      ? folderMove
+      : inOrOut
+      ? folderMove
+      : itemPosition;
```
MattElcock, line 109: Could this be changed to not use short hand, please? It's difficult to understand and I think we would benefit from having it use the tranditional if statements instead for if/when we read back over this.

MattElcock (changes requested):

martyncolmer (dismissed):
Adding a question to the root puts the new question inside the un-abled folder. The question should go in a folder on it's own.

MattElcock (commented):
Cheers for making those changed. Happy to tick when that other bug is sorted.

tbaggaley:
54om4n

eq-author/src/components/PositionModal/index.js
```diff
@@ -32,64 +32,126 @@ const Trigger = styled.button.attrs({ type: "button" })`
...
+      positionCalculation = sameMove > 0 ? sameMove : 0;
+    } else if (differentSection || fromFolderToFolder || inOrOut) {
+      positionCalculation = folderMove;
+    } else {
+      const selectedItemPosition = orderedOptions.findIndex(
+        ({ id }) => id === selected.id
+      );
+      positionCalculation = selectedItemPosition;
+    }
+
```
tbaggaley, line 113: Could you add a couple wee comments to this bit to explain what the logic behind all this arithmetic is? Could save Shane a headache when it comes to adding in moving folders ;-)

tjamesmac: Will do, I mean it looks self explanatory to me... ;)

tjamesmac: this i solved this

farres1 (dismissed):

eq-author/src/App/page/Design/MovePageModal/index.test.js
```diff
@@ -89,3 +99,72 @@ describe("MovePageModal: questionnaire not loaded", () => {
...
+    expect(output.filter(({ id }) => id.startsWith("folder-1-page"))).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          parentEnabled: true,
+          parentId: output[0].id,
+          position: expect.any(Number),
+        }),
+      ])
+    );
+  });
```
MattElcock, line 164:
```suggestion
  it("should include parent details on enabled folders", () => {
  
    const enabledFolders = output.filter(({ id }) => id.startsWith("folder-1-page"));
    expect(enabledFolders).toEqual(
      expect.arrayContaining([
        expect.objectContaining({
          parentEnabled: true,
          parentId: output[0].id,
          position: expect.any(Number),
        }),
      ])
    );
  });
```

The assertion of this test is slightly longer than the last, so you could toss the enabled folders into a variable to make it a little nicer on the eye.

eq-author/src/components/PositionModal/index.js
```diff
@@ -32,64 +32,117 @@ const Trigger = styled.button.attrs({ type: "button" })`
...
+  const positionButtonId = uniqueId("PositionModal");
   const [isOpen, setIsOpen] = useState(false);
-  const [selectedPosition, setSelectedPosition] = useState(previousPosition);
-  const [positionButtonId] = useState(uniqueId("PositionModal"));
+  const previousIndex = options.findIndex(({ id }) => id === selected.id);
+  const previousPosition = useRef(previousIndex > -1 ? previousIndex : 0);
+  const [{ position, item }, setOption] = useState({
+    position: previousPosition.current,
+    item: options[previousPosition.current],
+  });
```
MattElcock, line 47: I'm wondering if the variable names could be improved here. For instance:

`previousPosition` suggests it contains an integer, but actually it's the previous item in the list (I think). You could consider changing it to something like `previousItemRef`.

`previousIndex` becomes `previousItemIndex`.

```suggestion
  const previousItemIndex = options.findIndex(({ id }) => id === selected.id);
  const previousItemRef = useRef(previousItemIndex > -1 ? previousItemIndex : 0);
  const [{ position, item }, setOption] = useState({
    position: previousItemRef.current,
    item: options[previousItemRef.current],
  });
```

eq-author/src/components/PositionModal/index.js
```diff
@@ -32,64 +32,117 @@ const Trigger = styled.button.attrs({ type: "button" })`
...
-  const handleOpen = () => setIsOpen(true);
-
-  const handleChange = ({ value }) => setSelectedPosition(parseInt(value, 10));
+  const handleChange = ({ value }) => {
+    const count =
+      orderedOptions[value].__typename === "Folder" && value - position >= 0
+        ? orderedOptions.filter(
+            ({ parentId }) => parentId === orderedOptions[value].id
+          ).length
+        : 0;
```
MattElcock, line 74: You could split this up a bit to improve readability.

```suggestion
    const filteredOptions = orderedOptions.filter(
            ({ parentId }) => parentId === orderedOptions[value].id
          );
          
    const count =
      orderedOptions[value].__typename === "Folder" && value - position >= 0
        ? filteredOptions.length : 0;
```

MattElcock (approved):

eq-author/src/components/PositionModal/index.js
tjamesmac: previousPosition is an integer, `findIndex` returns a number

farres1 (approved):

jasehumphr (approved):

MattElcock (approved):

PR #1876, opened 2021-03-11
EAR-1220 Fix the qCode page

tjamesmac:
Still missing a few things from the qCode page

This ticket was initially made for the shortcode. But there was a few other things

### How to review
- Create a questionnaire
- Add every answer type 
- Date ranges should have two qcodes options
- everything else should have one
- multiple questions on one page should be 'nested'
- radio options should have one option (binary choice)
- mutually exclusive options should appear separate from other options


jasehumphr (approved):
More Ronseal

MattElcock (approved):

PR #1891, opened 2021-03-16
EAR-994 add focus to Create

zimboflyman:

**Actual**
When creating a new questionnaire, if you tab down to the 'create' button no focus styling appears around the button.

**Expected**
**When creating a new questionnaire, if you tab down to the 'create' button, focus styling should appear

### How to review

tab to cancel and create buttons to make sure focus is working
make sure you have a questionnaire name and type in order to activate the 'create' button


jasehumphr (approved):
Works innit

MattElcock (approved):

PR #1908, opened 2021-03-23
EAR-1236 Fix for move page

martyncolmer:
When moving the last page of a section into a different section the menu doesn't refresh with the new question in the old section.
This fixes that.

To Test- 
Spin it up and try


jasehumphr (approved):
LUSH

MattElcock (approved):

PR #1909, opened 2021-03-23
Ear 928 refactor resolvers

zimboflyman:

Right now, the bulk of the resolver utils live within a single file (`schema/resolvers/utils.js`) and the odd few are scattered around the codebase (e.g. create section is inside `schema/resolvers/base.js`).

This isn't ideal as `utils.js` is very big and difficult to maintain as more functions need to be added or modified. Additionally, having some rogue functions throughout the codebase which should live with the rest of the utils is making the codebase untidy.

This task is to tidy up the utils and, if possible, modularise them into groupings of related utils. For example, ones related to sections (e.g. getters and create section) would be within their own file.

### How to review

make sure everything still works!!!!


zimboflyman:
experiment

PR #1913, opened 2021-03-24
EAR-928 moved create methods into business logic

martyncolmer:
This PR is to refactor the API.
To test -
Spin it up.
Test everything still works.


jasehumphr (approved):
Lovely

farres1:
The code's looking really good! I've left some bugs I found when testing it

1 - Creating a folder does not work
2 - Duplicating a page does not work
3 - Duplicating a question does not work
4 - Duplicating a folder does not work
5 - Moving a question does not work

farres1 (approved):

PR #1919, opened 2021-03-30
Ear 1208 - Add previewTheme support to publisher v2

tbaggaley:

[Ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-1208) 

Publisher will now use the questionnaire's defined `previewTheme` to look up a theme in the `themes` array to use when generating the runner JSON.

The runner `eq_id`, `form_type`, `theme` and `legal_basis` will be set according to the theme referenced by `previewTheme` in the Author JSON.

The publisher changes provide defaults if previewTheme etc. are missing to match existing behaviour (for unmigrated surveys) - can potentially strip out once we're happy the migrations are working as intended.

Also fixes a couple of oversights on the API ticket re: social surveys:
- Social surveys now get the `social` theme by default when created
- The migration now gives social surveys the `social` theme as their default theme

### How to review

- Check publisher still works for existing questionnaires
- Check business questionnaires get `default` as their previewTheme / only theme in the `themes` array of the Author JSON
- Check social questionnaires get `social` as their previewTheme / only theme in the `themes` array of the Author JSON
- If you're feeling adventurous try adding themes / updating previewTheme using the mutations and see if the publisher output does the right thing
- Check out v3 PR too - [https://github.com/ONSdigital/eq-publisher-v3/pull/21](https://github.com/ONSdigital/eq-publisher-v3/pull/21)

### DIFF
--- a/eq-publisher/src/eq_schema/Questionnaire.test.js
+++ b/eq-publisher/src/eq_schema/Questionnaire.test.js
@@ -1,9 +1,10 @@
 const { last } = require("lodash");
 
 const { BUSINESS, SOCIAL } = require("../constants/questionnaireTypes");
+const legalBases = require("../constants/legalBases");
 const {
   types: { NOTICE_1, VOLUNTARY },
-} = require("../constants/legalBases");
+} = legalBases;
 
 const Questionnaire = require("./Questionnaire");
 const Summary = require("./block-types/Summary");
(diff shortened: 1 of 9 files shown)


eq-publisher/src/eq_schema/Questionnaire.test.js
```diff
@@ -78,7 +79,7 @@ describe("Questionnaire", () => {
     expect(questionnaire.legal_basis).toEqual(undefined);
   });
 
-  it("should set the theme based on the type", () => {
```
tjamesmac, line 81: t th the them theme?

tjamesmac (approved):

PR #1943, opened 2021-04-07
✨ EAR 1249 - Add API for locking questionnaires

tbaggaley:

[Ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-1249)

Adds mutation:

`setQuestionnaireLocked` : `{ questionnaireId, locked }` ->
  locks / unlocks questionnaire with given ID based on value of `locked`

Adds field:

`locked` :
  Boolean value determining whether current questionnaire is locked /
  unlocked

Adds migration to add `locked` field to existing questionnaires,
defaulting to `false`.

Prevent modifying questionnaires which are locked by making
`createMutation` check for questionnaire's locked status. Can be
overriden through passing in new `options` object with truthy `ignoreLockStatus`
to `createMutation` - needed so that the locking / unlocking mutation can still operate on
locked questionnaires.

Prevent modifying comments of questionnaires which are locked per AC.

Duplicating a locked questionnaire results in the duplicate being
unlocked per AC.

Have tested using dev tools and all appears to work as intended

### How to review

- Create a questionnaire
- Attempt to lock it using the `setQuestionnaireLocked` mutation. Can use this mutation in the Apollo dev tools:

`mutation Lock($input: SetQuestionnaireLockedInput!) {
  setQuestionnaireLocked(input: $input) {
    id
    locked
  }
}
`

For variables, use:

`{
  "input": {
    "questionnaireId": "<your-questionnaire-id>",
    "locked": true
  }
}`

- No changes should save to the questionnaire, no comments should save or be editable
- Unlock it by running the above mutation with `"locked": false` in your `input` object.
- Should be able to edit as normal

### DIFF
--- a/eq-author-api/schema/resolvers/withPermissions.js
+++ b/eq-author-api/schema/resolvers/withPermissions.js
@@ -2,7 +2,7 @@ const { ForbiddenError } = require("apollo-server-express");
 
 const hasWritePermission = (questionnaire, user) =>
   questionnaire.createdBy === user.id ||
-  questionnaire.editors.indexOf(user.id) > -1 ||
+  questionnaire.editors.includes(user.id) ||
   user.admin === true;
 
 const enforceHasWritePermission = (questionnaire, user) => {
--- a/eq-author-api/schema/resolvers/withPermissions.js
+++ b/eq-author-api/schema/resolvers/withPermissions.js
@@ -1,3 +1,5 @@
+const { ForbiddenError } = require("apollo-server-express");
+
 const hasWritePermission = (questionnaire, user) =>
   questionnaire.createdBy === user.id ||
   questionnaire.editors.indexOf(user.id) > -1 ||
@@ -5,15 +7,21 @@ const hasWritePermission = (questionnaire, user) =>
 
 const enforceHasWritePermission = (questionnaire, user) => {
   if (!hasWritePermission(questionnaire, user)) {
-    throw new Error(
+    throw new ForbiddenError(
       "User does not have write permission for this questionnaire"
     );
   }
 };
 
+const enforceQuestionnaireLocking = (questionnaire) => {
+  if (questionnaire.locked) {
+    throw new ForbiddenError("Questionnaire is locked");
+  }
+};
+
 const enforceHasAdminPermission = (user) => {
   if (!user.admin) {
-    throw new Error("User does not have admin permission");
+    throw new ForbiddenError("User does not have admin permission");
   }
 };
 
@@ -21,4 +29,5 @@ module.exports = {
   hasWritePermission,
   enforceHasWritePermission,
   enforceHasAdminPermission,
+  enforceQuestionnaireLocking,
 };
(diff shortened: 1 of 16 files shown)


eq-author-api/schema/tests/comments.test.js
```diff
@@ -152,6 +151,22 @@ describe("comments", () => {
...
+          { ...ctx, questionnaire: { ...ctx.questionnaire, locked: true } },
+          {
+            componentId,
+            commentText: "a new comment is created",
+          }
+        )
+      ).rejects.toThrow();
+
+      questionnaire.locked = false;
+    });
```
MattElcock, line 167: You could enhance this test by checking that the type of error thrown and message it's thrown with is what is expected.

eq-author-api/schema/tests/questionnaire.test.js
```diff
@@ -237,6 +238,49 @@ describe("questionnaire", () => {
...
+          },
+          ctx
+        );
+
+        expect(
+          updateSurveyId({
+            questionnaireId: ctx.questionnaire.id,
+            surveyId: "1337",
+          })
+        ).rejects.toThrow();
```
MattElcock, line 280: You could enhance this test, too, by asserting that the right error is thrown.

eq-author-api/schema/tests/questionnaire.test.js
```diff
@@ -658,6 +702,19 @@ describe("questionnaire", () => {
...
+      await setQuestionnaireLocked(
+        {
+          questionnaireId: ctx.questionnaire.id,
+          locked: true,
+        },
+        ctx
+      );
+
+      expect(deleteQuestionnaire(ctx, ctx.questionnaire.id)).rejects.toThrow();
+    });
```
MattElcock, line 717: You could enhance this test, too, by asserting that the right error is thrown.

MattElcock (dismissed):
Looks good to me; nice one, Tom. I've left a few comments but you don't have to do them.

eq-author-api/schema/tests/questionnaire.test.js
tbaggaley: Aye, cheers Matt - I did try and have them assert for a `ForbiddenError` but it doesn't actually percolate through to the tests - our (or Apollo's?) wrapper which executes the mutations for the tests always returns a plain old `Error` exception on failure rather the one I actually threw, so not much point checking for that.

eq-author-api/schema/resolvers/withPermissions.js
```diff
@@ -1,24 +1,33 @@
+const { ForbiddenError } = require("apollo-server-express");
+
 const hasWritePermission = (questionnaire, user) =>
   questionnaire.createdBy === user.id ||
   questionnaire.editors.indexOf(user.id) > -1 ||
```
tjamesmac, line 5:
```suggestion
  questionnaire.editors.includes(user.id) ||
```
?

tjamesmac (approved):
That's cool

eq-author-api/schema/resolvers/withPermissions.js
tbaggaley: Aye, why not ;-)

MattElcock (approved):

tjamesmac (approved):

PR #1948, opened 2021-04-08
Ear 1005 move a folder

tjamesmac:
[Prototype](https://overflow.io/s/FEJC8QOL?node=0f85b6ef&prototype) | [Ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-1005)

TODO

- [x] Use position modal for section swapper
- [x] fix folder moving
- [x] switch folders between sections (currently doesn't work)
- [x] tests
- [x] make sure the rest of the functionality works as normal

Allows users to move folders within and between sections

### How to review
- Move a folder
- Move a folder into a new section
- Move the last folder in a section to another section


farres1:
This is looking really good! I've added some changes here.

1 - After moving a folder, the modal does not close automatically.
2 - When you select a question/folder to move, it shows the name of what is being moved instead of the select text. This change might be intentional but I added it in case it isn't.

jasehumphr (approved):
Awesome. Works as expected

farres1 (approved):

PR #1959, opened 2021-04-12
✨ EAR 1250 & 1251 - Add locking to UI

tbaggaley:

Allows locking and unlocking questionnaires to prevent changes.

Implements [EAR-1250](https://collaborate2.ons.gov.uk/jira/browse/EAR-1250) (add lock to home page) and [EAR-1251](https://collaborate2.ons.gov.uk/jira/browse/EAR-1251) (update permissions banner).

#### Change overview
- Adds lock button to home page ( questionnaires view )
- Adds new `lockStatusUpdated` subscriptions
- Re-implement `setQuestionnaireLockStatus` resolver to avoid using `createMutation` (since it needs to work when there is no questionnaire in the `ctx` - e.g. on home page)
- Adds lock subscriptions to home page & QuestionnaireDesignPage
- Adds new locked warning banner to PermissionsBanner with Unlock button
- Re-order columns on home page per new design choices
- Adds ability to use hook wrapper around ConfirmationModal
- Adds QuestionnaireLockingModal using hook wrapper
- Re-factor QuestionnaireDesignPage, move out existing subscriptions, purge all remaining HOCs
- Add / update tests for coverage

### How to review

See the list of AC on the [main ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-849).

Rough summary:

- Should be able to lock / unlock from home page (with modal showing "lock" vs "unlock" appropriately)
- Icons should update appropriately
- Should see warning banner on locked questionnaires
- Should be able to unlock questionnaires from locked warning banner
- Changes to locked questionnaires / comments should not be saved
- If you lock a questionnaire from the homepage while looking at it (e.g. in another window / tab), the warning banner should appear when locked without having to refresh the page

#### Home page, confirmation modal
Screenshot 2021-04-13 at 07 55 54

#### Locked warning banner
Screenshot 2021-04-13 at 07 56 23


eq-author/src/components/modals/ConfirmationModal/index.js
```diff
@@ -73,16 +77,42 @@ const ConfirmationModal = ({
...
 
+export const useConfirmationModal = ({ action, ...props }) => {
+  const [isOpen, setIsOpen] = useState(false);
+
+  const handleConfirm = useCallback(() => {
+    action();
+    setIsOpen(false);
+  }, [action]);
+
+  const handleCancel = useCallback(() => setIsOpen(false));
```
tjamesmac, line 96:
```suggestion
  const handleCancel = useCallback(() => setIsOpen(false), []);
```
It never needs to change because it's using a static value or you want it to change everytime?

eq-author/src/components/modals/ConfirmationModal/index.js
```diff
@@ -73,16 +77,42 @@ const ConfirmationModal = ({
...
+
+  return {
+    trigger: useCallback(() => setIsOpen(true)),
```
tjamesmac, line 99: Same here

tbaggaley: Aye, I assumed no second param would imply [] but it seems not. Cheers! Will push in a sec.

tbaggaley:
Lock button is now disabled when you don't have write permissions for a questionnaire

martyncolmer (approved):

farres1 (commented):
Locking is looking really good!

The only change I found is that the title column on the home page is indented slightly more right on the designs/previous versions.

tbaggaley:
Cheers Sam - indented the title column a bit more. See what you think

martyncolmer (approved):

martyncolmer (dismissed):

eq-author/src/components/BaseLayout/PermissionsBanner.js
```diff
@@ -6,18 +6,30 @@ import PropTypes from "prop-types";
...
   display: flex;
 `;
 
+const UnlockButton = styled(Button).attrs(() => ({
+  small: true,
+  variant: "tertiary-light",
+}))`
+  margin-left: 1em;
+  padding: 0.25em 0.5em;
+  border: 1px solid white;
```
MattElcock, line 30:
```suggestion
  border: 1px solid ${colors.white};
```

MattElcock (approved):
Looks good, Tom. I especially like the way you've made the modal into a hook rather than a standard component. It makes handling the state of that component a lot easier when other things influence it. 

Would you mind demoing this and explaining the benefits and drawbacks of this method at the next tech session?

I've approved this as the suggestion I've made is really minor and doesn't need to block the ticket if you're wanting to get it out fast.

MattElcock (approved):

eq-author/src/components/BaseLayout/PermissionsBanner.js
tbaggaley: Cheers Matt - have pushed colour constant change.

And aye, thought it was kind of neat to hide away all the other stuff that otherwise leaks out (e.g. parents have to manually open / close modals and maintain their state even though it's really not part of their jurisdiction). Can discuss at tech session sure.

Would be cool if we could make a generic hook in that style for all similar modals - would just have to standardise the prop names so it's always `onConfirm` and `onCancel` etc.

martyncolmer (approved):

MattElcock (approved):

martyncolmer (approved):

PR #1968, opened 2021-04-16
🛠 EAR 1227 - Validate routing destinations

tbaggaley:

[Ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-1227)

Fix validation for routing destinations - make else destination
validation consistent with routing rules' destination validation

Remove 'Default' destination type and use `null` instead to signify
empty destination in need of selection.

Don't delete answer IDs from routing rules when question pages / sections deleted -
use AJV validation rules instead to check if answers exist and report
accordingly

Move out destination validation in AJV schema so it can be re-used across else +
routing rules - add `validationErrorInfo` resolver to `Destination2` type

Delete redundant validation code in `updateRoutingRule2` and `updateRouting2`
mutations - should all now be handled by AJV schema; cases previously caught & thrown exceptions
for will all cause validation errors for clients

Factor out common destination updating logic into new businessLogic file `updateDestination.js`

### How to review

- Check both `else` and routing rule destination selectors have "moved" and "deleted" validation errors appropriately when you move / delete target question pages
- Same but with a section as the target
- Check I ain’t broke other validation for routing (e.g. on expressions themselves)


MattElcock (approved):

MattElcock:
Looks great and works like a dream, @tbaggaley ! Well done. I like the refactoring you've done to the big switch statement; a lot nicer now.

tjamesmac (approved):

farres1 (approved):

PR #1973, opened 2021-04-19
Ear 1033 validate other label radio and checkbox

zimboflyman:
Finally....

**_Empty "Other label" field_**

GIVEN I've added an "Other" option on a radio answer
WHEN the "Other label" field is empty
THEN I see a validation message on the "Other label" field which reads "Enter a label"

GIVEN I've added an "Other" option on a checkbox answer
WHEN the "Other label" field is empty
THEN I see a validation message on the "Other label" field which reads "Enter a label"

_**Not empty "Other label" field**_

GIVEN I've added an "Other" option on a radio answer
WHEN the "Other label" field is not empty
THEN I see no validation message on the "Other label" field

GIVEN I've added an "Other" option on a checkbox answer
WHEN the "Other label" field is not empty
THEN I see no validation message on the "Other label" field

### How to review

create a checkbox answer and select "add checkbox - other option" to add a checkbox with other option
make sure the additional 'other label' has working validation with correct message

do the same for Radio answer


eq-author/src/App/page/Design/answers/MultipleChoiceAnswer/MultipleChoiceAnswer.test.js
```diff
@@ -125,27 +125,27 @@ describe("MultipleChoiceAnswer", () => {
...
+
+  //   const Options = wrapper.find(Reorder).renderProp("children")(
+  //     extraProps,
+  //     option
+  //   );
+
+  //   expect(
+  //     Options.find("Field__StyledField")
+  //   ).toHaveLength(1);
+  // });
```
MattElcock, line 148:
```suggestion
```

MattElcock (changes requested):

MattElcock:
image

The distance between the error message and the bottom of the component needs to be increased a smidge, I think.

eq-author/src/App/page/Design/answers/MultipleChoiceAnswer/MultipleChoiceAnswer.test.js
zimboflyman: woops - now sorted

MattElcock (dismissed):
Hi Shane, well done on this. Thanks, too, for modernising the class component. It's good to see that we're making these improvements when we come across them.

Everything looks nice, tidy, and works as expected. :)

farres1 (commented):
The validation is really good!

I agree with Matt that the error message should be slightly higher. 

Also, the validation error message for unique radio/checkboxes no longer works. The error badge is still displayed, but the error message is not shown.

.gitignore
```diff
@@ -35,6 +35,8 @@ graphql.config.json
 public/status.json
 .env
 .env.*local
+env.test
+env.*local
```
tbaggaley, line 39:
```suggestion
```

Haven't looked properly yet but can we get rid of these from the `.gitignore`? :-)

tbaggaley (changes requested):

farres1 (approved):

MattElcock (approved):

PR #1991, opened 2021-04-26
🛠 EAR 1161 - Fix date range piping validation

tbaggaley:

[Ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-1161)

Fix piping of date range answers in question titles.

Were previously showing as always "deleted" due to the way in which
`from` and `to` are always appended to their IDs by the piping code
in order to differentiate their from vs. to dates, which were
non-existent IDs.

#### Changes:
- Strip "from" and "to" postfix in validation code when checking if answer exists
- Add test to check valid date range piping doesn't cause errors
- Rejig `validatePipingInTitle` custom keyword to use `idExists` and `getAbsolutePositionById`

Also took the opportunity to delete some flotsam I missed in another
PR (`getNextDestination` in `businessLogic` - no longer used by API)

### How to review

- Pipe a date range answer (to and/or from) into another question title
- Move the date range answer after the question with piping - should see the "moved" error
- Delete the date range answer - should see the "deleted" error


MattElcock (approved):
Looks good, Tom! Thanks for sorting this. Nice one on refactoring what was already there at the same time, too, and cutting the length of code in half. 🎉🎉

farres1 (approved):

PR #2016, opened 2021-05-07
✨ EAR 1315 - Add API resolver to import questions from another questionnaire

tbaggaley:

[Ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-1315)

Adds `importQuestions` resolver  which imports questions specified by `questionIds` from questionnaire with ID `questionnaireId` at the provided `Position` (specified by `folderId`, `sectionId`, and the `index` to which the items should be inserted from).

Resolver mimics `movePage` position handling in that you can specify either a `folderId` or `sectionId`, `folderId` (being more specific) takes higher precedence. The `index` specifies the position to start inserting from in *either* a section (if `sectionId` and not `folderId` provided) or folder (if `folderId` provided).

`UserInputErrors` are thrown for any possible failure conditions to do with invalid input.

The input questions are mapped to imported questions like so:
- All IDs are remapped
- All nested `qCode` and `secondaryQCode` attributes are set to `null`
- Piping is *not* deleted for now after conversation with Joe - this will preserve the `piping deleted` errors when questions are imported (since IDs will not be resolvable within the questionnaire) for purpose of the initial MVP
- If the target is a `Section`, each imported question is wrapped in its own disabled folder per ticket spec.

The resolver returns `Section` for now (the section modified in the target questionnaire) for ease of integrating the response on the client, e.g. utilising automatic cache updates (idiomatic graphql would probably be to return only the folder(s) of imported content, requiring manual cache writes - can do that instead if you want).

### How to review

- Make a source questionnaire with a set of `questionIds` - make a note of these
- Make a target questionnaire
- While on the target questionnaire:
- Use Apollo dev tools to execute a query like the following:

```graphql
mutation ImportStuff($input: ImportQuestionsInput!) {
    importQuestions(input: $input) {
      id
      <...whatever else you want from Section>
   }
})
```

and as input:

```graphql
input: {
  questionIds: [<question IDs from source>],
  questionnaireId: <source questionnaire ID>,
  position: {
    sectionId: <target section ID in current questionnaire>,
    index: <index you'd like questions to be inserted into in target section>
  }
}
```
- Should get the output you'd expect from the above
- Refresh the page & you should see the new imported content
- qCodes should be empty for imported stuff
- If you want, try the same thing but with a target `folderId` instead


eq-author-api/schema/resolvers/importing.js
```diff
@@ -0,0 +1,88 @@
...
+          questionIds
+        );
+        if (pages.length !== questionIds.length) {
+          throw new UserInputError(
+            `Not all page IDs in [${questionIds}] exist in source questionnaire ${questionnaireId}.`
+          );
+        }
+
+        // Re-create UUIDs, strip QCodes, routing and skip conditions from imported pages
+        // Keep piping intact for now - will show "[Deleted answer]" to users when piped ID not resolvable
```
MattElcock, line 47: It should strip out piping; the frontend UI will inform the user that piping won't be imported. RAD agreed that we'll address piping in the next iteration.

If you think doing this will be more work than what you have now, can you bring it up with the RAD team and then update the ticket with the result, please? We'll then also need to update the design to change the guidance we give to users.

eq-author-api/schema/resolvers/importing.js
```diff
@@ -0,0 +1,88 @@
...
+          section = getSectionByFolderId(ctx, folderId);
+        } else {
+          section = getSectionById(ctx, sectionId);
+          if (!section) {
+            throw new UserInputError(
+              `Section with ID ${sectionId} doesn't exist in target questionnaire.`
+            );
+          }
+
+          // Insert each imported page into its own disabled folder per EAR-1315
```
MattElcock, line 76:
```suggestion
```

You can probably remove this comment as the code is fairly simple to understand.

MattElcock (changes requested):

MattElcock (approved):
Hey Tom. If you saw my earlier comments, ignore them -- I had another look and changed my mind (and also realised one of them was just wrong, lol). 

This all looks really good, thanks for doing this. The code reads well and is nicely formatted. Well done.

zimboflyman (approved):

PR #2017, opened 2021-05-07
Ear 1275 reusable questionnaire table

farres1:

Reusable table component
The table's required columns are set using the enabledRows list in the QuestionnairesTable file
Removing an item from the list will remove the column from the table
Only items in the list are displayed - if a column is not required, excluding it from the list prevent it from being displayed  

### How to review

Check:
- [ ] The questionnaire table is displayed with the correct content and headings
- [ ] All questionnaires can be selected in the table
- [ ] Data can be sorted in the questionnaires table
- [ ] Removing an item from the enabledRows list removes it from the questionnaire table


eq-author/src/App/QuestionnairesPage/QuestionnairesView/QuestionnairesTable/index.js
```diff
@@ -1,24 +1,28 @@
...
+import { SORT_ORDER } from "constants/sort-order.js";
+import * as Headings from "constants/table-headings";
 
-import Row from "./Row";
-import TableHead from "./TableHead";
+import QuestionnaireTable from "components/QuestionnaireTable";
 import Panel from "components/Panel";
 
+import tableHeadings from "../../../../components/QuestionnaireTable/TableHeadings";
+
```
MattElcock, line 13: You shouldn't need to pass these in as a prop to the component now, as they live alongside it. Instead, you can import them directly into the component file.

eq-author/src/components/QuestionnaireTable/Row.js
```diff
@@ -12,15 +12,16 @@ import IconButtonDelete from "components/buttons/IconButtonDelete";
...
 
 import { colors } from "constants/theme";
 import { WRITE } from "constants/questionnaire-permissions";
-import Button from "components/buttons/Button";
+import * as Headings from "constants/table-headings";
 
-import FormattedDate from "./FormattedDate";
-import questionConfirmationIcon from "./icon-questionnaire.svg";
+import FormattedDate from "../../App/QuestionnairesPage/QuestionnairesView/QuestionnairesTable/Row/FormattedDate.js";
+import questionConfirmationIcon from "../../App/QuestionnairesPage/QuestionnairesView/QuestionnairesTable/Row/icon-questionnaire.svg";
```
MattElcock, line 24: Can you move this svg into the `assets` directory, please? If there's already one there, you can use that instead and delete the one living in `QuestionnairesTable/row`

See lines 25-27 for an example on how to import an SVG from the assets directory.

eq-author/src/components/QuestionnaireTable/Row.js
```diff
@@ -258,6 +261,118 @@ export const Row = ({
   const hasWritePermission = permission === WRITE;
   const canDelete = hasWritePermission && !locked;
 
+  const renderEnabled = (tableHeadings, columnHeading) =>
+    tableHeadings.map(({ heading, enabled }) => {
+      if (enabled) {
+        if (heading === columnHeading) {
```
MattElcock, line 267: These two statements could be combined to be:

```
if (enabled && heading === columnHeading) {
 . . .
}
```

eq-author/src/components/QuestionnaireTable/Row.js
```diff
@@ -258,6 +261,118 @@ export const Row = ({
...
+          if (heading === Headings.TITLE) {
```
MattElcock, line 268: As each if statement returns, you could swap the series of `if` statements to be a `switch` block.

```
switch (heading) {
   case Headings.TITLE:
      . . .
     break;
   case Headings.OWNER:
      . . .
     break;
   case Headings.CREATED:
     . . .
     break;
   . . .
   default:
     return null;
}
```

MattElcock (changes requested):

MattElcock (dismissed):
Thanks for making those changes, Sam. This looks great. Nice work.

eq-author/src/components/QuestionnaireTable/TableHead.js
```diff
@@ -107,61 +108,45 @@ const UnsortableTH = styled(TH)`
...
-        </SortableTH>
-        <UnsortableTH colWidth="10%">Permissions</UnsortableTH>
+function checkEnabled(enabledRows, headings) {
+  enabledRows.map((heading) => {
+    const headingElement = headings.find(
+      (element) => element.heading === heading
+    );
+    headingElement.enabled = true;
+    return null;
+  });
```
tbaggaley, line 118: May wish to use `forEach` instead for cases like this when you don't actually care about the return value (you're mapping everything => `null` here)

tbaggaley: Might also find it easier for the general flow of things to `filter` out the non-enabled headings instead of doing this and pass around only the headings you really want to be rendered - would also simplify your downstream code (no need to keep checking if stuff is enabled)

eq-author/src/components/QuestionnaireTable/Row.js
```diff
@@ -268,76 +377,14 @@ export const Row = ({
...
-          </div>
-        </TD>
+        {renderEnabled(tableHeadings, Headings.TITLE)}
+        {renderEnabled(tableHeadings, Headings.OWNER)}
+        {renderEnabled(tableHeadings, Headings.CREATED)}
+        {renderEnabled(tableHeadings, Headings.MODIFIED)}
+        {renderEnabled(tableHeadings, Headings.PERMISSIONS)}
+        {renderEnabled(tableHeadings, Headings.LOCKED)}
+        {renderEnabled(tableHeadings, Headings.STARRED)}
+        {renderEnabled(tableHeadings, Headings.ACTIONS)}
```
tbaggaley, line 387: If you did filter out stuff (see following comment), then this would simplify to a simple `map`. Could probably replace it with iteration even if you didn't, though (it's doing the same-ish thing 8 times)

tbaggaley (changes requested):
Hey Sam, just a few general comments on this after leafing through the code. Let me know if you disagree ;-) Left a few minor ideas re: code but have a couple of higher-level questions about the thing:

I wonder if there would be benefit in including some of the behaviours inside the "reusable component" too - e.g. the sorting functionality, etc.? Otherwise we'll have to re-implement sorting each time it gets re-used. Might be worth having a chat with Matt E or Martyn about to decide what they really want to get out of the component - if it's just a view then great, but otherwise as it stands we'd need to fetch the questionnaires list & provide our own logic for sorting, deleting, locking, etc. etc. Personally I think it'd be nicer if those things were included within the component (not personal criticism - this isn't mentioned in ticket AC).

One of the things we do need though is to have a way to find out when a questionnaire gets selected - since this is the purpose of this component being reusable within a modal. We could use a `onQuestionnaireSelected` callback or something exposed via props that the "import questions" modal can draw on - can't see anything like that at the mo, but apologies if it does exist in a different form and I've overlooked it.

The main `QuestionnairesPage` would then need to provide an `onQuestionnaireSelected` callback which actually opens the questionnaire (which at present is the onClick behaviour on a questionnaire row).

Hope that makes sense. Can discuss later if you want in more detail perhaps also w/ Matt E if that's helpful.

MattElcock (approved):

tbaggaley (approved):

PR #2019, opened 2021-05-07
Ear 1276 make question list reusable

zimboflyman:

Make the question List (currently used in Calc sum) a reusable component - ready for 'import Questions' ticket

### How to review

I've adapted Calsum answer picker to use this new component so you can test it there.
i've added props for the following:

1)title - title name to be displayed - required
2)showTypes - Show types in header
3)showSearch - Show a search bar

I've also updated the search bar (used in the Home page Questionnaire list) to be a seperate component
This takes the following porps:

1)size: large - otherwise will default to the smaller style
2)paddingType: large - otherwise will default to smaller padding

These props are to account for the different size search bars in the Questionnaire list on home page and the List in the new import modals


MattElcock (approved):
Hi Shane, this looks great. Thanks for making the search bar reusable as well.

jasehumphr (approved):

PR #2032, opened 2021-05-11
EAR 1207 - ✨ Add validation error when all themes disabled

tbaggaley:

[Ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-1207)

Adds validation errors when all themes are disabled.

Fairly significant re-jigging of themes design to make this possible / more streamlined:
- `themes` and `previewTheme` wrapped up in new GraphQL type `ThemeSettings` instead of being on root `Questionnaire` type, allowing adding a `validationErrorInfo` resolver at that level (for group errors rather than individual theme level errors)
- Author API now returns all themes and their enabled/disabled statuses - but does *not* persist these into the JSON unless the user has enabled them; a hybrid approach between "return everything" and "just save what you need" which greatly simplifies the UI code - most of the gnarly bits of `ThemesPage` are now gone (no need to manually update cache) at the expense of a small increase of complexity on the API side

Misc changes:
- Don't store `questionnaireId` in `Theme` as per [this](https://surveydatacollection.slack.com/archives/C0101NY9PN0/p1620822255033500).

Most of the scaffolding for adding validation errors to individual `Theme`s is also constructed here - should be fairly trivial to add validation for individual `eqId`s, etc. as well as group-level errors (by adding to the JSON schema for `theme` or `themeSettings` respectively).

### How to review

- [ ] Open a Q
- [ ] Check Themes page works as before - should be able to enable, disable, update themes and survey ID, etc.
- [ ] Check validation error appears with all themes disabled
- [ ] Check view survey button disables itself when validation error present
- [ ] Check red dot appears in themes tab bar
- [ ] Check red dot appears in Settings button on main navigation pane
- [ ] Check publisher still works as before

### Behold the dots, the errors!

Screenshot 2021-05-12 at 12 31 30


eq-author/src/App/settings/ThemesPage.js
```diff
@@ -87,7 +91,7 @@ const EqIdInput = ({ eqId = "", questionnaireId, shortName }) => {
     value = value.trim();
     updateQuestionnaireTheme({
       variables: {
-        input: { questionnaireId, shortName, eqId: value },
+        input: { questionnaireId, shortName, eqId: value.trim() },
```
MattElcock, line 94: There's already a `value.trim()` a couple of lines above, so you don't need to do it again here.

```suggestion
        input: { questionnaireId, shortName, eqId: value },
```

MattElcock (changes requested):

eq-author/src/App/settings/ThemesPage.js
tbaggaley: Haha, oops. I thought I'd deleted the `value = value.trim()` line. Will push in a sec, cheers.

MattElcock (approved):

jasehumphr (approved):

PR #2035, opened 2021-05-13
Ear 1305 percentage front end

zimboflyman:

get the front end ready according to the prototype
https://overflow.io/s/DJOA8OQS?node=b4da5dab&prototype&view=fill

### How to review

**Acceptance criteria**
GIVEN I'm on a question page
WHEN I add a percentage answer
THEN I can add an "Or" option to the percentage answer

**Disable "Add another answer" button if there is an "or" option on the page**

GIVEN I'm on a question page
WHEN an "or" option is toggled on for any answer
THEN the "Add another answer" button is disabled

**Disable "Or" option toggle switch if there is more than one answer on a page**

GIVEN I'm on a question page
WHEN there is more than one answer on the page
THEN the "Or" option toggle is disabled


eq-author/src/App/page/Design/QuestionPageEditor/AnswersEditor/index.js
```diff
@@ -23,6 +23,10 @@ export const AnswersEditor = ({
   onDeleteAnswer,
   moveAnswer,
 }) => {
+
+    let multipleAnswers = false;
+    answers?.length > 1 ? multipleAnswers = true : multipleAnswers =false;
```
MattElcock, line 28:
```suggestion
    const multipleAnswers = answers?.length > 1 ? true : false;
```

eq-author/src/App/page/Design/answers/BasicAnswer/index.js
```diff
@@ -15,9 +26,24 @@ import ValidationErrorInfoFragment from "graphql/fragments/validationErrorInfo.g
...
+  margin-bottom: 0.4em;
+
+  > * {
+    margin-bottom: 0;
+  }
+`;
+
+const ToggleWrapper = styled.div`
+  opacity: ${(props) => (props.disabled ? "0.6" : "1")};
+  pointer-events: ${(props) => (props.disabled ? "none" : "auto")};
```
MattElcock, line 45:
```suggestion
  opacity: ${({disabled}) => (disabled ? "0.6" : "1")};
  pointer-events: ${({disabled}) => (disabled ? "none" : "auto")};
```

MattElcock (approved):
Hi Shane, this looks really good. Well done; I like how you tied it to the answer lable + description to demonstrate it working. 

There's two comments, small refactoring suggestions, but you don't need to implement them; they're just a few things I noticed as I was reading through, but what you have is perfect.

eq-author/src/App/page/Design/QuestionPageEditor/AnswersEditor/index.js
zimboflyman: Ahh yes - nice and tidy - thanks!

eq-author/src/App/page/Design/answers/BasicAnswer/index.js
zimboflyman: coolio - makes sense - thanks Matt!

eq-author/src/App/page/Design/QuestionPageEditor/AnswersEditor/index.js
zimboflyman: actually @MattElcock  this looks even cleaner!
const multipleAnswers = answers?.length > 1;
otherwise the linter jumps up and down!!!😁

zimboflyman: mmm - not so clever - now multipleAnswers won't exist as a required prop - I'm gonna settle on this...
let multipleAnswers = false;
multipleAnswers = answers?.length > 1;

eq-author/src/App/page/Design/QuestionPageEditor/AnswersEditor/index.js
```diff
@@ -23,6 +23,10 @@ export const AnswersEditor = ({
   onDeleteAnswer,
   moveAnswer,
 }) => {
+
+  let multipleAnswers = false;
+  multipleAnswers = answers?.length > 1;
```
tbaggaley, line 28:
```suggestion
const multipleAnswers = answers?.length > 1;
```

🎬 Take 3 ;-)

eq-author/src/App/page/Design/answers/BasicAnswer/index.js
```diff
@@ -34,9 +60,21 @@ export const StatelessBasicAnswer = ({
...
+  multipleAnswers,
 }) => {
+  let [toggled, setToggled] = useState(false);
   const errorMsg = buildLabelError(MISSING_LABEL, `${lowerCase(type)}`, 8, 7);
 
+  useEffect(() => {
+    if (multipleAnswers) {
+      setToggled(false);
+    }
+  }, [multipleAnswers]);
```
tbaggaley, line 72: What's the rationale behind this `useEffect`? Just curious

eq-author/src/App/page/Design/QuestionPageEditor/AnswersEditor/index.js
zimboflyman: If I do this @tbaggaley then multipleAnswers doesn't exist as a prop that gets passed down.
I guess I could make that prop not required, but seems like it really should be required

eq-author/src/App/page/Design/answers/BasicAnswer/index.js
zimboflyman: mmm - interesting question - initially the toggle refused to turn off without it, but now seems to work fine without it

eq-author/src/App/page/Design/QuestionPageEditor/AnswersEditor/index.js
tbaggaley: Yeah it doesn't affect any of your other code. You're just declaring & defining `multipleAnswers` and then sticking that straight away in a prop without any further processing

zimboflyman: const multipleAnswers = answers?.length > 1 ? true : false;
This would make sense but the linter gets sad ☹️

eq-author/src/App/page/Design/answers/BasicAnswer/index.js
zimboflyman: ahhh - thats it - now I remember -  if the toggle is ON and then you 'add another answer' - this is to re-render the toggle with updated props and hence turn off...

MattElcock (approved):

tbaggaley (approved):

PR #2045, opened 2021-05-20
Ear 1350 stitch backend with frontend mutually exclusive percentage

MattElcock:

Now that the back and front end are done, this ticket links them together.

### How to review

Make sure it works end-to-end, without Publisher.


eq-author/src/App/page/Design/answers/BasicAnswer/index.js
```diff
@@ -62,9 +68,28 @@ export const StatelessBasicAnswer = ({
...
+  const [createMutuallyExclusiveOption] = useMutation(
+    CREATE_MUTUALLY_EXCLUSIVE_OPTION
+  );
+  const [updateOption] = useMutation(UPDATE_OPTION_MUTATION);
+  const [deleteOption] = useMutation(DELETE_OPTION);
+
   const errorMsg = buildLabelError(MISSING_LABEL, `${lowerCase(type)}`, 8, 7);
 
+  useEffect(() => {
+    const { options } = answer;
```
zimboflyman, line 83: const { options } = answer;
little curious why this is not global  as it's defined in 3 places?

zimboflyman (commented):
Looks really nice and clean and defo looks like it will work great with all the basic answer types so should be easy to merge with all the new front end type work.

Just one tiny query below

eq-author/src/App/page/Design/answers/BasicAnswer/index.js
MattElcock: Honestly, there's no reason. I can take a look at making it global if you'd like? I'd be tempted to do something like this:

```
const [options, setOptions] = useState({});

useEffect(() => {
   setOptions(answer.options);
}, [answer])

useEffect(() => {
   const mutuallyExclusive = options?.find(
      ({ mutuallyExclusive }) => mutuallyExclusive === true
    );
    if (mutuallyExclusive) {
      setToggled(true);
      setMutuallyExclusiveLabel(mutuallyExclusive.label);
    }
}, [options])

. . .
```
The reason I'd do it like this is to ensure that things update and run only when the props update; they all depend on `answer` changing. At least, most of them do.

zimboflyman: no worries - looks more simple to declare locally!

zimboflyman (dismissed):

eq-author/src/App/page/Design/answers/BasicAnswer/index.js
```diff
@@ -62,9 +68,28 @@ export const StatelessBasicAnswer = ({
   optionErrorMsg,
   multipleAnswers,
 }) => {
-  let [toggled, setToggled] = useState(false);
+  const [toggled, setToggled] = useState(false);
+  const [mutuallyExclusiveLabel, setMutuallyExclusiveLabel] = useState();
```
tbaggaley, line 72: Would suggest not using state for this - this is stuff that should flow in from above when the parent re-renders post update (when Apollo result comes back) imo.

Toggle state would therefore be managed by whatever's set in the Apollo cache

eq-author/src/App/page/Design/answers/BasicAnswer/index.js
```diff
@@ -62,9 +68,28 @@ export const StatelessBasicAnswer = ({
...
   const errorMsg = buildLabelError(MISSING_LABEL, `${lowerCase(type)}`, 8, 7);
 
+  useEffect(() => {
+    const { options } = answer;
+    const mutuallyExclusive = options?.find(
+      ({ mutuallyExclusive }) => mutuallyExclusive === true
+    );
+    if (mutuallyExclusive) {
+      setToggled(true);
+      setMutuallyExclusiveLabel(mutuallyExclusive.label);
```
tbaggaley, line 89: Could then get rid of othis

eq-author/src/App/page/Design/answers/BasicAnswer/index.js
```diff
@@ -73,6 +98,29 @@ export const StatelessBasicAnswer = ({
...
+        variables: { input: { answerId: answer.id, label: "" } },
+      });
+    } else {
+      const { options } = answer;
+      const { id } = options?.find(
+        ({ mutuallyExclusive }) => mutuallyExclusive === true
+      );
+
+      deleteOption({ variables: { input: { id } } });
+      setMutuallyExclusiveLabel("");
```
tbaggaley, line 113: And this

eq-author/src/App/page/Design/answers/BasicAnswer/index.js
tbaggaley, line 89: Could then get rid of this

eq-author/src/App/page/Design/answers/BasicAnswer/index.js
MattElcock: Hey Tom. Thanks for the comments! 

I've actioned the one regarding the toggle state; that was indeed a bit overkill. Although, the label text needs to have a `useState` otherwise the user can't update the input box. It uses a `useEffect` to populate the state whenever the `answer` changes, so it's initial value is being handeled by the cache.

tbaggaley: Oops, aye, that makes sense re: the label if you're using a controlled input.

tbaggaley (approved):

zimboflyman (approved):

PR #2058, opened 2021-05-26
Ear 1234 validate form type

farres1:

Adds validation error when a form type of the incorrect format is entered

### How to review

**Check:**
- [ ] Entering a form type with a non-numeric character provides the correct error message
- [ ] Entering a form type with more than four numbers provides the correct error message
- [ ] Entering a form type with less than four numbers provides the correct error message
- [ ] Entering a form type with four numbers does not provide an error message
- [ ] A small error badge is displayed in the settings main navigation item when there is a form type validation error
- [ ] An error badge containing error count is displayed in the themes settings vertical tab when there is a form type validation error
- [ ] No error is displayed when form type input is empty
- [ ] The view survey button is enabled when there is a form type error and no other errors
- [ ] The view survey button is disabled when there is a form type error with other errors


eq-author/src/App/QuestionnaireDesignPage/MainNavigation/index.js
```diff
@@ -88,6 +89,7 @@ export const UnwrappedMainNavigation = ({
   useSubscription(publishStatusSubscription, {
     variables: { id: params.questionnaireId },
   });
+  const totalErrorCountNoTheme = totalErrorCount - themeErrorCount;
```
MattElcock, line 92: I'm not sure if we should make this a little more specific. 

If the user has turned off all themes there won't be a preview theme selected and therefore the View Survey button won't, or shouldn't, work.

But, if they haven't entered something like the form type or survey ID, then they should still be able to view.

I don't think this was in the ACs though, so it's probably something we can add on another iteration. 

What do you think?

eq-author/src/App/QuestionnaireDesignPage/index.js
```diff
@@ -73,6 +73,10 @@ export const QuestionnaireDesignPage = () => {
     data: { questionnaire } = {},
   } = useQuestionnaireQuery(questionnaireId);
 
+  const themeErrorCount = questionnaire?.themeSettings?.validationErrorInfo?.errors.filter(
```
MattElcock, line 76:
```suggestion
  const formTypeErrors = questionnaire?.themeSettings?.validationErrorInfo?.errors.filter(
```

Could this be renamed to make it a little clearer that it's just form type errors, rather than all theme errors, please?

eq-author/src/App/QuestionnaireDesignPage/index.js
```diff
@@ -102,6 +106,7 @@ export const QuestionnaireDesignPage = () => {
                           questionnaire?.themeSettings?.validationErrorInfo
                             ?.totalCount > 0
                         }
+                        themeErrorCount={themeErrorCount}
```
MattElcock, line 109:
```suggestion
                        themeErrorCount={formTypeErrors}
```

If you do the above.

eq-author/src/App/settings/ThemesPage.js
```diff
@@ -123,34 +123,47 @@ const ThemesPage = ({ questionnaire }) => {
...
 
+  const renderThemeErrors = (validationErrorInfo) =>
+    validationErrorInfo.errors.map(
+      ({ errorCode }, index) =>
+        errorCode === "ERR_FORM_TYPE_FORMAT" && (
+          <ValidationError key={index} right>
+            {THEME_ERROR_MESSAGES.ERR_FORM_TYPE_FORMAT}
+          </ValidationError>
+        )
+    );
```
MattElcock, line 134: I would encapsulate this into the `FormType` component, as it's looking for an error specifically to that.

Maybe something like:

```
const findErrorsByCodePrefix = ({errors}, prefix) => 
    errors.map( ({errorCode, ...rest}) => 
        errorCode.contains(prefix) ?? {errorCode, ...rest})
```

Then, when you're using the `FormType` component on the themes page:

```
<FormType
              formType={formType}
              questionnaireId={id}
              shortName={shortName}
              errors={findErrorsByCodePrefix('ERR_FORM_TYPE')}
/>
```

Then, in the `FormType` component, you'd have something akin to your current `renderThemeErrors` function that maps over the errors passed into the `FormType` component and renders the specific styling.

Alternatively, if you want to keep your current set up (which is totally vali as it ensures all errors on the themes page are standardised), you could modify your `renderThemeErrors` like so:

```suggestion
  const renderThemeErrors = ({errors}, prefix) =>
    errors.map(
      ({ errorCode }, index) =>
        errorCode.contains("ERR_FORM_TYPE") && (
          <ValidationError key={index} right>
            {THEME_ERROR_MESSAGES.ERR_FORM_TYPE_FORMAT}
          </ValidationError>
        )
    );
```

As, then it can be used when we come to add validation onto other things in the page.

MattElcock (changes requested):

eq-author/src/App/QuestionnaireDesignPage/MainNavigation/index.js
farres1: The ticket’s AC included allowing view survey to be clicked when the form type is invalid. I think the name themeErrorCount makes it misleading as you mentioned in another comment, that value represents only the form type validation error. This code is allowing view survey to be clicked when there is a form type error, and prevents it from being clicked when there is any other error, including if no theme is enabled.

jasehumphr (approved):

MattElcock (approved):

jasehumphr (approved):

PR #2059, opened 2021-05-26
EAR-1197-set-preview-theme

MattElcock:

This ticket adds the 'Preview theme' setting to each theme, which facilitates what theme is applied when viewing in Runner through Author.

I have also made some other refactoring changes along the way:

#### Accessibility refactoring

Vertical tabs heading is now a `h2` element so it can be read by screen readers as a heading to jump to.
Themes page heading is now a `h2` element so it can be read by screen readers as a heading to jump to.

I've made the above two the same level as each other as neither are beneath the other in importance. They're both a `h2` as `h1` is taken by the "Settings" title.

Themes page text is now a `p` element instead of a `label`, as it is not a label.

Unecessary `field` elements have been removed.

#### General refactoring

To encapsulate functionality, I've moved the EQ ID and Form Type logic into their own component that's defined in another file.

Created two new functions, `renderErrors` and `renderThemes`, to encapsulate their respective logic.

### How to review

Have a play around and make sure it all works OK.

### DIFF
--- a/eq-author/src/App/settings/EqId/index.js
+++ b/eq-author/src/App/settings/EqId/index.js
@@ -40,19 +40,25 @@ const Container = styled.div`
   float: left;
 `;
 
-const EqId = ({ eId, questionnaireId, shortName }) => {
+const EqId = ({ eqId, questionnaireId, shortName }) => {
   return (
     <Container>
       <Field>
         <Label>eQ ID</Label>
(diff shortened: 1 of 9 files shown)


martyncolmer (approved):
It works

farres1 (approved):

PR #2060, opened 2021-05-26
Disabling a theme set to be previewed now swaps to next available

MattElcock:

I forgot to implement the following AC:

Given I'm on the Themes tab of the Settings page
when I turn a theme off
and that theme was set as the Preview theme
and there are other themes turned on
then whichever theme is turned on first on the page will be set as the preview theme

### How to review

Make sure it works now


martyncolmer (approved):

tbaggaley (approved):
💥

PR #2067, opened 2021-06-03
Ear 1316 build modal for selecting questionnaire

zimboflyman:

Create a Storybook page for Popup modal containing questionnaireTable component
Style as per design:
https://overflow.io/s/EEGKYVGY?node=c55226b4&prototype

build it beneath the 'Patterns' directory in Storybook.

_npm run storybook in eq-author to see storybook_

I have created a component called **_QuestionnaireSelectModal_** which is a custom modal that will take the **_QuestionnairesView_** with a few extra props to help with the customized display of the list of questionnaires within the modal

https://collaborate2.ons.gov.uk/jira/browse/EAR-1316


farres1 (commented):
Hi Shane. I've had a look at the questionnaire select modal, and I really like how it's looking!

~~When I looked at it in Storybook, I initially tried to find it in the `Components/Modals` folder, where the confirmation modal is also found. It might be best for the questionnaire modal to go there too, so that all of the modal components are found in the same place. Let me know if you disagree with this though.~~ Discussed with Shane, the questionnaire select modal is found under patterns based on the ticket's description.

I think some tests are also needed for the modal.

eq-author/src/components/modals/QuestionnaireSelectModal/index.js
```diff
@@ -0,0 +1,111 @@
...
+    padding: 0;
+    width: 55em;
+  }
+`;
+
+const Container = styled.div`
+  background: white;
+`;
+
+const ModalTitle = styled.div`
```
MattElcock, line 28:
```suggestion
const ModalTitle = styled.h2`
```

This should be a heading tag to more accurately represent what it contains. Might need to update the styles when this is changed.

Chose h2 as h1 is the survey title in Author.

eq-author/src/components/modals/QuestionnaireSelectModal/index.js
```diff
@@ -0,0 +1,111 @@
...
+  color: ${colors.text};
+  margin-bottom: 0.75em;
+`;
+
+const ModalSubtitle = styled.div`
+  font-size: 1em;
+  color: ${colors.text};
+`;
+
+const ModalHeader = styled.div`
```
MattElcock, line 40:
```suggestion
const ModalHeader = styled.header`
```

Would be good I think if we as a team start using HTML5 elements, as they're more semantic and some assistive technologies hook onto them when giving aid. 

You can use `header` as both a heading for the whole page, or as a header for a section of content on the page.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header

eq-author/src/components/modals/QuestionnaireSelectModal/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+import styled from "styled-components";
+import { colors } from "constants/theme";
+import Modal from "components/modals/Modal";
+import Button from "components/buttons/Button";
+import ButtonGroup from "components/buttons/ButtonGroup";
+import ScrollPane from "components/ScrollPane";
+import Panel from "components/Panel";
+
+const ModalFooter = styled.div`
```
MattElcock, line 12:
```suggestion
const ModalFooter = styled.footer`
```

Would be good I think if we as a team start using HTML5 elements, as they're more semantic and some assistive technologies hook onto them when giving aid.

You can use `footer` as both a footer for the whole page, or as a footer for a section of content on the page.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/footer

MattElcock (changes requested):

eq-author/src/components/modals/QuestionnaireSelectModal/index.js
zimboflyman: nice one - thanks Matt - makes good sense!

tbaggaley (changes requested):
Hey Shane,

Design is looking good.

Some things I've noticed: 

- We could use implementing something that highlights the selected row blue (per the prototype). Probably will have to modify the modular version of `QuestionnairesTable` to do that.
- The search bar doesn't work in Storybook. Is this intentional (if so what will need to be done to make it work when things get plumbed together)?
- Same re: editor toggle switch.
- When I set `isOpen` to `false` in Storybook the modal still displays - so there's no way to get to view the docs page.

MattElcock (approved):

MattElcock:
Nice one Shane, looks great and will definitely help to save a lot of time when we piece it all together.

tbaggaley:
Hi Shane,

Looking good now the editor toolbar etc. is functional.

There seems to be no way to get the selected questionnaire out of the modal - is this required for this task or not? I note old ACs specified to use React context to store selected questionnaire details (which we decided against in the end - a plain old callback like `onQuestionnaireSelected` on the modal would suffice so the parent knows which Q has been chosen).

As it stands during the "connection" task we'd need to go back and re-jig this modal to add a means to get the currently selected questionnaire state as it isn't exposed right now. A problem for now - or a problem for later?

zimboflyman:
> As it stands during the "connection" task we'd need to go back and re-jig this modal to add a means to get the currently selected questionnaire state as it isn't exposed right now. A problem for now - or a problem for later?

Thanks Tom - ticket has this to say about it...
"Implement how the next modal is shown once the 'Select' button is pressed. (This will be done as part of a different subtask)." - so it's been assigned as a problem for later 😁

tbaggaley (approved):

PR #2069, opened 2021-06-04
Ear 1360 drag and drop mvp

MattElcock:

This PR introduces Drag and Drop into the secondary navigation menu, following the same style as. the P.O.C in an earlier ticket.

### How to review

Try it out! Try dragging and dropping under different scenarios, and see what happens.

You shouldn't be able to:

- Drag an item from a folder into the parent section, and vice versa.
- Drag a folder inside a folder.
- Drag anything between folders.
- Drag anything between sections.

You should be able to:

- Drag a page and confirmation question together.
- Reorder items in a section.
- Reorder items in a folder.
- Still use the old move modal to move content.


farres1:
Hi Matt.
Moving a folder with drag and drop sometimes places it below the intended position.
Sometimes it happens when moving a folder above its original position, and sometimes it happens when moving a folder below its original position. It never happens for both below and above the original position at the same time.
When this happens, questions are sometimes grouped together, and it is not possible to place the folder between the grouped questions.

farres1:
Hi Matt.

Adding a bug from the message earlier
> Calculated summary pages flicker when moved, and if a section contains a calculated summary then moving anything in that section also flickers.

tbaggaley:
Hi Matt,

Good job with this, looks good functionality wise! There may be something we can do to refactor the optimistic response generation etc. to avoid repeating it so much - but we could think about that at a later date when we come to rounding off the final behaviours (moving sections, etc.).

Just one Q about styling: it seems like only the icons trigger dragging but we get the "hand" cursor over the entire row, making you think you can drag the page ( / folder) title. Could we have it so either the icon only has the hand cursor - or have it so you can just drag the whole row? (Is there a reason we went for icon only?)

Otherwise happy to approve

MattElcock:
> Hi Matt,
> 
> Good job with this, looks good functionality wise! There may be something we can do to refactor the optimistic response generation etc. to avoid repeating it so much - but we could think about that at a later date when we come to rounding off the final behaviours (moving sections, etc.).
> 
> Just one Q about styling: it seems like only the icons trigger dragging but we get the "hand" cursor over the entire row, making you think you can drag the page ( / folder) title. Could we have it so either the icon only has the hand cursor - or have it so you can just drag the whole row? (Is there a reason we went for icon only?)
> 
> Otherwise happy to approve

Just by chance really, I agree with you it does make it a bit confusing. I'll look at changing that today so it's the entire row.

Reg the refactoring, definitely could do with some refactoring. But I also agree with you, we can do that when we come to tidy it up at the end. I think there will likely be some changes needed to it with some of the upcoming additions the RAD team are thinking about. We can do it after those.

tbaggaley (approved):
💥

zimboflyman (approved):
wow!  nice one Matt!

PR #2084, opened 2021-06-10
EAR 1369 Additional properties component

farres1:
> ### BEFORE MAKING YOUR PR
>
> Please ensure:
>
> - There are no linting errors, all tests must pass
> - PR is named after JIRA ticket number e.g. EAR-###
> - **Accesibility** checks are completed:
>   - Elements have discernible and consistent focus states
>   - Elements can be navigated to and used by just a keyboard
>   - Elements and text can be read out by a screen reader, and the descriptions are understandable
> - Your feature / bug fix works across **GCP** and **AWS** (where appropriate)
>   - Are modifications to eq-publisher-v3 required?
>   - Are modifications to the Firestore data source required?

---

Implementing additional content inside of a collapsible. The collapsible is a reusable component.

The error badge in the prototype will be added in a separate ticket

### How to review

**Check:**
- [ ] All answer types have a collapsible which includes the question's additional content
- [ ] Additional content has been removed from the right side bar for all answer types (the side bar will be removed in EAR 1370)
- [ ] Additional content is saved correctly when changes are made to it, including if toggle switch is enabled
- [ ] If any additional content toggle switch is enabled, the collapsible is open by default
- [ ] If no additional content toggle switch is enabled, the collapsible is closed by default
- [ ] Surveys display additional content correctly


tbaggaley (changes requested):
Hey Sam, just checked out the design (I'll look through code properly when you're finished working on it)

Looks really good - think it will be much nicer than using the old right side nav! All the toggles etc. seem to work as before.

Just one change I've noticed to do to match the prototype design wise: the "additional content" collapsible is underneath the question title (between title + answers) in the prototype, whereas here you've got it below all the answers. Just needs moving up and all should be good.

Screenshot 2021-06-17 at 12 36 20

eq-author/src/components/Collapsible/ContentCollapsible.js
```diff
@@ -0,0 +1,165 @@
+import React, { useState } from "react";
```
MattElcock, line 1: How come you've chosen to make this into a new file instead of editing the original to accept a varients prop to change it's appearence?

MattElcock (changes requested):

MattElcock (changes requested):

eq-author/src/App/page/PropertiesPanel/QuestionProperties/index.js
```diff
@@ -70,70 +56,114 @@ export class UnwrappedQuestionProperties extends React.Component {
...
+    const defaultOpen = () => {
+      if (
+        descriptionEnabled ||
+        definitionEnabled ||
+        guidanceEnabled ||
+        additionalInfoEnabled
+      ) {
+        return true;
+      }
+    };
```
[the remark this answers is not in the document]
tbaggaley: Ya - also doesn't need to be a function; a plain old value would do (since it has no inputs + just gets called straight away)

jasehumphr (approved):
SWEET

jasehumphr (approved):

eq-author/src/components/Collapsible/index.js
```diff
@@ -102,16 +147,30 @@ const Collapsible = ({
...
+        <Title
+          className="collapsible-title"
+          data-test="collapsible-title"
+          variant={variant}
+        >
           <ToggleCollapsibleButton
             isOpen={isOpen}
-            onClick={() => setIsOpen((isOpen) => !isOpen)}
+            onClick={
+              variant === "default" && (() => setIsOpen((isOpen) => !isOpen))
```
tbaggaley, line 167:
```suggestion
              variant === "default" ? (() => setIsOpen((isOpen) => !isOpen)) : undefined
```

tbaggaley (changes requested):

eq-author/src/components/Collapsible/index.js
```diff
@@ -102,16 +147,30 @@ const Collapsible = ({
...
-    <Wrapper className={className} data-test="collapsible">
-      <Header className="collapsible-header" data-test="collapsible-header">
-        <Title className="collapsible-title" data-test="collapsible-title">
+    <Wrapper className={className} data-test="collapsible" variant={variant}>
+      <Header
+        className="collapsible-header"
+        data-test="collapsible-header"
+        variant={variant}
+        onClick={
+          variant === "content" && (() => setIsOpen((isOpen) => !isOpen))
```
tbaggaley, line 156:
```suggestion
          variant === "content" ? (() => setIsOpen((isOpen) => !isOpen)) : undefined
```

This will stop you from getting console errors when the provided fn evaluates to `false`.

What's the idea with the header vs. the `ToggleCollapsibleButton` being clickable (or not) based on the "default" vs "content" variant? Just curious since I'm not at all familiar with the right hand panel redesign stuff :-)

tbaggaley (changes requested):
Storybook maybe needs updating too since your changes to make stuff into variants, it's failing to load after hitting the line: 

```javascript
import ContentCollapsible from "../../components/Collapsible/ContentCollapsible";
```

in `eq-author-app/eq-author/src/storybook/Components/Collapsible.stories.mdx`

(the referenced file doesn't exist anymore)

eq-author/src/components/Collapsible/index.js
farres1: When using default collapsible, we only want the collapsible to close/open when the collapsible text (ToggleCollapsibleButton) itself is clicked. The header is not visible, however extends across the page, so clicking that invisible header without the check would cause the collapsible to open. 

When using the content collapsible, we want the collapsible to close/open when either the header or the text (inside the ToggleCollapsibleButton) is clicked, as the content collapsible's header is visible.

Using that check makes sure that the entire header is clickable if it is a content collapsible, including the text itself. It also makes sure that only the text is clickable when using a default collapsible. Without the check either the header is clickable for default collapsible, or the text is not clickable for the content collapsible (only the header would be clickable).

jasehumphr (approved):
All Green Go Go

PR #2140, opened 2021-07-01
Ear 1348 section settings for hub

zimboflyman:

1. Mandatory section

GIVEN I am on the section page
WHEN the navigation is set to section navigation or none
THEN I don't see the mandatory section toggle

GIVEN I am on the section page
WHEN I have enabled hub navigation 
THEN I see the mandatory section toggle

GIVEN I have enabled hub navigation
ANDF i am on the section page
WHEN I set the mandatory section toggle
THEN that section will be set as mandatory section

GIVEN I have enabled hub navigation
AND i am on the section page
WHEN I unset the mandatory section toggle
THEN that section will not be set as mandatory section

2. Show on hub

GIVEN I have enabled hub navigation
AND i am on the section page
THEN the show on hub toggle will be ON by default but disabled

GIVEN I have enabled hub navigation
AND i am on the section page
WHEN I set the mandatory section toggle
THEN the show on hub toggle will be enabled

GIVEN I have enabled hub navigation
AND i am on the section page
WHEN I unset the mandatory section toggle
THEN the show on hub toggle will be disabled

GIVEN I have enabled hub navigation
AND i am on the section page
AND I set the mandatory section toggle
WHEN I set the show on hub toggle
THEN that section will appear on the hub

GIVEN I have enabled hub navigation
AND i am on the section page
AND I set the mandatory section toggle
WHEN I unset the show on hub toggle
THEN that section will not appear on the hub

### How to review

in Author front end - run the following to enable hub:
REACT_APP_FEATURE_FLAGS="hub" yarn start

Create a new section
you should see the new hub settings collapsible near the top of the section page as per this design:
 https://overflow.io/s/A6X68CMU?node=d5f36c8f&prototype&view=fill


martyncolmer (dismissed):

tbaggaley (changes requested):
Will we need a migration to set `showOnHub` to be `true` for existing questionnaires? They'll act as if it were `false` otherwise (so presumably both toggles will show as disabled initially on existing ones).

Haven't actually tried it out yet, just looking at code.

martyncolmer:
> Will we need a migration to set `showOnHub` to be `true` for existing questionnaires? They'll act as if it were `false` otherwise (so presumably both toggles will show as disabled initially on existing ones).
> 
> Haven't actually tried it out yet, just looking at code.

Migrated questionnaires work ok and will be edge cases, i'm happy for it not to be defaulted in that situation.

martyncolmer:
> Will we need a migration to set `showOnHub` to be `true` for existing questionnaires? They'll act as if it were `false` otherwise (so presumably both toggles will show as disabled initially on existing ones).
> 
> Haven't actually tried it out yet, just looking at code.

Yes, we will need a migration to add both requiredCompleate and showOnHub

martyncolmer (approved):

tbaggaley (approved):

PR #2165, opened 2021-07-08
✨ EAR 1318 - Create design for `review questions` modal

tbaggaley:

[Ticket](https://collaborate2.ons.gov.uk/jira/browse/EAR-1318)
[Prototype](https://overflow.io/s/EEGKYVGY?node=7601c426&prototype&prototype-hints=off)

Implement the component for reviewing selected questions as part of the importing process.

Other miscellany:

- Added `useModal` wrapper hook for conveniently triggering modals (without managing `isOpen` state). This is a generalisation of the existing hook I wrote for `useConfirmationModal` (see storybook files for usage). 
- Use subfolder of `Patterns` in storybook for components as per Matt E's ticket.
- Adds the `Wizard` modal component (a modal with a Back button). The review modal derives from this.
- Adds storybook page / docs for `Wizard` modal.

### How to review

- `yarn storybook`
- Check out the modal under `Patterns` -> `Importing content` -> `Review questions`
- Clicking `select questions` will select some mock questions for now (pending integration ticket to wire up with other modal)
- Should be able to remove single / all questions
- Import button should be disabled when no questions selected

#### No questions chosen
Screenshot 2021-07-08 at 15 41 09

#### Questions chosen
Screenshot 2021-07-08 at 15 41 44

### DIFF
--- a/eq-author/src/App/QuestionnaireSettingsModal/QuestionnaireMeta/__snapshots__/QuestionnaireMeta.test.js.snap
+++ b/eq-author/src/App/QuestionnaireSettingsModal/QuestionnaireMeta/__snapshots__/QuestionnaireMeta.test.js.snap
@@ -140,14 +140,12 @@ exports[`QuestionnaireMeta should render 1`] = `
       horizontal={true}
     >
       <Button
-        btn-focus={true}
         type="button"
         variant="secondary"
       >
         Cancel
       </Button>
       <Button
-        btn-focus={true}
         data-test="questionnaire-submit-button"
         disabled={true}
         type="submit"
--- a/eq-author/src/App/QuestionnaireSettingsModal/QuestionnaireMeta/index.js
+++ b/eq-author/src/App/QuestionnaireSettingsModal/QuestionnaireMeta/index.js
@@ -157,7 +157,7 @@ export const StatelessQuestionnaireMeta = ({
         </EnableDisableWrapper>
       </ToggleWrapper>
       <ButtonGroup horizontal align="right">
-        <Button onClick={onCancel} variant="secondary" type="button" btn-focus>
+        <Button onClick={onCancel} variant="secondary" type="button">
           Cancel
         </Button>
         <Button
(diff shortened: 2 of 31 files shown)


MattElcock (changes requested):
Hey Tom.

Code looks great; I love the storybook file. 😂

Some tweaks for the design;

- The back button doesn't match the design quite right (colour is blue where it should be grey, text isn't bold and the chevron isn't quite the same;
- The close 'x' button is blue where it should be grey;
- The onFocus styles for the following buttons aren't consistent with the rest of the app (you can import `focusStyles` from `theme.js` and use that to ensure they're all the same);
- - 'Remove all'
- - 'x' to remove a selected question
- - 'Select questions'
- - 'Import'
- - 'Cancel'

I also noticed that some of the clickable elements change the cursor when you hover over them (great) but don't change colour like other elements, is that intentional? Not sure if they all need to change colour onHover if we have the cursor changing, as it might be enough. Thought I'd mention it just in case 'though. (The elements are the 'back' button, 'remove all' text and the 'x' on questions.)

tbaggaley:
> Hey Tom.
> 
> Code looks great; I love the storybook file. 😂
> 
> Some tweaks for the design;
> 
> * The back button doesn't match the design quite right (colour is blue where it should be grey, text isn't bold and the chevron isn't quite the same; ✅
> * The close 'x' button is blue where it should be grey; ✅
> * The onFocus styles for the following buttons aren't consistent with the rest of the app (you can import `focusStyles` from `theme.js` and use that to ensure they're all the same); 
> * * 'Remove all' ✅
> * * 'x' to remove a selected question ✅
> * * 'Select questions' ✅
> * * 'Import' ✅
> * * 'Cancel' ✅
> 
> I also noticed that some of the clickable elements change the cursor when you hover over them (great) but don't change colour like other elements, is that intentional? Not sure if they all need to change colour onHover if we have the cursor changing, as it might be enough. Thought I'd mention it just in case 'though. (The elements are the 'back' button, 'remove all' text and the 'x' on questions.) ✅

Cheers Matt!

Styling changes implemented. Let me know if you still think anything needs tweaking.

I've added the `focusStyle` to `components/Button` so that buttons have that styling by default. Also changed `border-color: none` to `border-color: transparent` while I was in there because Chrome was complaining that `none` wasn't a valid property value - hence all the snapshot updates.

MattElcock (approved):

ashtuck (approved):
Followed the steps check. Matches the expected behaviour.

PR #2166, opened 2021-07-08
fix for expression migration

martyncolmer:
Fixing migration for schemas that have partial info or don't need deleteing.


jasehumphr (approved):

tbaggaley (approved):
Oops. Cheers Martyn ;-)

PR #2170, opened 2021-07-09
EAR-1269-Add-Eslint-rules-for-hooks

zimboflyman:

We are starting to use hooks a lot more. Would be good to have this eslint rule to catch any bugs we could introduce with using them incorrectly.

### How to review

in eq-Author:

1) update your yarn file - run yarn
2) check all tests pass -  yarn test
3) test new linter for hooks are all passing - yarn lint
4) test the hell out of author as we've updated a lot of dependencies in the callbacks that the linter has showed are missing, so we need to make sure everything builds nicely and still works!!!
5) mostly, the linter had required the adding of missing dependencies to the useCallback and useMemo arrays, But in some cases like 'MoveEntityModal' and 'Autocomplete' I've had to make changes to the way some constants were being used within the hook - so these functions would need solid testing in author to make sure things still work as expected 😁


eq-author/src/components/Autocomplete/temp.js
```diff
@@ -0,0 +1,316 @@
+import React, { useState, useRef, useCallback, useEffect } from "react";
```
tbaggaley, line 1: Hey Shane, was this file accidentally included...? :-)

zimboflyman: woops - now deleted! thanks

MattElcock (approved):

MattElcock (approved):

tbaggaley (approved):

PR #2171, opened 2021-07-09
EAR-1414-move-delete-options-not-visible-on-answer-options

ashtuck:
What is the context of this PR?

Fix for the button not appearing until it was hovered over. The problem was that the colour has been changed to white. Have changed it back to what it was, but it looks like there was a design change. Don’t know if this matches what the current design is supposed to me. But it’s a temporary fix until Monday.

How to review

Launch Author and double check that the button are visible and functioning when adding Checkbox Answers.

### DIFF
--- a/eq-author/src/App/introduction/Design/IntroductionEditor/CollapsiblesEditor/CollapsibleEditor/__snapshots__/index.test.js.snap
+++ b/eq-author/src/App/introduction/Design/IntroductionEditor/CollapsiblesEditor/CollapsibleEditor/__snapshots__/index.test.js.snap
@@ -7,6 +7,7 @@ exports[`CollapsibleEditor should render 1`] = `
 >
   <CollapsibleEditor__DetailHeader>
     <MoveButton
+      color="secondary"
       data-test="move-up-btn"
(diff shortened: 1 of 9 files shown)


farres1 (dismissed):

MattElcock (changes requested):
Hey Tucker,

Thanks for looking at this.

Unfortunately, these buttons are also used within an answer editor, where they need to be white. So, changing the colour back to `colors.secondary` isn't the fix we need. Or, it is, but an extra bit of work is needed to make it so we can recolour the buttons where we need to.

Here's an example of where the answer editor buttons are (currently) (the toolbar next to 'Number answer'):

image

Here's a screenshot of what they should be:

image

ashtuck:
Changed the `DeleteButton` & `MoveButton` components to accept a `color` prop. Right now just accepts `white` and `secondary` but defaults to `secondary` which is the most common use through out the app.

Changed the `AnswerEditor` to reflect the white design.

MattElcock (approved):

tbaggaley (approved):

PR #2196, opened 2021-07-27
Ear 1401 collapsible sections requires section titles

zimboflyman:

As an author
when I have answer summary and collapsible sections turned on
I want to able to add a section title
so they can be populated and also be visible in Runner

Section title enabled
GIVEN I have turned off section/hub navigation and turned on answer summary and collapsible sections
WHEN I am on a section page
THEN the section title field is enabled and can be populated and validated if empty

### How to review

Create a new Questionnaire
in settings:
Make sure Hub or Section Nav is NOT enabled
make sure answer summary and collapsible sections is turned on
you should be able to enter a section title on any section
An empty section title should have a validation message

Turn collapsible sections off in settings
section title should be disabled in all sections

### DIFF
--- a/eq-author-api/src/validation/schemas/section.json
+++ b/eq-author-api/src/validation/schemas/section.json
@@ -14,6 +14,9 @@
         },
         {
           "requiredWhenQuestionnaireSetting": "hub"
+        },
+        {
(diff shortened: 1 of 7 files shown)


zimboflyman:
made in error

PR #2227, opened 2021-08-12
Ear 1418 layout changes answer options

zimboflyman:
https://collaborate2.ons.gov.uk/jira/browse/EAR-1418

When I add multiple checkboxes or radio answer options we get active icons for moving the answer option up/down or deleting it. These are on the right, and should be left aligned to match the new design.
https://overflow.io/s/VUCP89DN?node=c7beadfc&prototype

### How to review

**Create any new Radio answer**
make sure the move and delete icons for options are aligned to the left as per design
you will need to add more than 2 options to make the delete option active

**Create any new Checkbox answer**
and a 2nd option - 'add checkbox'
make sure the move and delete icons for options are aligned to the left as per design

_make sure move and delete still functions correctly_


jasehumphr (approved):
SWEET

martyncolmer (approved):

PR #2307, opened 2021-09-24
Ear 1229 nested validation dont work with pubsub disabled

ashtuck:
### Context of PR ###

When PubSub is disabled. Error messages weren't updating due to not enough data being passed down in the GraphQL calls.

### How to review ###

- Go to `eq-author-api/schema/resolvers/createMutation.js`.
- Comment out lines `51-55`. This disables PubSub.
- Create a question with a Routing Rule.
- Check that the error messages disappear and update correctly.
- Create a some Skip Logic and check that the error messages disappear and update correctly.
- UnComment lines `51-55` in `createMutation.js` to reenable PubSub.

### DIFF
--- a/eq-author-api/schema/resolvers/createMutation.js
+++ b/eq-author-api/schema/resolvers/createMutation.js
@@ -15,48 +15,46 @@ const {
   UNPUBLISHED,
 } = require("../../constants/publishStatus");
 
-const createMutation = (mutation, { ignoreLockStatus = false } = {}) => async (
-  root,
-  args,
-  ctx
-) => {
-  let hasBeenUnpublished;
-  enforceHasWritePermission(ctx.questionnaire, ctx.user); // throws ForbiddenError
-  if (!ignoreLockStatus) {
-    enforceQuestionnaireLocking(ctx.questionnaire); // throws ForbiddenError
-  }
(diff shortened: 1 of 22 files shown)


jasehumphr (approved):

farres1 (changes requested):
This looks really good!

If you click the plus button on a routing rule or skip rule, the error badge doesn't update for the new errors until the page is refreshed

martyncolmer (approved):

farres1 (changes requested):
The previous error is fixed for me once an operator is selected, but before that the errors are still displayed. This means if you select an operator before selecting AND / OR, either refresh is required to remove the error or the operator must be reselected.

ashtuck:
> The previous error is fixed for me once an operator is selected, but before that the errors are still displayed. This means if you select an operator before selecting AND / OR, either refresh is required to remove the error or the operator must be reselected.

This has been fixed. There might be some other edge cases, if any others come up. I've gone through and i think that should be all of it i now. But let me know.

farres1 (changes requested):
The error stays when an answer is selected from "Select an answer" on an AND / OR logic condition

farres1 (approved):

MattElcock (approved):

PR #2315, opened 2021-09-28
Update to pass an error if meta data has been deleted in `Additional …

ashtuck:
### Context of PR ###

When piping in MetaData into the `Additional Contents` `Question Description`, if that MetaData was then deleted, the errors don't pass down meaning it isn't obvious to the user that there's an error.

### How to review ###

- Create a question and answer.
- Add a `Question Description` text.
- Create a custom MetaData.
- Pipe that MetaData into the  `Question Description`.
-  Then delete that MetaData tag and you should see the errors appear in the Design Page.

### DIFF
--- a/eq-author-api/src/validation/customKeywords/validatePipingMetadataInTitle.js
+++ b/eq-author-api/src/validation/customKeywords/validatePipingMetadataInTitle.js
@@ -18,7 +18,7 @@ module.exports = (ajv) =>
     keyword: "validatePipingMetadataInTitle",
     validate: function isValid(
       _schema,
-      title,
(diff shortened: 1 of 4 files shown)


farres1 (changes requested):
The error badge is being displayed correctly in the navigation bar, however the error message is not displayed on the design page. I think the question description text editor's border will also need to be changed to the error primary colour when there is an error

ashtuck:
Error message issue is now fixed.

farres1 (approved):

MattElcock (approved):

PR #2317, opened 2021-09-29
Ear 1125 deleted piped answer update title in navbar

zimboflyman:

Expected **behaviour**
Create 2 questions
Pipe the answer to the first question into the second question
Delete the first question
The question shows "[Deleted answer]" in the question title in the left nav
The question shows the deleted piped value in the question title on the page itself until you click on another question

Actual **behaviour**
Create 2 questions
Pipe the answer to the first question into the second question
Delete the first question
The question shows "[Piped value]" in the question title in the left nav
The question shows "[Piped value]" in the question title on the page itself until you click on another question

The same fix will be implemented for page.description

### DIFF
--- a/eq-author-api/schema/resolvers/base.js
+++ b/eq-author-api/schema/resolvers/base.js
@@ -656,10 +656,10 @@ const Resolvers = {
     ),
     deleteAnswer: createMutation((_, { input }, ctx) => {
       const page = getPageByAnswerId(ctx, input.id);
-
+      const pages = getPages(ctx);
       const deletedAnswer = first(remove(page.answers, { id: input.id }));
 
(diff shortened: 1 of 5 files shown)


eq-author/src/components/RichTextEditor/RichTextEditor.test.js
```diff
@@ -245,90 +245,19 @@ describe("components/RichTextEditor", function () {
         },
       ];
 
-      const metadata = [
-        {
-          id: "1",
-          displayName: "metadata 1",
-        },
-      ];
+      // const metadata = [
```
jasehumphr, line 248: How come we are saving this in a comment rather than deleting?

jasehumphr (approved):

PR #2419, opened 2021-11-24
Ear 1564 themed panel component

oms09:

> - Panel Component/ Error Summary Component Redesign.

### How to review

> - Check the theme colour in the storybook matches the ons design system.
> - Check the prop of paragraphLabel matches the border colour.


zimboflyman (commented):
All looks good except one small thing.
According to the design system - the error with header panel should have each error as a lin
Err panel Example
k?

oms09:
> All looks good except one small thing. According to the design system - the error with header panel should have each error as a lin Err panel Example k?

eq-author/src/components-themed/panels/panel.js
```diff
@@ -0,0 +1,184 @@
...
+        <>
+          <Header>
+            <HeaderLabel> {headerLabel} </HeaderLabel>
+          </Header>
+        </>
+      )}
+      {variant === "success" && (
+        <SuccessPanelIconContainer>
+          <SpanIcon>
+            <StyledImage src={icon} />
```
martyncolmer, line 154: As the success icon doesn't change, we should hard code it rather than pass it as a prop.

martyncolmer (changes requested):

eq-author/src/components-themed/panels/panel.js
martyncolmer: Also don't need the img tag if we are just importing the svg

zimboflyman (dismissed):
Nice!

eq-author/src/components-themed/panels/panel.js
```diff
@@ -0,0 +1,175 @@
...
+  variant: PropType.string,
+  children: PropType.node,
+  paragraphLabel: PropType.string,
+  headerLabel: PropType.string,
+  withLeftBorder: PropType.bool,
+};
+
+Panel.defaultProps = {
+  variant: "info",
+  type: "panel",
```
martyncolmer, line 172: Don't need to set the type

martyncolmer (changes requested):

zimboflyman (approved):

martyncolmer (approved):

PR #2421, opened 2021-11-25
Ear 879 remove default bluebox

oms09:
> ### BEFORE MAKING YOUR PR

---

> Blue border of the buttons when clicked or tabbed.
> Removed a function/unit test that tested for a default blue box.
> FocusMenuItem, handledEntered is called when onEntered which led to a default blue box. Removed this and added a focus selector which had all the same functionality as before just without it defaulting to a blue border when adding an answer.

### How to review

> Check that once clicked the border is blue.
> Check that you can tab through and the border is blue.
> Check that you can hold click on the button for the blue border.


zimboflyman (approved):
Work well and thanks for the PR details - helps to quickly work out whats going on!😁

farres1 (approved):

PR #2459, opened 2021-12-13
Ear 1533 logic reset

oms09:
> ### BEFORE MAKING YOUR PR
>
> Please ensure:
>
> 
---

> If you fill in an if condition on routing or skip logic and go back and reselect the if condition it will cause the condition, number/option to reset and error is shown.

> condition on expressions object is set to null as well as the whole of the right side object which contains the number/options depending on whether the answer is number or radio. 

> created a getRightSide to ensure its not null on the resolver (binaryExpression2). 

> Ternary operation on the expression condition to check if it has already has a condition so it doesn't reset to null.

> test cases to get if the logic is not null on reselecting the if condition.

### How to review

> check if the if condition on the routing/skip logic is being reset when reselecting the same condition.


zimboflyman (dismissed):
Works good - needs PR description for those that don't know the issue

eq-author-api/schema/resolvers/logic/binaryExpression2/index.js
```diff
@@ -168,9 +168,15 @@ Resolvers.Mutation = {
...
+      ...expression.right,
+    };
+
     expression.left = updatedLeftSide;
-    expression.right = null;
-    expression.condition = answerTypeToConditions.getDefault(answer.type);
+    expression.right = getRightSide;
+    expression.condition = expression.condition
+      ? expression.condition
+      : answerTypeToConditions.getDefault(answer.type);
```
farres1, line 179:
```suggestion
    expression.condition = expression.condition ?? answerTypeToConditions.getDefault(answer.type);
```

This change is called "Nullish coalescing operator":
- If the left side condition (`expression.condition`) is not null, it returns the left side (`expression.condition`).
- If the left side condition (`expression.condition`) is null, it returns the right side (`answerTypeToConditions.getDefault(answer.type)`)

This change will return the same result as the original code

farres1 (changes requested):
This fix is really good! Great work!

When a routing condition is created, previously the operator select box showed the option "Select an operator" by default. After the change, the operator select box shows "Equals" by default

eq-author-api/schema/resolvers/logic/binaryExpression2/index.js
zimboflyman: Nice Sam!

ashtuck (approved):

farres1 (approved):

PR #2524, opened 2022-01-24
Ear 348 deleted page crash

oms09:
> ### BEFORE MAKING YOUR PR
>
> Please ensure:
>
> - There are no linting errors, all tests must pass
> - PR is named after JIRA ticket number e.g. EAR-###
> - **Accesibility** checks are completed:
>   - Elements have discernible and consistent focus states
>   - Elements can be navigated to and used by just a keyboard
>   - Elements and text can be read out by a screen reader, and the descriptions are understandable
> - Your feature / bug fix works across **GCP** and **AWS** (where appropriate)
>   - Are modifications to eq-publisher-v3 required?
>   - Are modifications to the Firestore data source required?

---

> When directed to a page or section that is deleted the app crashes (gives a white screen error).
> E.g. If you add a new page and delete it and try going back to it with the url you will get an error because you're going back  to a delete page which is now null.
> Redirected the user to go to the first page/ section of the questionnaire instead of the deleted question/ page design
> Used RedirectRoute component to take the user from the url entered to the first page or section of the questionnaire depending whether or not they go to a deleted page or section.
> added additional checks to fix some type error mainly on section code to make sure a section exist before checking its properties. 

### How to review

- run author...
- make a questionnaire (business or social)
- create a page and add contents in it.
- copy the url and then delete the page.
- paste the url of the delete page and see if you get a error.
- if you're redirected back to the first page of the questionnaire then its working.
- same with section.


eq-author/src/App/section/Design/index.js
```diff
@@ -167,20 +187,25 @@ export const UnwrappedSectionRoute = (props) => {
     );
   };
 
-  const hasIntroductionContent = Boolean(
-    section.introductionTitle || section.introductionContent
-  );
+  const hasIntroductionContent = () => {
+    return (
+      section != null &&
+      Boolean(section.introductionTitle || section.introductionContent)
```
farres1, line 193:
```suggestion
      Boolean(section?.introductionTitle || section?.introductionContent)
```

farres1 (dismissed):
The fix is really good!

I've added two small changes, they're just a different way of handling null bugs

eq-author/src/App/page/Design/index.js
```diff
@@ -82,7 +94,25 @@ export const UnwrappedPageRoute = (props) => {
...
+        <RedirectRoute
+          from={"/q/:questionnaireId/page/:pageId/design"}
+          to={
+            "/q/:questionnaireId/page/" +
+            questionnaire?.sections[0]?.folders[0]?.pages[0]?.id +
+            "/design"
+          }
+        />
+      );
+    }
```
farres1, line 115:
```suggestion
    if (isEmpty(page)) {
      return (
        <RedirectRoute
          from={"/q/:questionnaireId/page/:pageId/design"}
          to={
            "/q/:questionnaireId/page/" +
            questionnaire?.sections[0]?.folders[0]?.pages[0]?.id +
            "/design"
          }
        />
      );
    }
```

The ?s will handle null conditions

eq-author/src/App/section/Design/index.js
```diff
@@ -98,6 +100,19 @@ export const UnwrappedSectionRoute = (props) => {
...
+      return (
+        <RedirectRoute
+          from={"/q/:questionnaireId/section/:sectionId/design"}
+          to={`/q/:questionnaireId/section/${questionnaire.sections[0].id}/design`}
+        />
+      );
+    }
+  };
+
+  const { questionnaire } = useQuestionnaire();
```
farres1, line 114:
```suggestion
  const redirectPage = () => {
    if (section === null) {
      return (
        <RedirectRoute
          from={"/q/:questionnaireId/section/:sectionId/design"}
          to={`/q/:questionnaireId/section/${questionnaire?.sections[0]?.id}/design`}
        />
      );
    }
  };

  const { questionnaire } = useQuestionnaire();
```

Opening a deleted section in a separate tab causes a crash without ?s

farres1 (approved):

jasehumphr (approved):

PR #2600, opened 2022-03-29
EAR-1677-BUG-import-content

zimboflyman:
https://collaborate2.ons.gov.uk/jira/browse/EAR-1677

2 issues reported in this ticket

1. Imported content was not showing up and needed a browser refresh
2. Importing a new question page when INSIDE a folder (eg - on a question) causes the imported question to appear inside of a new folder, instead of being imported into the pre-existing folder.

### How to review

Point 1. above does no longer seem to be an issue so I've fixed point 2.

Have a good play with importing, both from WITHIN a folder and when ON a folder
Check everything is importing as expected 😁


ashtuck (approved):

farres1 (changes requested):
The folder bug is fixed, this looks really good!

Currently only the else condition can be met, as folders no longer have an enabled value. The else statement can be removed - as folders do not have an enabled value, only the if condition needs to be met. The else condition will always place the imported content in the first position of the folder, instead of after the current question.

The code wrapped in the if condition no longer needs to be inside of the if condition and should always be run - every question page is now inside a folder, therefore importing will now always work in the same way it did when folders had an enabled value.

The position value on line 162 is unused and can be removed.

farres1 (approved):

ashtuck (approved):

PR #2644, opened 2022-05-04
filtered metadata options doesn't reset

oms09:

> Metadata key column gets reset when you press the arrow down key on filtered searches.
> Scrolling through the metadata with arrow down still works but the highlighted index is invisible.

### How to review

- Start up author aws/gcp
  - create a questionnaire and goto the metadata section of the questionnaire.
  - then, type in something to filter the metadata list e.g. 're'.
  - press arrow down and see if the list is reset (it shouldn't reset).
  - check you can scroll through the list with arrow down and you can see which key is highlighted.
  - check if hover still highlights key.

### DIFF
--- a/eq-author/src/App/metadata/MetadataTable/Controls/KeySelect.js
+++ b/eq-author/src/App/metadata/MetadataTable/Controls/KeySelect.js
@@ -39,17 +39,17 @@ class KeySelect extends Component {
   };
 
   handleStateChange = (changes) => {
-  const { name, onChange, onUpdate } = this.props;
+    const { name, onChange, onUpdate } = this.props;
     const { inputValue: value, selectedItem } = changes;
-    //isssue lies here
     if (selectedItem) {
       onChange({ name, value: selectedItem.value }, onUpdate);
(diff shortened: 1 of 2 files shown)


eq-author/src/App/metadata/MetadataTable/Controls/KeySelect.js
```diff
@@ -39,14 +39,17 @@ class KeySelect extends Component {
   };
 
   handleStateChange = (changes) => {
-    const { name, onChange, onUpdate } = this.props;
+  const { name, onChange, onUpdate } = this.props;
     const { inputValue: value, selectedItem } = changes;
-
+    //isssue lies here
```
farres1, line 44: This comment can be removed

farres1 (changes requested):
This looks a really good fix!

There is a comment in the code that can be removed, and some lines of code have not been formatted using Prettier.

martyncolmer (approved):

zimboflyman (approved):

PR #2686, opened 2022-06-22
EAR-1701 valid ons email

Paul-Joel:
### PR Context

https://jira.ons.gov.uk/browse/EAR-1701

- Changes primarily to the SignIn > CreateAccount code.
- Changes allow for the enforcement of a set of 'valid domains' for the submitted email address, via a new environment variable REACT_APP_VALID_EMAIL_DOMAINS. A second new envar (REACT_APP_ORGANISATION_ABBR) allows to set the 'organisation abbreviation' for use in error messages (with a no value fall back). Both envars have fallback code if neither they are not present/have no value.
- Changes also introduce the ability to add a field description to InputWithContextError and Password inputs, with label spacing adjustments if a description is present, and specific error message on the email input field for errors
- Update to colour declaration on link in header error declaration to resolve colour contrast issue (change to existing oceanBlue from blue)
- Addition of new test, and update to existing tests to cope with needing a valid email address now

### How to review

- Run the updated code, navigating to the Create Account page.
- Run both with and without the new environment variables to confirm behaviour
  - If REACT_APP_ORGANISATION_ABBR is not set, the 2 error messages and the field description should use 'authorised' instead
  - If REACT_APP_VALID_EMAIL_DOMAINS is not set the new validation will not apply, so any email will be allowed as now. The new field description will not display, as it does not apply
  - if REACT_APP_VALID_EMAIL_DOMAINS is set with a comma separated list of email domains (@ons.gov.uk etc), the email field will limit to email addresses based on that list and should work with single and multiple entries. A field description should appear to tell the user of the restrictions, and using a non listed email domain should produce a specific error
  - The password field should also have a new field description, but only on the Create Account page - SignIn should not display it

### DIFF
--- a/eq-author/src/App/SignInPage/CreateAccount/index.js
+++ b/eq-author/src/App/SignInPage/CreateAccount/index.js
@@ -29,8 +29,9 @@ const CreateAccount = ({
   const [fullName, setFullName] = useState("");
   const [password, setPassword] = useState("");
   let errorRefCreateAcc = useRef();
+  
   const textValidEmailError = process.env.REACT_APP_ORGANISATION_ABBR
-    ? "Enter a valid " + process.env.REACT_APP_ORGANISATION_ABBR + " email"
+    ? "Only " + process.env.REACT_APP_ORGANISATION_ABBR + " email addresses allowed"
     : "Enter a valid authorised email";
   const textValidEmailErrorMessage = process.env.REACT_APP_ORGANISATION_ABBR
     ? "Enter a valid " +
--- a/eq-author/src/App/SignInPage/CreateAccount/index.js
+++ b/eq-author/src/App/SignInPage/CreateAccount/index.js
@@ -29,10 +29,12 @@ const CreateAccount = ({
   const [fullName, setFullName] = useState("");
   const [password, setPassword] = useState("");
   let errorRefCreateAcc = useRef();
-  
+
   const textValidEmailError = process.env.REACT_APP_ORGANISATION_ABBR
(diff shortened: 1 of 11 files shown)


eq-author/src/App/SignInPage/CreateAccount/index.js
```diff
@@ -20,25 +20,56 @@ const CreateAccount = ({
...
+  setErrorMessageEmail,
   setVerificationEmail,
 }) => {
   const [createEmail, setCreateEmail] = useState("");
   const [fullName, setFullName] = useState("");
   const [password, setPassword] = useState("");
   let errorRefCreateAcc = useRef();
+  const textValidEmailError = process.env.REACT_APP_ORGANISATION_ABBR
+    ? "Enter a valid " + process.env.REACT_APP_ORGANISATION_ABBR + " email"
+    : "Enter a valid authorised email";
```
farres1, line 34: This looks really good, and works really well!

I've added a small change, based on the prototype.
```suggestion
  
  const textValidEmailError = process.env.REACT_APP_ORGANISATION_ABBR
    ? "Only " + process.env.REACT_APP_ORGANISATION_ABBR + " email addresses allowed"
    : "Enter a valid authorised email";
```

The first error banner's text in the prototype reads as "Only <Organisation> email addresses are allowed".

farres1 (changes requested):

farres1 (approved):

jasehumphr (approved):
sweet

farres1 (approved):

PR #2743, opened 2022-09-02
EAR-1805 Scroll Bar additional tweaks

Paul-Joel:
### PR context
Following PM discussion, it was decided to add further adjustments, as although the original changes work, when present the new scroll sits on top of the text and badge options, especially on windows devices. These additional changes widen the main navigation column to allow for a scroll bar to sit alongside the buttons, and to not obscure the badge icon if present

### How to review
View the Author interface, and ensure when the main navigation column has a vertical scroll bar it doesn't obscure the buttons, button text or any badges


farres1 (commented):
The work's looking really good!

Adding one small change, from the comment on the ticket, it looks like the users would like the scroll bar hidden on the main navigation and navigation sidebar. The code to do this would be:
`    &::-webkit-scrollbar {
      display: none;
    }
`
within the main navigation's and navigation sidebar's scroll panes.

This would also allow the styling changes to be reverted to their previous values.

martyncolmer (approved):

PR #2774, opened 2022-09-29
EAR-1837 Migration update

Paul-Joel:

Migration update following change to mutually exclusive qCode setting.
Adds migration update to ensure that existing qCode values for Mutually Exclusive options are used to set the Mutually Exclusive qCode value if one doesn't exist already. Options values are then removed to prevent error in Runner.

### How to review

Check that for an existing questionnaire with mutually exclusive options and no qCode value for the mutually exclusive answer the migration sets the ME qCode value to the first available value in MEO, and that all MEO qCode values are removed

### DIFF
--- a/eq-author-api/migrations/updateMutuallyExclusive.js
+++ b/eq-author-api/migrations/updateMutuallyExclusive.js
@@ -9,10 +9,7 @@ module.exports = (questionnaire) => {
 
   answers.map((answer) => {
     if (answer.type === MUTUALLY_EXCLUSIVE) {
-      if (
-        answer.options[0].qcode !== undefined ||
-        answer.options[0].qcode !== null
-      ) {
+      if (answer.options[0].qcode) {
         if (answer.qcode === undefined || answer.qcode === null) {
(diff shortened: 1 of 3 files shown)


eq-author-api/migrations/updateMutuallyExclusive.js
```diff
@@ -0,0 +1,25 @@
...
+
+  answers.map((answer) => {
+    if (answer.type === "MutuallyExclusive") {
+      if (answer.options !== undefined && answer.options !== null) {
+        answer.options.forEach((option) => {
+          if (option.qCode) {
+            if (!answer.qCode) {
+              answer.qCode = option.qCode;
+            }
+            option.qCode = undefined;
```
martyncolmer, line 17: Can we set the qCode to Null instead of undefined. Thanks.

martyncolmer (changes requested):

farres1 (approved):

martyncolmer (approved):

PR #2859, opened 2023-01-04
Ear 1800 page titles

Paul-Joel:

- Add new component to allow for the addition of new required fields to Section, Question, List Collector (3 fields), Confirmation Question, and Calculated Summary. 
- Adds additional functionality to collapsible to fix display issues with inconsistent use of margins 
- Add new component for output of inline/in text external links 
- Adds a new migration to ensure that the new fields show as mandatory when existing questionnaires are opened or imports made. 

Note: position of new field has changed from Jira ticket in discussion with Lauren. New field now appears directly after existing title field.

Requires matching eq-publisher-v3 changes to format and output field values for EQ
JIRA: https://jira.ons.gov.uk/browse/EAR-1800 
Publisher PR: https://github.com/ONSdigital/eq-publisher-v3/pull/137

### How to review

1. New Questionnaire (tests new fields)
* Create a new questionnaire
* Create a new section, ensuring that the new field shows in the correct location and method, is marked as required on the screen when empty, and contributes to the count of errors for the page correctly. Ensure entered text is stored, retrieved, and in the export. Ensure 'block' is displayed correctly for spacing and position
* Repeat for question
* Repeat for calculated summary
* Repeat for confirmation question
* Repeat for list collector, noting that this has 3 instances of page title, one for each section. Ensure all 3 behave correctly, and that they store and output their values independently
2. Existing questionnaire (test new migration)
* Open an existing questionnaire, and ensure that the new fields are marked as required and contribute to the error counts
3. Importing questions/sections (test new migration)
* Import sections and questions from a non-updated questionnaire, and ensure that they also immediately trigger the required state, and contribute to the error count
4. Test new 'block'
* Ensure text in new block matches spec


eq-author/src/App/questionConfirmation/Design/index.js
```diff
@@ -99,6 +99,7 @@ export const UnwrappedQuestionConfirmationRoute = ({
   return (
     <EditorLayout
       title={data?.questionConfirmation?.displayName || ""}
+      pageDescription={data?.questionConfirmation?.pageDescription || ""}
```
farres1, line 102:
```suggestion
```
This line can be removed as EditorLayout does not have a pageDescription prop, and so is unused

farres1 (changes requested):

eq-publisher/src/getQuestionnaire/queries.js
```diff
@@ -234,6 +234,7 @@ exports.getQuestionnaire = `
         title
         introductionTitle
         introductionContent
+        pageDescription
```
farres1, line 237:
```suggestion
```
This line can be removed  - this file is only used for v2 Publisher, which is no longer maintained

eq-publisher/src/getQuestionnaire/queries.js
```diff
@@ -276,6 +277,7 @@ exports.getQuestionnaire = `
             ... on QuestionPage {
               description
               descriptionEnabled
+              pageDescription
```
farres1, line 280:
```suggestion
```
This line can be removed  - this file is only used for v2 Publisher, which is no longer maintained

eq-publisher/src/getQuestionnaire/queries.js
```diff
@@ -287,6 +289,7 @@ exports.getQuestionnaire = `
               confirmation {
                 id
                 title
+                pageDescription
```
farres1, line 292:
```suggestion
```
This line can be removed  - this file is only used for v2 Publisher, which is no longer maintained

farres1 (changes requested):

farres1 (approved):

farres1 (dismissed):

farres1 (approved):

PR #2885, opened 2023-01-23
EAR 1815 Answer required and mutually exclusive

jchapman68:

> When a mutually exclusive answer is added then a banner is placed below the answer required toggle switch of the main answer to indicate that the answer required will also apply to the mutually exclusive answer.
>
> E.g. _"Currently, if you add a new question to Author the entire app crashed. This is because..."_

### How to review

- Create a questions with answers of type Number, Unit, Duration, TextField, TextArea, Date, Checkbox and Select.
- For each of the above check that Answer Required has a banner below it when a Mutually Exclusive answer is added and not shown before and after the Mutually Exlusive answer is removed.
- Check that DateRange and Radio cannot have mutually exclusive answer added and that the banner is not shown below the Answer Required toggle switch.

### DIFF
--- a/eq-author/src/App/page/Design/answers/DateSingle/index.js
+++ b/eq-author/src/App/page/Design/answers/DateSingle/index.js
@@ -28,7 +28,11 @@ export const DateSingle = ({ answer, onChange, onUpdate, ...otherProps }) => {
         showDummyDate={false}
         {...otherProps}
       />
-      <AnswerProperties answer={answer} updateAnswer={updateAnswer} />
+      <AnswerProperties
+        answer={answer}
+        updateAnswer={updateAnswer}
+        page={otherProps.page}
+      />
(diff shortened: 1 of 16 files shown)


Paul-Joel (dismissed):

masiketlamelo (approved):

sudeepkunhis (changes requested):
@jchapman68 
Just spotted something - in custom-prop-types/index.js and AnswerProperties/index.js - we don’t need the line for hasMutuallyExclusiveAnswer because it is not directly passed as a prop to ‘AnswerProperties’. It is present in Required/index.js. I have commented the two lines as shown below and the feature still works alright. All else looks good, okay to approve once you have a look, thanks.

Paul-Joel (approved):

masiketlamelo (approved):

PR #2895, opened 2023-01-31
EAR-1972 Change default theme

Paul-Joel:

Removes the existing 'GB'/default theme, and replaces it with the 'business' theme in the UI
Also changes all instances of the 'default' theme value in 'theme', 'previewTheme' and 'shortName' throughout the code
Uses a migration to update existing values from 'default' to 'business'

Note changes do not fix the known issue with the value of the top level 'theme' not updating, it only changes the value from 'default' to prevent an error with ENUM matching

### How to review

Pre-test
1. Open the UI on the master branch
2. Create a questionnaire set to the 'GB'/default theme
3. Create a questionnaire set to the 'Social' theme

Test change
1. Open the UI on the PR branch
2. Create a new questionnaire
4. Confirm in 'Settings' > 'Themes' that the 'Business' theme is selected by default.
5. Check the export to ensure that the 'business' theme is set to 'true', and the 'previewTheme' value is set to 'business'
6. Change the theme value via the UI, ensuring to deselect the 'business' theme.
7. Ensure the export reflects the change correctly.

Test migration 
1. Still on the PR branch
2. Open the questionnaire that was created with the 'GB'/default theme
3. Check it is now set to the 'Business' theme in the UI and the export 
4. Open the questionnaire that was created with the 'Social' theme
5. Check it is still set to the 'Social' theme in the UI and the export

### DIFF
--- a/eq-author-api/migrations/updateDefaultToBusinessTheme.test.js
+++ b/eq-author-api/migrations/updateDefaultToBusinessTheme.test.js
@@ -1,0 +1,27 @@
+const updateUKIStoBEISTheme = require("./updateDefaultToBusinessTheme");
+
+describe("updateDefaultThemes", () => {
+  it("should update default theme", () => {
+    const questionnaire = {
+      themeSettings: {
+        id: "1",
+        previewTheme: "default",
+        themes: [
+          {
+            enabled: true,
+            shortName: "default",
+            legalBasisCode: "VOLUNTARY",
+            eqId: "123",
+            formType: "1234",
+            id: "default",
+          },
+        ],
+      },
+    };
+
+    const updatedTheme = updateUKIStoBEISTheme(questionnaire);
+    expect(updatedTheme.themeSettings.themes[0].shortName).toBe("business");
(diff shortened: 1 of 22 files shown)


eq-author-api/migrations/updateDefaultToBusinessTheme.test.js
```diff
@@ -0,0 +1,27 @@
+const updateUKIStoBEISTheme = require("./updateDefaultToBusinessTheme");
```
martyncolmer, line 1: Function name doesn't match the action

martyncolmer (changes requested):

eq-author-api/migrations/updateDefaultToBusinessTheme.test.js
Paul-Joel: Renamed function

farres1 (approved):

sudeepkunhis (approved):

martyncolmer (approved):

PR #3034, opened 2023-08-03
EAR 1948 Option values for rad-chk-ans

jchapman68:
---

> Add the capability to specify values for each option in a radio/checkbox/select answer types.

### How to review

- Create a questionnaire with a radio, checkbox and select questions.
- Update settings to enable the Q Codes and Values tab
- Open to Q Codes and Values page and confirm layout is as per [design](https://www.figma.com/file/D0nxCipkp5HJxeF5FGl7fm/EAR---1948-Option-values-for-radio%2Fcheckbox%2Fselect-answers?type=design&node-id=0-1&mode=design&t=ZsCiolVJZfW8wUvG-0)
- Check that when values and q codes have been entered then validation messages are removed
- Check that when all validation messages have gone then the validation dot on the nav button also goes
- Check the survey export and convert schemas that option values are set
- Check that if an option value is missing then the label value is used instead


Paul-Joel (approved):

eq-author/src/App/qcodes/QcodesPage.js
```diff
@@ -30,9 +32,25 @@ const StyledMainCanvas = styled(MainCanvas)`
   max-width: 80em;
 `;
 
+const Padding = styled.div`
+  padding: 1em;
```
shelleytruong, line 36: const Padding = styled.div`
    margin: 2em auto 1em;
    width: 100%;
    padding: 0 0.5em 0 1em;
    max-width: 80em;
';

changes padding/margin to the same as the qcodes table when fullscreen

shelleytruong (changes requested):
Changes look good! 
Just a small change based on the prototype design.

shelleytruong (approved):

Paul-Joel (approved):

PR #3111, opened 2024-04-30
EAR-2397-Validation-message-is-not-displaying

Farhanam76:
Ticket: https://jira.ons.gov.uk/browse/EAR-2397

### How to review
1. Reproduce the piping  answer to label as shown in the ticket in steps to reproduce 
2. ensure that a validation message of : "The answer being piped is now later in the questionnaire" is shown when you moved the piped label above the questionnaire where the piped answer is created

3. - [ ] If there are a lot of commits or some are not helpful to read, squash them down
4. - [ ] Click the **Merge** button on this pull request
5. - [ ] Does this change mean the **Capability examples** questionnaire should be updated?
6. - [ ] Move the Jira ticket for this task into the next stage of the process


eq-author/src/App/page/Design/answers/BasicAnswer/index.js
```diff
@@ -150,9 +153,14 @@ export const StatelessBasicAnswer = ({
           if (error.errorCode === "PIPING_TITLE_DELETED") {
             message = PIPING_TITLE_DELETED.message;
           }
+          if (error.errorCode === "PIPING_TITLE_MOVED") {
```
Paul-Joel, line 156: As there are lots of errorCode values being checked for here, but only one message being generated it would be better to re-write this as case statements. Currently the message can constantly be re-written as it goes.

Also, other constants are prefixed with ERR_ (although not PIPING_TITLE_DELETED). Probably should be consistent with the naming conventions.

eq-author/src/App/page/Design/answers/BasicAnswer/index.js
```diff
@@ -150,9 +153,14 @@ export const StatelessBasicAnswer = ({
...
+            message = PIPING_TITLE_MOVED.message;
+          }
           return (
             error.field === "label" && (
-              <ValidationError key={error.id}>{message}</ValidationError>
+              <ValidationError key={error.id === "PIPING_TITLE_MOVED"}>
```
Paul-Joel, line 161: Not sure why this has been added. This is either not required, or it needs to be set to the correct value at all times. It can't be hardcoded to PIPE_TITLE_MOVED when the message can be from a different error entirely.

Paul-Joel (changes requested):
Some potential issues

Paul-Joel (approved):
LGTM

PR #3124, opened 2024-08-02
EAR 2317 homepage context

farres1:
*While "Search" is separate to "Filters" in the application, for this PR's review instructions, the term "filters" also includes "search"*

API changes to allow questionnaires to be filtered based on filters applied by the user

https://jira.ons.gov.uk/browse/EAR-2317

### How to review

**Check:**
- [ ] Questionnaires are filtered based on each individual filter field
- [ ] Questionnaires are filtered correctly when multiple filters are applied simultaneously
- [ ] Questionnaires are sorted correctly based on the `sortBy` value
- [ ] The `Search` filter is applied based on both questionnaire title and questionnaire short code
- [ ] On the "My questionnaires" page, only questionnaires the user has editor access to are displayed *(EAR 2323 will iterate on "My questionnaires" functionality to allow users to add any questionnaires they can view to their "My questionnaires" list)*
- [ ] On the "All questionnaires" page, all non-public questionnaires the user has editor access to and all public questionnaires are displayed - non-public questionnaires the user does not have editor access to should not be displayed
- [ ] 10 questionnaires are displayed on each page based on the default `resultsPerPage` value, with the exception of the last page which may have fewer questionnaires depending on the total number of questionnaires
- [ ] The 10 questionnaires after the `lastQuestionnaireIdOnPage` are retrieved when opening the "Next page"
- [ ] The 10 questionnaires before the `firstQuestionnaireIdOnPage` are retrieved when opening the "Previous page"
- [ ] The total page count is correctly calculated


eq-author-api/db/datastore/datastore-mongodb.test.js
```diff
@@ -340,6 +379,649 @@ describe("MongoDB Datastore", () => {
...
+          mockQuestionnaire({
+            title: "Test questionnaire 7",
+            ownerId: "user-4",
+            createdAt: new Date(2021, 4, 10, 5, 0, 0, 0),
+          }),
+          ctx
+        );
+      });
+
+      it("Should return questionnaires with title containing the search term", async () => {
```
Paul-Joel, line 479: *should return
Issue throughout

Paul-Joel (changes requested):

eq-author-api/db/datastore/datastore-mongodb.test.js
farres1: Based on a technical discussion held today, all tests in this file have been updated to follow the "should" format (lower case) instead of the previous "Should" format - this includes changes to both the new tests and the pre-existing tests in this file.

Paul-Joel (approved):

martyncolmer (approved):

PR #3125, opened 2024-08-08
EAR-2449 Session Timeout

sanjeevz3009:
Implement session time-out for Author.

Credit to @martyncolmer and @farres1 for the assistance/ part-taking in discussions on the direction of this ticket.

To add functionality for Author users to automatically have their sessions timed out and require the users to sign in once every 7 days.

Jira ticket: https://jira.ons.gov.uk/browse/EAR-2449

### Files changed: `src/App/MeContext.js`

Added the session timeout functionality inside the first `useEffect` within the `ContextProvider` function. 

### How to review

To test the functionality you can replace this line of code 
`const sessionDuration = 1000 * 60 * 60 * 24 * 7;` which is 7 days in milliseconds to 5 or 10 seconds by copying and pasting either this line
`sessionDuration = 10000` for 10 seconds (in milliseconds) or `sessionDuration = 5000` for 5 seconds (in milliseconds).

Next sign into the Author application and you can see it time out the user after 5 seconds or 10 seconds or after the duration set by the developer.

This should work regardless of whether the user is on the Author application or not. It should work even if the user closes the application and comes back.

### Important information: Things to consider before merging
All the users who are already signed in and have been signed in for 7 days will be signed out immediately when the code changes go into production. This will affect the majority of the users, as we are assuming the majority of the users have been signed in for more than 7 days and haven't signed out.

For more information please refer to the comment in the JIRA ticket: https://jira.ons.gov.uk/browse/EAR-2449

### DIFF
--- a/eq-author/src/App/MeContext.js
+++ b/eq-author/src/App/MeContext.js
@@ -101,9 +101,12 @@ const ContextProvider = ({ history, client, children }) => {
         } else {
           user.getIdTokenResult().then((idTokenResult) => {
             const authTime = idTokenResult.claims.auth_time * 1000;
-            const sessionDuration = 10000; // 10 seconds
+            // Seven days in milliseconds
+            // const sessionDuration = 1000 * 60 * 60 * 24 * 7;
+            const sessionDuration = 20000; // 10 seconds
             const millisecondsUntilExpiration =
               sessionDuration - (Date.now() - authTime);
+            console.log(millisecondsUntilExpiration);
             sessionTimeout = setTimeout(
               () => auth.signOut(),
               millisecondsUntilExpiration
--- a/eq-author/src/App/MeContext.js
+++ b/eq-author/src/App/MeContext.js
@@ -89,23 +89,28 @@ const ContextProvider = ({ history, client, children }) => {
 
   useEffect(() => {
     // be aware that the return from auth.onAuthStateChanged will change on firebase ver 4.0
-    // https://firebase.google.com/docs/reference/js/v8/firebase.auth.Auth#onauthstatechanged
+    // https://firebase.google.com/docs/reference/js/v8/firebase.auth.Auth#onauthstatechanged]
+    // This useEffect hook is responsible for handling the authentication state changes in Firebase.
+    // It listens for changes in the authentication state using the onAuthStateChanged method.
+    // When the authentication state changes, it updates the firebaseUser state and sets awaitingFirebase to false.
     auth.onAuthStateChanged((user) => {
       setFirebaseUser(user);
       setAwaitingFirebase(false);
+      // It also sets up a session timeout for the user if they are authenticated.
+      // If the user is not authenticated, the session timeout is cleared using clearTimeout.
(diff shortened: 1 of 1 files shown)


eq-author/src/App/MeContext.js
```diff
@@ -89,10 +89,34 @@ const ContextProvider = ({ history, client, children }) => {
 
   useEffect(() => {
     // be aware that the return from auth.onAuthStateChanged will change on firebase ver 4.0
-    // https://firebase.google.com/docs/reference/js/v8/firebase.auth.Auth#onauthstatechanged
+    // https://firebase.google.com/docs/reference/js/v8/firebase.auth.Auth#onauthstatechanged]
```
Paul-Joel, line 92: Square bracket has been added to the end of this line. Can't see why.

eq-author/src/App/MeContext.js
```diff
@@ -89,10 +89,34 @@ const ContextProvider = ({ history, client, children }) => {
...
+      // If the user is not authenticated, the session timeout is cleared using clearTimeout.
+      let sessionTimeout = null;
+      if (user === null || user === undefined) {
+        sessionTimeout && clearTimeout(sessionTimeout);
+        sessionTimeout = null;
+      } else {
+        // If the user is authenticated, it retrieves the ID token result and calculates the session duration.
+        user.getIdTokenResult().then((idTokenResult) => {
+          const authTime = idTokenResult.claims.auth_time * 1000;
+          // The session duration is set to 7 days in this example.
```
Paul-Joel, line 109: This isn't an example, it's the actual code.

eq-author/src/App/MeContext.js
```diff
@@ -89,10 +89,34 @@ const ContextProvider = ({ history, client, children }) => {
...
+          const sessionDuration = 7 * 24 * 60 * 60 * 1000;
```
Paul-Joel, line 110: I'm not keen on this as a sum - it really ought to be just the result of the sum. Putting the calculation in a comment after is fine to show how the value was determined though.

Paul-Joel (changes requested):
Couple of minor tweaks.

Testing this however raises a deployment issue - running the code immediately signs you out. This means that it would do that to every user as soon as it is deployed.
Suspect that the code cannot be merged until a specific rollout point is agreed, and all users are warned and notified.

Also, is there an implementation for Author v2? All changes in v1 must be either implemented in v2, or a ticket raised for v2 work to ensure that the changes are not missed. As v2 has already implemented sign in, it really ought to be implemented in v2 now.

eq-author/src/App/MeContext.js
sanjeevz3009: Comment updated!

eq-author/src/App/MeContext.js
sanjeevz3009: The calculation for the session duration updated!

Paul-Joel (approved):

sudeepkunhis (approved):

PR #3142, opened 2025-06-02
EAR-2609: Publish only valid questionnaires

Paul-Joel:

Changes add a 'valid' trap to the 'Publish' button, only enabling the button to trigger the publish process when the Questionnaire has no errors, and no QCodeErrors

### How to review

Launch Author
Within a questionnaire, ensure that the publish button (on the Publish page) is disabled and not clickable when there are errors with the questionnaire and/or q codes. Only when no errors from both sources should the publish button be enabled.


farres1 (approved):

oms09 (approved):
looks good 👍

PR #3170, opened 2026-04-24
add check for GIUD

martyncolmer:
This PR adds checks to ensure the right GUID is returned from CIR.
When publishing to CIR we provide a GUID, this check ensures the GUID that is returned for CIR is the same as the one we provided.

### DIFF
--- a/eq-author-api/schema/resolvers/utils/publishSchema.js
+++ b/eq-author-api/schema/resolvers/utils/publishSchema.js
@@ -150,10 +150,19 @@ const publishSchema = async (ctx) => {
       `guid=${ctx.questionnaire.questionnaireVersionId}&validator_version=0.0.0`,
       publishResult
     );
+
+    //check CIR id matches the versionId to ensure the correct GUID has been used in CIR
+    if(publishResult.cirId !== ctx.questionnaire.questionnaireVersionId) {
+            publishResult.success = false;
+            publishResult.errorMessage = `CIR gateway did not return expected guid - expected ${ctx.questionnaire.questionnaireVersionId} but got ${publishResult.cirId}`;
+            publishResult.displayErrorMessage = "Contact eQ services team";
+    };
+
     logger.info(
       publishResult,
       `publish questionnaire with version id ${ctx.questionnaire.questionnaireVersionId}  - posted to first CIR gateway`
     );
+
 
     // post to second gateway if enabled and first gateway publish was successful
     // The ci_vserion is added to the second post to ensure the CI in both environments can be referenced by the same ci version number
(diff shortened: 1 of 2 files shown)


petechd (approved):

VirajP1002 (approved):

eq-author-api/schema/resolvers/utils/publishSchema.js
farres1, line 1: Prettier has not run on this file.

eq-author-api/schema/tests/publishSchema.test.js
farres1, line 1: Prettier has not run on this file.

eq-author-api/schema/resolvers/utils/publishSchema.js
```diff
@@ -168,18 +177,21 @@ const publishSchema = async (ctx) => {
...
-    );
 
-    //check CIR id matches the versionId to ensure the correct GUID has been used in CIR, commented out until CIR complete thier work to return the GUID
-    /* if(publishResult.cirId !== ctx.questionnaire.questionnaireVersionId) {
-            publishResult.success = false;
-            publishResult.errorMessage = `CIR gateway did not return expected guid - expected ${ctx.questionnaire.questionnaireVersionId} but got ${publishResult.cirId}`;
-            publishResult.displayErrorMessage = "Contact eQ services team";
-        } */
+
   } finally {
```
farres1, line 192: Currently the code includes a `try` block alongside a `finally` block.  The `finally` block runs regardless of the outcome of the `try` block (i.e. the `finally` block's code will be executed whether an error is caught or not), and therefore the `return` statement within the `finally` block is recognised as an error. Currently the error `Unsafe usage of ReturnStatement.` is provided as a result.

Since this change is related to pre-existing code, I will create a ticket to resolve the error separately.

farres1 (changes requested):

eq-author-api/schema/resolvers/utils/publishSchema.js
martyncolmer: Moved the return outside of the finally statement. I don't envisage there being a need for an error catch block as they are handled in the individual function calls and the try handles the flow control to exit the block early having had the publish result already updated with the errors..

martyncolmer: Put it back as the tests broke

farres1 (approved):

eq-author-api/schema/resolvers/utils/publishSchema.js
farres1: I have created the ticket to resolve this (EQS-899).

oms09 (approved):

PR #3180, opened 2026-07-28
Update publish endpoint and validator version

martyncolmer:
- Added validator version to publish process
- Updated endpoint for republish

### DIFF
--- a/eq-author-api/schema/resolvers/utils/publishSchema.js
+++ b/eq-author-api/schema/resolvers/utils/publishSchema.js
@@ -85,7 +85,7 @@ const postSchema = async (
     `${gateway}${endpoint}?${query}`,
     audience,
     {
-      method: method,
+      method,
       body: JSON.stringify(validatedQuestionnaire),
       headers: { "Content-Type": "application/json" },
     }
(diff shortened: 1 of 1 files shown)


oms09 (dismissed):
looks good but there is a typo   `guid=${ctx.questionnaire.questionnaireVersionId}&validator_version=${validatorVersion}&&ci_version=${publishResult.cirVersion}`, 

&& inside a string should be single

farres1 (approved):

eq-author-api/schema/resolvers/utils/publishSchema.js
```diff
@@ -67,20 +67,25 @@ const validateQuestionnaire = async (convertedQuestionnaire, publishResult) => {
...
   query,
   publishResult
 ) => {
   await authorisedRequest(
-    `${gateway}collection-instruments?${query}`,
+    `${gateway}${endpoint}?${query}`,
     audience,
     {
-      method: "POST",
+      method: method,
```
farres1:
```suggestion
      method,
```
Since this assigns to a value with the same name, this can be negated.

farres1 (changes requested):

farres1 (approved):

oms09 (approved):

